PR #12, opened 2026-03-19
Extend GQA support to packedQKV GQA in rewrite rules

### LINKED ISSUE #11: Consolidate GQA optimization: add packed QKV support and improve test coverage
# Consolidate GQA optimization: add packed QKV support and improve test coverage

## Before You Start

1. **Read the skill**: `.github/skills/writing-rewrite-rules/SKILL.md` — covers the `RewriteRuleClassBase` API, pattern matching, `check()`/`rewrite()` methods, file organization, and testing conventions.
2. **Read the existing rule**: `src/mobius/rewrite_rules/_group_query_attention.py` (224 lines) — this is the rule you will extend.
3. **Read its tests**: `src/mobius/rewrite_rules/_group_query_attention_test.py` (156 lines) — 7 existing tests.
4. **Read the ORT GQA spec**: https://github.com/microsoft/onnxruntime/blob/main/docs/ContribOperators.md — search for `GroupQueryAttention`. Note the packed QKV input format.

## Repo Conventions

```bash
# Install for development
pip install -e ".[transformers,testing]"

# Run rewrite rule tests only (~fast)
python -m pytest src/mobius/rewrite_rules/ -q --tb=short

# Run GQA tests specifically
python -m pytest src/mobius/rewrite_rules/_group_query_attention_test.py -sv

# Run full non-integration test suite
python -m pytest tests/build_graph_test.py tests/cli_test.py src/ -q \
  -k "not phi4mm and not apply_weights_unknown" --tb=short -n auto

# Lint and auto-format
lintrunner f --output oneline --all-files
```

- `from __future__ import annotations` in every module
- Ruff for linting/formatting (line-length=95, Python 3.10+)
- MyPy strict mode (excludes `*_test.py`)
- Git: linear commit history, always `--signoff`
- Test files must end in `_test.py`
- Co-located tests: `src/mobius/**/*_test.py` next to source

### IN scope
- Extend `RotaryAttentionToGQA` in `_group_query_attention.py` to attempt packed QKV, falling back to separate Q/K/V when packing is not possible
- Tests for packed QKV behavior (added to existing `_group_query_attention_test.py`)

### OUT of scope — do NOT implement
- Static cache + GQA (the GQA pattern requires `past_key`/`past_value` graph inputs and `attention_mask`; static cache uses `TensorScatter` + `nonpad_kv_seqlen` instead — the pattern cannot match)
- Creating new rule files or new rule classes — all changes go in the existing file and class
- Benchmark / arch diff entries for GQA-optimized models
- `optimize` parameter on `build_from_module()` API
- Changes to the Attention component or decoder layer
- New model architectures

### Existing Rule: `RotaryAttentionToGQA`

**File**: `src/mobius/rewrite_rules/_group_query_attention.py`
**Class**: `RotaryAttentionToGQA(RewriteRuleClassBase)` (line 31)

**What it matches** (`pattern()` at line 74):
```python
def pattern(self, op, q_pre, k_pre, v, attention_bias, past_key, past_value, cos, sin):
    q_rot = op.RotaryEmbedding(q_pre, cos, sin, _allow_other_attributes=True)
    k_rot = op.RotaryEmbedding(k_pre, cos, sin, _allow_other_attributes=True)
    return op.Attention(
        q_rot, k_rot, v, attention_bias, past_key, past_value,
        _allow_other_attributes=True,
        _outputs=["attn_out", "present_key", "present_value"],
    )
```

**What it checks** (`check()` at line 100):
- `scale`, `q_num_heads`, `kv_num_heads` attributes exist on the Attention node
- `cos`/`sin` are produced by `Gather` nodes (position-indexed cache tables)
- `past_key`/`past_value` are graph inputs (not computed — filters out vision encoder)

**What it produces** (`rewrite()` at line 132):
```python
outputs = op.op_multi_out(
    "GroupQueryAttention",
    inputs=[q_pre, k_pre, v, past_key, past_value,
            self._seqlens_k, self._total_seq_len,
            self._cos_cache, self._sin_cache],
    domain="com.microsoft",
    attributes={
        "num_heads": q_num_heads,
        "kv_num_heads": kv_num_heads,
        "scale": scale,
        "do_rotary": 1,
        "rotary_interleaved": 0,
    },
    num_outputs=3,
)
```

**Shared state** (computed once, reused across layers):
- `self._seqlens_k` = `Cast(ReduceSum(attention_mask, axis=1) - 1, INT32)` (line 167-174)
- `self._total_seq_len` = `Cast(Gather(Shape(attention_mask), 1), INT32)` (line 176-182)
- `self._cos_cache` / `self._sin_cache` = traced from cos/sin Gather inputs (line 154-156)

### Existing Tests: 7 passing

**File**: `src/mobius/rewrite_rules/_group_query_attention_test.py`

| Test | What it verifies |
|------|-----------------|
| `test_rules_returns_rule_set` | Returns `RewriteRuleSet` |
| `test_replaces_attention_with_gqa` | 28 Attention → 0, 28 GQA (Qwen3-0.6B) |
| `test_absorbs_rotary_embedding` | 56 RotaryEmbedding → 0 |
| `test_absorbs_rotary_without_qk_norm` | 30 Attention → 30 GQA (SmolLM2, no QK norm) |
| `test_preserves_non_matching_model` | Vision encoder Attention untouched |
| `test_rewritten_model_runs_with_ort` | Tiny model → GQA rewrite → ORT inference → correct output shape |
| `test_combined_gqa_and_skip_norm_runs_with_ort` | GQA + SkipNorm → valid ORT model |

### Graph Pattern (what the ONNX graph actually looks like per layer)

For a standard causal LM (e.g., llama, qwen2) built with `build_from_module()`:

```
hidden_states                           # [B, seq, hidden_size]
  ├─ Transpose → MatMul (q_proj)        # hidden → q_dim
  ├─ Transpose → MatMul (k_proj)        # hidden → kv_dim
  └─ Transpose → MatMul (v_proj)        # hidden → kv_dim
       │              │           │
  RotaryEmbedding  RotaryEmbedding  │
       │              │           │
       └──── Attention(q, k, v, attn_bias, past_k, past_v) ────→ out, present_k, present_v
                                  │
                     Transpose → MatMul (o_proj)
```

Note: The `Transpose` before each `MatMul` is because weights are stored transposed (fused_matmul rule can handle this separately). The `q_proj`, `k_proj`, `v_proj` are separate `Linear` modules producing separate `MatMul` nodes.

### Extend `RotaryAttentionToGQA` with Packed QKV

**Goal**: Modify the existing `RotaryAttentionToGQA` class in `src/mobius/rewrite_rules/_group_query_attention.py` so that its `rewrite()` method attempts to pack Q/K/V into a single fused QKV weight. If packing is not possible (QK norm present, bias complications, incompatible shapes), it falls back to the current separate Q/K/V behavior. Single rule, two paths — no new file, no new class.

**File to modify**: `src/mobius/rewrite_rules/_group_query_attention.py`

**Design — single rule, two code paths in `rewrite()`**:

The `pattern()` and `check()` methods stay the same — they already match the `RotaryEmbedding + Attention` pattern. The change is in `rewrite()`:

```python
def rewrite(self, op, q_pre, k_pre, v, attention_bias, past_key, past_value,
            cos, sin, attn_out, present_key, present_value, **_):
    # ... existing setup (get attributes, trace cos/sin, compute seqlens_k) ...

    # TRY packed QKV path
    packed_qkv = self._try_pack_qkv(op, q_pre, k_pre, v)
    if packed_qkv is not None:
        # Packed path: single QKV input, remaining Q/K/V slots are None
        outputs = op.op_multi_out(
            "GroupQueryAttention",
            inputs=[packed_qkv, None, None, past_key, past_value,
                    self._seqlens_k, self._total_seq_len,
                    self._cos_cache, self._sin_cache],
            domain="com.microsoft",
            attributes={...same attributes...},
            num_outputs=3,
        )
    else:
        # Fallback: separate Q, K, V (current behavior — unchanged)
        outputs = op.op_multi_out(
            "GroupQueryAttention",
            inputs=[q_pre, k_pre, v, past_key, past_value,
                    self._seqlens_k, self._total_seq_len,
                    self._cos_cache, self._sin_cache],
            domain="com.microsoft",
            attributes={...same attributes...},
            num_outputs=3,
        )
    return outputs[0], outputs[1], outputs[2]
```

**What `_try_pack_qkv()` must do** (new private method on the class):

1. **Trace back** from `q_pre`, `k_pre`, `v` to their producing MatMul ops
2. **Verify all 3 MatMuls** share the same first input (`hidden_states`)
3. **Verify no QK norm** — check that `q_pre` and `k_pre` come directly from MatMul, not through RMSNormalization/LayerNorm nodes
4. **Extract weight initializers** from all 3 MatMuls — they must be constant (`ir.Tensor`) values, not computed
5. **Handle Transpose** — weights may be transposed (Transpose → MatMul pattern from `Linear` module). Account for this when concatenating
6. **Concatenate weights**: `W_qkv = numpy.concatenate([W_q, W_k, W_v], axis=0)` (along output dim, accounting for transpose)
7. **Handle bias** if present: concatenate biases similarly, or return `None` to fall back
8. **Create packed MatMul**: single `MatMul(hidden, W_qkv)` (or `FusedMatMul` with `transB=1` if weights were transposed)
9. **Return** the packed QKV tensor, or `None` if packing failed at any step

**When packing should fail (return None)**:
- `q_pre`, `k_pre`, or `v` don't come from MatMul nodes
- The 3 MatMuls don't share the same input
- QK norm nodes exist between MatMul and RotaryEmbedding (e.g., Qwen3 with `attn_qk_norm=True`)
- Weight initializers are not available as constants
- Shape incompatibility

**Success criteria**:
```python
# After applying group_query_attention_rules() to a llama/qwen2 model (no QK norm):
assert count_ops(model).get("Attention", 0) == 0
assert count_ops(model).get("RotaryEmbedding", 0) == 0
assert count_ops(model)["GroupQueryAttention"] == num_layers
# MatMul count should decrease by 2 * num_layers (3 separate → 1 packed per layer)
assert count_ops(model)["MatMul"] == original_matmul_count - 2 * num_layers

# After applying group_query_attention_rules() to a qwen3 model (HAS QK norm):
# Packing fails gracefully → falls back to separate Q/K/V GQA (current behavior)
assert count_ops(model).get("Attention", 0) == 0
assert count_ops(model)["GroupQueryAttention"] == num_layers
# MatMul count stays the same (no packing occurred)
assert count_ops(model)["MatMul"] == original_matmul_count

# ALL 7 existing tests must continue to pass unchanged
# Model must still run with ORT and produce correct output shape
session = OnnxModelSession(model)
result = session.run(feeds)
assert result["logits"].shape == (batch, seq, vocab)
```

**Gotchas**:
- The `FusedMatMul` rule may have already replaced `Transpose+MatMul` with `FusedMatMul(transB=1)`. Your `_try_pack_qkv()` must handle BOTH cases: trace through `Transpose+MatMul` AND `FusedMatMul`. Or: document that GQA rule should be applied BEFORE fused_matmul (simpler approach).
- Weight concatenation must happen in the `rewrite()` method. Use `numpy` to concatenate the constant values. Create a new `ir.Tensor` for the packed weight.
- ORT's GQA expects packed QKV in layout `[q_dim + kv_dim + kv_dim]` (Q first, then K, then V). Verify this against the ORT spec.
- All 7 existing tests must still pass. The packed path is additive — it kicks in when conditions allow, falls back otherwise.
- No new rule class, no new rule file, no new registration in `__init__.py` or CLI `rule_map`. The rule name stays `group_query_attention`.

### Tests for Packed QKV (add to existing test file)

**File to modify**: `src/mobius/rewrite_rules/_group_query_attention_test.py`

**Add these test cases** to the existing test class:

```python
@pytest.mark.parametrize("model_type", ["llama", "qwen2", "phi3"])
def test_packed_qkv_on_all_compatible_models(self, model_type):
    """Build each model with tiny config, apply rule, verify packing succeeded:
    - 0 Attention ops remain
    - 0 RotaryEmbedding ops remain
    - GQA ops == num_layers
    - MatMul count decreased by 2 * num_layers (3 separate → 1 packed)
    Catches model-specific edge cases across architectures.
    """

def test_packed_weight_shape_is_correct(self):
    """Build tiny llama, apply rule, inspect the packed W_qkv initializer.
    - Verify shape is [hidden, q_dim + 2*kv_dim] (or transposed equivalent)
    - Verify the original 3 separate q_proj/k_proj/v_proj MatMul nodes are gone
    - Verify the packed weight contains the correct concatenation of Q, K, V weights
    """

def test_falls_back_to_separate_qkv_with_qk_norm(self):
    """Build qwen3 (attn_qk_norm=True), apply rule.
    QK norm prevents packing — rule falls back to separate Q/K/V.
    GQA ops should still be produced, but MatMul count unchanged.
    """

def test_packed_model_runs_with_ort(self):
    """Build tiny llama model, apply rule, fill weights, run with ORT.
    Verify output shape matches expected (batch, seq, vocab).
    """

def test_combined_packed_gqa_and_skip_norm(self):
    """Apply GQA (with packing) + skip_norm, verify ORT inference works."""
```

**Use existing test helpers** from `src/mobius/rewrite_rules/_testing_utils.py`:
- `count_ops(model)` — returns `dict[str, int]` of op type counts
- `fill_random_weights(model)` — fills all initializers with random data
- `make_prefill_feeds(session)` — creates prefill input feeds for ORT

## Key Files Map

| File | Lines | What to do |
|------|-------|-----------|
| `src/mobius/rewrite_rules/_group_query_attention.py` | 224 | **MODIFY** — add `_try_pack_qkv()` method, update `rewrite()` to try-pack-first |
| `src/mobius/rewrite_rules/_group_query_attention_test.py` | 156 | **MODIFY** — add 6 new test cases for packed QKV behavior (13 total) |
| `src/mobius/rewrite_rules/_testing_utils.py` | ~60 | READ ONLY — use `count_ops`, `fill_random_weights`, `make_prefill_feeds` |
| `.github/skills/writing-rewrite-rules/SKILL.md` | — | READ FIRST — rewrite rule conventions |

### Build Pipeline (for reference)

```
build("model_id")
  → HF config → registry lookup → module instantiation
  → build_from_module(module, config, task)
    → task.build() → graph construction (forward() builds ONNX nodes)
    → _optimize() → 5 default IR passes (2 iterations):
        IdentityEliminationPass, LiftConstantsToInitializersPass,
        CommonSubexpressionEliminationPass, RemoveUnusedNodesPass,
        RemoveUnusedOpsetsPass
  → set graph names → download + apply weights → return ModelPackage
```

### Rewrite Rule API (onnxscript.rewriter)

```python
class MyRule(RewriteRuleClassBase):
    def pattern(self, op, ...):
        """Return the output node(s) of the pattern to match."""
        return op.SomeOp(...)

    def check(self, context, **bindings) -> MatchResult:
        """Return MatchResult — call .fail('reason') to reject a match."""
        result = MatchResult()
        if bad_condition:
            return result.fail("why")
        return result

    def rewrite(self, op, **bindings):
        """Return replacement node(s). Same output count as pattern."""
        return op.op_multi_out("NewOp", inputs=[...], domain="com.microsoft", ...)

def my_rules() -> RewriteRuleSet:
    return RewriteRuleSet([MyRule().rule()])

# Usage:
from onnxscript.rewriter import rewrite
rewrite(model, pattern_rewrite_rules=my_rules())
```

### Static Cache — Why It's Out of Scope

Static cache (PR #9) produces this Attention pattern:
```
TensorScatter(key_cache, k_new, write_indices)
TensorScatter(value_cache, v_new, write_indices)
Attention(q, updated_k, updated_v, None, None, None, nonpad_kv_seqlen,
          is_causal=1, ...)
```

The GQA rewrite checks `past_key.producer() is not None` → fails (past_key is a TensorScatter output, not a graph input). Also, no `attention_mask` graph input exists to compute `seqlens_k`. Completely different graph topology — needs a separate rule if/when ORT GQA supports it.

## References

- [ORT GroupQueryAttention spec](https://github.com/microsoft/onnxruntime/blob/main/docs/ContribOperators.md) — search for `GroupQueryAttention`
- PR #9: Static cache support (StaticCacheCausalLMTask, StaticCacheState) — merged
- `.github/skills/writing-rewrite-rules/SKILL.md` — read before starting
- `onnxscript.rewriter` source: used for `rewrite()`, `RewriteRuleClassBase`, `RewriteRuleSet`, `MatchResult`

titaiwangms:
Fix #11

This pull request significantly enhances the GroupQueryAttention rewrite rule by introducing support for packing Q, K, and V projections into a single MatMul when possible, improving model efficiency. It also adds comprehensive tests to verify both the new packed QKV path and fallback behavior for models that cannot be packed (such as those with QK norm, e.g., Qwen3). The documentation and test infrastructure are updated to reflect these changes.

Key changes:

**GroupQueryAttention rewrite rule improvements:**

* Added logic to detect when Q, K, and V projections can be packed into a single MatMul by checking for shared input and absence of QK norm, then concatenating the weights and emitting a packed MatMul for GroupQueryAttention. The fallback path for models with QK norm remains unchanged. [[1]](diffhunk://#diff-05ea0a505b5dca0ba7c7a7a17ba02bc4c50a211a2af554617c600e17e463481cR13-R19) [[2]](diffhunk://#diff-05ea0a505b5dca0ba7c7a7a17ba02bc4c50a211a2af554617c600e17e463481cR179-R235) [[3]](diffhunk://#diff-05ea0a505b5dca0ba7c7a7a17ba02bc4c50a211a2af554617c600e17e463481cL184-R305) [[4]](diffhunk://#diff-05ea0a505b5dca0ba7c7a7a17ba02bc4c50a211a2af554617c600e17e463481cL197-R320)
* Updated documentation to describe the packed QKV optimization and fallback path, including code snippets for both scenarios. [[1]](diffhunk://#diff-05ea0a505b5dca0ba7c7a7a17ba02bc4c50a211a2af554617c600e17e463481cR13-R19) [[2]](diffhunk://#diff-05ea0a505b5dca0ba7c7a7a17ba02bc4c50a211a2af554617c600e17e463481cL48-L53)

**Helper and utility additions:**

* Introduced `_extract_weight` helper function to robustly extract and handle weight matrices from MatMul nodes, supporting both direct and transposed weights.

**Testing enhancements:**

* Added new test cases for the packed QKV path, including tests that verify MatMul reduction, correct packed weight shape, fallback behavior for QK norm models, and end-to-end execution with ONNX Runtime. [[1]](diffhunk://#diff-63c5e9ca11114f2bbd3649ba2e0af172f03e4c4cd3c357523380696ede6aeb66R22-R56) [[2]](diffhunk://#diff-63c5e9ca11114f2bbd3649ba2e0af172f03e4c4cd3c357523380696ede6aeb66L86-R123) [[3]](diffhunk://#diff-63c5e9ca11114f2bbd3649ba2e0af172f03e4c4cd3c357523380696ede6aeb66L121-R250)
* Refactored test configuration setup for clarity and reusability by introducing `_LLAMA_CONFIG` and `_QWEN3_CONFIG`.
* Updated existing tests to use the new configuration constants for consistency. [[1]](diffhunk://#diff-63c5e9ca11114f2bbd3649ba2e0af172f03e4c4cd3c357523380696ede6aeb66L86-R123) [[2]](diffhunk://#diff-63c5e9ca11114f2bbd3649ba2e0af172f03e4c4cd3c357523380696ede6aeb66L121-R250)

### DIFF
--- a/src/mobius/rewrite_rules/_group_query_attention.py
+++ b/src/mobius/rewrite_rules/_group_query_attention.py
@@ -293,7 +293,11 @@ class PackQKVForGQA(RewriteRuleClassBase):
     data files for large models.
     """
 
-    _pack_counter: int = 0
+    _pack_counter: int
+
+    def __init__(self):
+        super().__init__()
+        self._pack_counter = 0
 
     # ------------------------------------------------------------------ pattern
 
--- a/src/mobius/rewrite_rules/_group_query_attention.py
+++ b/src/mobius/rewrite_rules/_group_query_attention.py
@@ -36,7 +36,6 @@ from onnxscript.rewriter._rewrite_rule import (
     RewriteRuleClassBase,
     RewriteRuleSet,
 )
-from onnxscript.rewriter.pattern import OrValue
 
 
 class RotaryAttentionToGQA(RewriteRuleClassBase):
@@ -226,26 +225,19 @@ class RotaryAttentionToGQA(RewriteRuleClassBase):
 
 
 def _get_weight_tensor(proj_node):
-    """Extract the constant weight tensor from a projection node.
+    """Extract the constant weight tensor from a MatMul projection node.
 
-    Handles three patterns:
+    Handles two patterns:
 
-    1. ``Transpose(weight, perm=[1,0]) -> MatMul(x, w_t)``
+    1. ``Transpose(weight, perm=[1,0]) → MatMul(x, w_t)``
+       — weight shape is ``(out_features, hidden_size)``, transposed=True
     2. ``MatMul(x, weight)``  (no transpose)
-    3. ``FusedMatMul(x, weight, transB=1)``  (post fused_matmul rewrite)
+       — weight shape is ``(hidden_size, out_features)``, transposed=False
 
     Returns:
         ``(numpy_array, is_transposed)`` on success, or ``(None, False)``
         if the weight cannot be extracted as a constant.
     """
-    if proj_node.op_type == "FusedMatMul":
-        trans_b = proj_node.attributes.get_int("transB", 0)
-        tensor = ir.convenience.get_const_tensor(proj_node.inputs[1])
-        if tensor is None:
-            return None, False
-        return tensor.numpy(), bool(trans_b)
-
-    # MatMul -- weight may be behind a Transpose
     weight_input = proj_node.inputs[1]
     producer = weight_input.producer()
     if producer is not None and producer.op_type == "Transpose":
@@ -267,29 +259,33 @@ class PackQKVForGQA(RewriteRuleClassBase):
 
     This rule runs **after** ``RotaryAttentionToGQA`` and looks for
     ``GroupQueryAttention`` nodes whose Q, K, V inputs each come from a
-    separate projection (``MatMul`` or ``FusedMatMul``) that shares
-    the same ``hidden_states`` input.
+    separate ``MatMul`` projection that shares the same ``hidden_states``
+    input.  The ``fused_matmul`` rewrite must run **after** this rule
+    so that projections are still plain ``MatMul`` nodes when this rule
+    matches.
 
     **Matched pattern:**
 
     .. code-block:: text
 
-        q = MatMul|FusedMatMul(hidden, W_q)
-        k = MatMul|FusedMatMul(hidden, W_k)
-        v = MatMul|FusedMatMul(hidden, W_v)
+        q = MatMul(hidden, W_q)
+        k = MatMul(hidden, W_k)
+        v = MatMul(hidden, W_v)
         out, pkey, pval = GroupQueryAttention(q, k, v, ...)
 
     **Replacement:**
 
     .. code-block:: text
 
-        W_qkv = concatenate([W_q, W_k, W_v])
-        packed = MatMul(hidden, W_qkv_t)
+        W_qkv = concatenate([W_q, W_k, W_v])  # normalized to (out, hidden)
+        packed = MatMul(hidden, Transpose(W_qkv))
         out, pkey, pval = GroupQueryAttention(packed, None, None, ...)
 
-    The packed weight is stored as a graph initializer (not a
-    ``Constant`` node attribute) so it can be serialised to external
-    data files for large models.
+    Each weight is independently normalized to ``(out_features,
+    hidden_size)`` before concatenation, so mixed transpose patterns
+    across Q/K/V are handled correctly.  The packed weight is stored as
+    a graph initializer (not a ``Constant`` node attribute) so it can
+    be serialised to external data files for large models.
     """
 
     _pack_counter: int
@@ -314,40 +310,9 @@ class PackQKVForGQA(RewriteRuleClassBase):
         cos_cache,
         sin_cache,
     ):
-        # Each projection may be MatMul or FusedMatMul (post fused_matmul)
-        q = OrValue(
-            [
-                op.MatMul(hidden, q_w),
-                op.FusedMatMul(
-                    hidden,
-                    q_w,
-                    _domain="com.microsoft",
-                    _allow_other_attributes=True,
-                ),
-            ]
-        )
-        k = OrValue(
-            [
-                op.MatMul(hidden, k_w),
-                op.FusedMatMul(
-                    hidden,
-                    k_w,
-                    _domain="com.microsoft",
-                    _allow_other_attributes=True,
-                ),
-            ]
-        )
-        v = OrValue(
-            [
-                op.MatMul(hidden, v_w),
-                op.FusedMatMul(
-                    hidden,
-                    v_w,
-                    _domain="com.microsoft",
-                    _allow_other_attributes=True,
-                ),
-            ]
-        )
+        q = op.MatMul(hidden, q_w)
+        k = op.MatMul(hidden, k_w)
+        v = op.MatMul(hidden, v_w)
 
         return op.GroupQueryAttention(
             q,
@@ -366,28 +331,17 @@ class PackQKVForGQA(RewriteRuleClassBase):
 
     # ------------------------------------------------------------------ check
 
-    def check(self, context, q_w, k_w, v_w, gqa_out, **_):
+    def check(self, context, gqa_out, **_):
         result = MatchResult()
 
         gqa_node = gqa_out.producer()
-        # Extract projection nodes from the GQA inputs
-        projs = []
         for i, name in enumerate(("q", "k", "v")):
             proj = gqa_node.inputs[i].producer()
             if proj is None:
                 return result.fail(f"{name} projection missing")
-            projs.append(proj)
-
-        # All weights must be extractable constants with consistent layout
-        is_transposed = None
-        for name, proj in zip(("q", "k", "v"), projs):
-            w_np, transposed = _get_weight_tensor(proj)
+            w_np, _ = _get_weight_tensor(proj)
             if w_np is None:
                 return result.fail(f"{name} weight not constant")
-            if is_transposed is None:
-                is_transposed = transposed
-            elif is_transposed != transposed:
-                return result.fail("Mixed transpose states")
 
         return result
 
@@ -397,9 +351,6 @@ class PackQKVForGQA(RewriteRuleClassBase):
         self,
         op,
         hidden,
-        q_w,
-        k_w,
-        v_w,
         past_key,
         past_value,
         seqlens_k,
@@ -414,22 +365,17 @@ class PackQKVForGQA(RewriteRuleClassBase):
         gqa_node = gqa_out.producer()
         graph = gqa_node.graph
 
-        q_proj = gqa_node.inputs[0].producer()
-
-        # Extract and concatenate weights
-        weights_np = []
-        is_transposed = None
+        # Extract weights, normalizing each to (out_features, hidden_size)
+        weights = []
         for i in range(3):
             proj = gqa_node.inputs[i].producer()
             w_np, transposed = _get_weight_tensor(proj)
-            if is_transposed is None:
-                is_transposed = transposed
-            weights_np.append(w_np)
+            if not transposed:
+                w_np = w_np.T
+            weights.append(w_np)
 
-        # Transposed: (out_features, hidden) -> concat axis=0
-        # Non-transposed: (hidden, out_features) -> concat axis=1
-        concat_axis = 0 if is_transposed else 1
-        w_qkv_np = np.concatenate(weights_np, axis=concat_axis)
+        # Concatenate along axis=0: all weights are (out, hidden)
+        w_qkv_np = np.concatenate(weights, axis=0)
 
         # Store packed weight as a graph initializer
         self._pack_counter += 1
@@ -440,20 +386,9 @@ class PackQKVForGQA(RewriteRuleClassBase):
         )
         graph.register_initializer(packed_w)
 
-        # Create the packed projection
-        if is_transposed:
-            if q_proj.op_type == "FusedMatMul":
-                packed_qkv = op.op(
-                    "FusedMatMul",
-                    inputs=[hidden, packed_w],
-                    domain="com.microsoft",
-                    attributes={"transB": 1},
-                )
-            else:
-                packed_w_t = op.Transpose(packed_w, perm=[1, 0])
-                packed_qkv = op.MatMul(hidden, packed_w_t)
-        else:
-            packed_qkv = op.MatMul(hidden, packed_w)
+        # Transpose + MatMul for the packed projection
+        packed_w_t = op.Transpose(packed_w, perm=[1, 0])
+        packed_qkv = op.MatMul(hidden, packed_w_t)
 
         # Forward all original GQA attributes
         attrs = {key: gqa_node.attributes[key].value for key in gqa_node.attributes}
--- a/src/mobius/rewrite_rules/_group_query_attention.py
+++ b/src/mobius/rewrite_rules/_group_query_attention.py
@@ -21,11 +21,42 @@ These rules are **not applied by default**.  Apply them post-export::
 
 from __future__ import annotations
 
+import numpy as np
+import onnx_ir as ir
 from onnxscript.rewriter._basics import MatchResult
 from onnxscript.rewriter._rewrite_rule import (
     RewriteRuleClassBase,
     RewriteRuleSet,
 )
+
+
+def _extract_weight(matmul_node):
+    """Extract the weight numpy array from a MatMul node.
+
+    Handles the ``Transpose(weight, perm=[1,0]) → MatMul`` pattern
+    commonly emitted by ``Linear`` layers.
+
+    Returns:
+        A tuple ``(numpy_array, is_transposed)`` where
+        ``is_transposed`` is ``True`` when the weight was preceded by a
+        ``Transpose``.  Returns ``(None, False)`` if the weight cannot
+        be extracted (e.g. not a constant initializer).
+    """
+    weight_input = matmul_node.inputs[1]
+    producer = weight_input.producer()
+
+    if producer is not None and producer.op_type == "Transpose":
+        perm = producer.attributes.get("perm", None)
+        if perm is not None and list(perm.value) == [1, 0]:
+            raw_weight = producer.inputs[0]
+            if raw_weight.const_value is None:
+                return None, False
+            return raw_weight.const_value.numpy(), True
+
+    # Direct weight (no Transpose)
+    if weight_input.const_value is None:
+        return None, False
+    return weight_input.const_value.numpy(), False
 
 
 class RotaryAttentionToGQA(RewriteRuleClassBase):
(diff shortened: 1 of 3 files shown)


src/mobius/rewrite_rules/_group_query_attention.py
```diff
@@ -127,6 +176,63 @@ def check(self, context, attn_out, cos, sin, past_key, past_value, **_):
 
         return result
 
+    # ------------------------------------------------------------ pack QKV
+
+    def _try_pack_qkv(self, op, q_pre, k_pre, v):
```
gramalingam: A couple of suggestions (for copilot):
* Packing can be a separate standalone rule, instead of being fused into the GQA rule. Specifically, we look for a GQA op with separate Q, K, and V that enables packing and pack it.
* And, furthermore, copilot is not using the pattern-matching abilities of the rewriter ... the logic below can be expressed better as a pattern. That leads to more readable rules, easier to verify.

gramalingam:
Would be useful to think about sharing of rewrite-rules between the exporter and model-builder ... do we want to duplicate rules like this in the onnxscript repo and here? May be better if we could find a way to have them in one place.

src/mobius/rewrite_rules/_group_query_attention.py
titaiwangms: I actually told Copilot to extend GQA rule instead of making a new one to avoid conflict. Do you think a two-step conversion of GQA is better?

gramalingam: I think the 2-step conversion is better (it has some perf overhead, but is more general and easier to verify/read etc.)

titaiwangms: Agree

src/mobius/rewrite_rules/_group_query_attention.py
```diff
@@ -332,15 +219,259 @@ def rewrite(
...
+        v,
+        past_key,
+        past_value,
+        seqlens_k,
+        total_seq_len,
+        cos_cache,
+        sin_cache,
+    ):
+        return op.GroupQueryAttention(
+            q,
```
gramalingam, line 318: I am wondering whether it would be better to do the following. We can capture more of the pattern right here declaratively: basically say that q (and k and v) must be either a MatMul or FusedMatMul using the OrValue pattern, as below:
```py
   q_matmul = op.MatMul(hidden, q_wt)
   q_fused_matmul = op.FusedMatMul(hidden, q_wt, _domain="com.microsoft") # may need something for attributes
   q = pattern.OrValue([q_matmul, q_fused_matmul])
```

gramalingam: My attribute comment is that if you want the attribute value for further checks, you can get that too via the pattern, allowing for no attribute using the can_match_none flag

titaiwangms: Make sense

src/mobius/rewrite_rules/_group_query_attention.py
```diff
@@ -422,15 +445,15 @@ def rewrite(
...
                     "FusedMatMul",
-                    inputs=[hidden_states, packed_w],
+                    inputs=[hidden, packed_w],
                     domain="com.microsoft",
                     attributes={"transB": 1},
                 )
             else:
                 packed_w_t = op.Transpose(packed_w, perm=[1, 0])
-                packed_qkv = op.MatMul(hidden_states, packed_w_t)
+                packed_qkv = op.MatMul(hidden, packed_w_t)
```
gramalingam: I am trying to understand: do we want to say that all Q/K/V must be same kind (that is, all use matmul or all use fused-matmul)? If it is matmul, do we really care whether the weight input is transposed or not?

gramalingam: I wonder if we should delay the FusedMatMul fusion rule until after all "Packed matmul" rules are applied.

titaiwangms: Yeah, let me check if FusedMatMul was fused early for a reason.

src/mobius/rewrite_rules/_group_query_attention.py
```diff
@@ -209,15 +219,217 @@ def rewrite(
...
+        self._pack_counter += 1
+        w_name = f"packed_qkv_weight_{self._pack_counter}"
+        packed_w = ir.Value(
+            name=w_name,
+            const_value=ir.Tensor(w_qkv_np, name=w_name),
+        )
+        graph.register_initializer(packed_w)
+
+        # Transpose + MatMul for the packed projection
+        packed_w_t = op.Transpose(packed_w, perm=[1, 0])
```
gramalingam, line 390: I am a bit confused: why do we transpose earlier (line 374), and transpose back here? Is this to enable some later fusion/rewrite?

src/mobius/rewrite_rules/_group_query_attention.py
```diff
@@ -209,15 +219,217 @@ def rewrite(
...
+        )
+
+    # ------------------------------------------------------------------ check
+
+    def check(self, context, gqa_out, **_):
+        result = MatchResult()
+
+        gqa_node = gqa_out.producer()
+        for i, name in enumerate(("q", "k", "v")):
+            proj = gqa_node.inputs[i].producer()
```
gramalingam, line 339: Nit: this logic is duplicated across this check method and rewrite. I would simply drop the check method, and do everything in rewrite to avoid the duplication.

src/mobius/rewrite_rules/_group_query_attention.py
```diff
@@ -209,15 +219,217 @@ def rewrite(
...
+        graph = gqa_node.graph
+
+        # Extract weights, normalizing each to (out_features, hidden_size)
+        weights = []
+        for i in range(3):
+            proj = gqa_node.inputs[i].producer()
+            w_np, transposed = _get_weight_tensor(proj)
+            if not transposed:
+                w_np = w_np.T
+            weights.append(w_np)
```
gramalingam, line 375: I would alter _get_weight_tensor and write this as:
```py
   q_w_np = _get_weight_tensor(q_w, "q_w")
   k_w_np = _get_weight_tensor(k_w, "k_w")
   ...
```
This requires using q_w from the pattern in parameter list above. Function _get_weight_tensor can throw a MatchFailureError exception to indicate failure (eg., see [this line](https://github.com/microsoft/onnxscript/blob/19e5284a705f01a9c2e6f6f83bba30388ef2e253/onnxscript/rewriter/_fusion_utils.py#L31)).

gramalingam (approved):
Feel free to merge this. I can work with copilot to update it for style/readability

PR #13, opened 2026-03-20
Add --device option to qwen35_text_generation.py for CUDA support

justinchuby:
Adds a `--device` CLI option to `examples/qwen35_text_generation.py` supporting `cpu` and `cuda` for both ONNX Runtime and PyTorch inference.

## Changes
- New `--device` argparse argument with choices=['cpu', 'cuda'], default='cpu'
- Device forwarded to all 4 `OnnxModelSession` instantiations via `**load_kwargs`
- Device parameter added to HF comparison functions replacing hardcoded `.to('cpu')`
- Docstring updated with `--device cuda` usage example

## Bug fix
- Fixed LinearAttention CUDA crash caused by dynamic Shape→Gather→Div→Tile computing GQA head expansion ratio at runtime. ORT's CUDA EP misplaced the int64 shape tensors, causing ~422TB allocation attempts.
- Replaced with static build-time ratio computation: ratio==1 skips Tile entirely, ratio>1 uses Constant repeats
- Fixed Reshape 5D→4D shape bug (ONNX 0-sentinel copies from input positional index, not output)
- Added divisibility guard for num_v_heads % num_k_heads

### DIFF
--- a/src/mobius/functions/linear_attention.py
+++ b/src/mobius/functions/linear_attention.py
@@ -121,10 +121,12 @@ def linear_attention(
     op = gb.op
 
     # --- GQA: expand Q/K heads to match V head count ---
+    if num_v_heads % num_k_heads != 0:
+        raise ValueError(
+            f"num_v_heads ({num_v_heads}) must be divisible by num_k_heads ({num_k_heads})"
+        )
     gqa_ratio = num_v_heads // num_k_heads
-    query_expanded, key_expanded = _expand_kv_heads(
-        op, query, key, gqa_ratio=gqa_ratio
-    )
+    query_expanded, key_expanded = _expand_kv_heads(op, query, key, gqa_ratio=gqa_ratio)
 
     # --- Build Scan for sequential recurrence ---
     scan_body = _build_recurrence_body(uses_decay, uses_beta)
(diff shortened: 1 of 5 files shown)


src/mobius/functions/linear_attention.py
```diff
@@ -269,50 +278,52 @@ def _build_recurrence_body(
...
+    # We cannot use '0' sentinels because ONNX Reshape copies from the same
+    # positional index, and the 5D→4D dimension mapping doesn't align.
     b_dim = op.Shape(query, start=0, end=1)
     t_dim = op.Shape(query, start=2, end=3)
-    d_k_dim = op.Shape(query, start=3, end=4)
+    dk_dim = op.Shape(query, start=3, end=4)
     expanded_shape = op.Concat(
         b_dim,
-        op.Reshape(h, op.Constant(value_ints=[1])),
+        op.Constant(value_ints=[-1]),
```
gramalingam, line 324: nit: i guess `[-1]` would work as well?

gramalingam (approved):

src/mobius/functions/linear_attention.py
justinchuby: Most likely. I can do a batch clean up

PR #19, opened 2026-03-20
Refactor data type handling and improve model precision in Qwen3.5

tadani3:
This pull request improves precision handling and numerical stability for model inference and component computations, especially around ONNX export and attention mechanisms. It introduces support for configurable dtype (float16/float32) throughout the pipeline and ensures all mathematical operations and intermediate values are consistently cast to the appropriate precision, reducing the risk of numerical errors.

**Precision and dtype handling improvements:**

* Added a `--dtype` command-line argument to `examples/qwen35_text_generation.py`, allowing users to select between float16 (`f16`) and float32 (`f32`) precision for ONNX model export and inference. The selected dtype is now propagated to all relevant build and model instantiation functions.
* In `src/mobius/components/_gated_deltanet.py`, all bias, scaling, and normalization constants are now explicitly cast to match the precision of model weights or activations using `op.CastLike`, ensuring consistent dtype usage even when running in mixed precision.

**Numerical stability in attention computation:**

* The attention mechanism in `src/mobius/components/_gated_deltanet.py` now casts all relevant activations and parameters to float32 before performing operations prone to underflow/overflow (such as exponentiation and softplus), then casts results back to the model's working precision. This mirrors best practices in the reference Hugging Face implementation and prevents numerical instability, especially in float16 mode.

**ONNX function definition consistency:**

* The ONNX function for causal convolution (`src/mobius/functions/causal_conv.py`) no longer sets the `overload` attribute in its definition, ensuring compatibility with call sites and preventing issues during serialization.

**Notes for the Team**

To test that the changes work in FP16, use the following command. I was able to get the ONNX model producing identical results to HF. If there are questions let me know:

python examples/qwen35_text_generation.py --dtype f16  --model Qwen/Qwen3.5-0.8B --compare-hf

### DIFF
--- a/src/mobius/components/_gated_deltanet.py
+++ b/src/mobius/components/_gated_deltanet.py
@@ -234,34 +234,29 @@ class GatedDeltaNet(nn.Module):
         g = op.Mul(neg_a, softplus_val)  # (B, S, num_v_heads), float32
 
         # === LinearAttention ===
-        # The LinearAttention ir.Function declares all inputs as FLOAT.
-        # Cast activations to float32 before the call; cast outputs back.
-        # Transpose from (B, S, H, D) to (B, H, S, D)
+        # The LinearAttention function handles float32 casting internally.
+        # Transpose from (B, S, H, D) to (B, H, S, D).
         # Q/K keep native num_k_heads; V uses num_v_heads.
         # GQA expansion happens inside the function.
-        query_bhsd = op.Cast(op.Transpose(query, perm=[0, 2, 1, 3]), to=ir.DataType.FLOAT)
-        key_bhsd = op.Cast(op.Transpose(key, perm=[0, 2, 1, 3]), to=ir.DataType.FLOAT)
-        value_bhsd = op.Cast(op.Transpose(value, perm=[0, 2, 1, 3]), to=ir.DataType.FLOAT)
-        recurrent_state_f32 = op.Cast(recurrent_state, to=ir.DataType.FLOAT)
+        query_bhsd = op.Transpose(query, perm=[0, 2, 1, 3])
+        key_bhsd = op.Transpose(key, perm=[0, 2, 1, 3])
+        value_bhsd = op.Transpose(value, perm=[0, 2, 1, 3])
 
         # beta/decay: (B, S, H) -> (B, H, S)
-        beta_bhs = op.Cast(op.Transpose(beta, perm=[0, 2, 1]), to=ir.DataType.FLOAT)
-        g_bhs = op.Transpose(g, perm=[0, 2, 1])  # already float32
+        beta_bhs = op.Transpose(beta, perm=[0, 2, 1])
+        g_bhs = op.Transpose(g, perm=[0, 2, 1])  # float32 from decay computation
 
         output_4d, new_recurrent_state = op.LinearAttention(
             query_bhsd,  # (B, H_kv, S, d_k)
             key_bhsd,  # (B, H_kv, S, d_k)
             value_bhsd,  # (B, H, S, d_v)
-            recurrent_state_f32,  # (B, H, d_k, d_v)
-            g_bhs,  # (B, H, S) — decay in log-space
+            recurrent_state,  # (B, H, d_k, d_v)
+            g_bhs,  # (B, H, S) — decay in log-space, float32
             beta_bhs,  # (B, H, S) — update rate
             update_rule="gated_delta",
             _domain="com.microsoft",
             _outputs=2,
         )
-        # Cast outputs back to model precision (no-op for f32)
-        output_4d = op.CastLike(output_4d, hidden_states)
-        new_recurrent_state = op.CastLike(new_recurrent_state, hidden_states)
         # output_4d: (B, H, S, d_v) -> (B, S, H, d_v)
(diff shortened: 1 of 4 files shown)


justinchuby (commented):
Rebase? I think main contains a part of the changes here.

titaiwangms (approved):

src/mobius/components/_gated_deltanet.py
```diff
@@ -223,35 +223,45 @@ def forward(
...
         # === LinearAttention ===
+        # The LinearAttention ir.Function declares all inputs as FLOAT.
+        # Cast activations to float32 before the call; cast outputs back.
         # Transpose from (B, S, H, D) to (B, H, S, D)
         # Q/K keep native num_k_heads; V uses num_v_heads.
         # GQA expansion happens inside the function.
-        query_bhsd = op.Transpose(query, perm=[0, 2, 1, 3])
-        key_bhsd = op.Transpose(key, perm=[0, 2, 1, 3])
-        value_bhsd = op.Transpose(value, perm=[0, 2, 1, 3])
+        query_bhsd = op.Cast(op.Transpose(query, perm=[0, 2, 1, 3]), to=ir.DataType.FLOAT)
```
gramalingam: Hmmm ... why should attention be done in full-precision? The comment a few lines above makes me wonder if if it is the LinearAttention function that should be generalized to support different types (or at least its type annotation)?

gramalingam: I think we should mimic hf/pytorch model behavior for precision of ops. May be worth checking (or asking copilot to check, which might be better) for this part.

gramalingam (approved):

src/mobius/components/_gated_deltanet.py
```diff
@@ -239,14 +244,14 @@ def forward(
...
-        g_bhs = op.Transpose(g, perm=[0, 2, 1])
+        g_bhs = op.Transpose(g, perm=[0, 2, 1])  # float32 from decay computation
 
         output_4d, new_recurrent_state = op.LinearAttention(
             query_bhsd,  # (B, H_kv, S, d_k)
             key_bhsd,  # (B, H_kv, S, d_k)
             value_bhsd,  # (B, H, S, d_v)
             recurrent_state,  # (B, H, d_k, d_v)
-            g_bhs,  # (B, H, S) — decay in log-space
+            g_bhs,  # (B, H, S) — decay in log-space, float32
```
justinchuby, line 249: So this is always float32?

justinchuby:
TODO: Run the example script

src/mobius/components/_gated_deltanet.py
tadani3: Seems like it. Copilot suggests that this line mimics the following one in the Transformers file: [https://github.com/huggingface/transformers/blob/main/src/transformers/models/qwen3_5/modeling_qwen3_5.py#L338C1-L340C6].

There's an explicit cast to Float32 there that needs to be accounted for.

tadani3:
> TODO: Run the example script

Ran with latest changes on tommasoadani/f16_debugging, got this traceback:

onnxruntime.capi.onnxruntime_pybind11_state.InvalidArgument: [ONNXRuntimeError] : 2 : INVALID_ARGUMENT : Unexpected input data type. Actual: (tensor(float)) , expected: (tensor(float16))

justinchuby (approved):

justinchuby:
@tadani3 can you fix lint?

src/mobius/functions/linear_attention.py
```diff
@@ -128,19 +111,30 @@ def linear_attention(
...
+    # --- Cast to float32 for Scan recurrence precision ---
+    # The recurrence requires float32 for numerical stability; inputs may be
+    # float16 or bfloat16.  decay is already FLOAT (declared above), so its
+    # Cast is always a no-op — kept for uniformity with the other inputs.
+    query_f32 = op.Cast(query_expanded, to=ir.DataType.FLOAT)
+    key_f32 = op.Cast(key_expanded, to=ir.DataType.FLOAT)
+    value_f32 = op.Cast(value, to=ir.DataType.FLOAT)
+    state_f32 = op.Cast(past_state, to=ir.DataType.FLOAT)
+    decay_f32 = op.Cast(decay, to=ir.DataType.FLOAT)  # no-op: decay is always FLOAT
+    beta_f32 = op.Cast(beta, to=ir.DataType.FLOAT)
```
justinchuby, line 128: I wonder if this is making sense in meaningfully using lower precision? @tadani3 @gramalingam @titaiwangms @guschmue

justinchuby:
Also add bfloat16 to the list?

src/mobius/functions/linear_attention.py
guschmue: casting QKV, state, decay, beta is somewhat odd - what is the point of fp16 then?
Like state - that is huge, you cannot cast this to fp32.
One thing to realize is that these days nobody trains a model on fp32 - it is bfloat16 at a maximum.
There can't not be much need for fp32 since the original model itself is already fp16.

I have a simple script to fix the qwen3.5 graph after mobius so I can run the model with fp16/int4 on webgpu without any fp32, seems fine. But of course, only a full eval can tell.

justinchuby:
I will merge this and create a follow up

PR #23, opened 2026-03-23
Add support for Nemotron models

gramalingam:
* Add support for Nemotron models
* Update skills for dtype handling

(Created by copilot. Reviewed it. Tested fp32 and fp16 models on CPU and CUDA EPs. CUDA requires a workaround for RMSNorm since the kernel doesn't seem to handle 2D scale tensor. Produces reasonably similar answers as the HF version, but not identical due to numeric differences.)

### DIFF
--- a/.github/skills/adding-a-new-model/SKILL.md
+++ b/.github/skills/adding-a-new-model/SKILL.md
@@ -748,15 +748,15 @@ match HuggingFace, which explicitly upcasts with `.float()` /
 
 1. **Naive `CastLike` everywhere** — keeps everything in the model dtype
    (e.g. bf16), but `exp` overflows and the SSM state diverges.
-2. **Naive `Cast(to=1)` everywhere** — computes in fp32 but forgets to cast
+2. **Naive `Cast(to=ir.DataType.FLOAT)` everywhere** — computes in fp32 but forgets to cast
    back, producing type mismatches with downstream bf16 ops.
 
 **Correct pattern — upcast → compute → cast back:**
 ```python
 # 1. Upcast to fp32 for the sensitive region
-dt_f32 = op.Cast(dt, to=1)
+dt_f32 = op.Cast(dt, to=ir.DataType.FLOAT)
 dt_f32 = op.Softplus(dt_f32)
-a_neg = op.Neg(op.Exp(op.Cast(self.A_log, to=1)))
+a_neg = op.Neg(op.Exp(op.Cast(self.A_log, to=ir.DataType.FLOAT)))
 da = op.Exp(op.Mul(dt_4d, a_4d))  # all fp32 here
 ...
 # 2. Cast back to input dtype at the boundary
@@ -772,7 +772,7 @@ that the ONNX graph must replicate.
 
 | Region | HF evidence | ONNX pattern |
 |--------|-------------|-------------|
-| SSM recurrence (A, dt, exp, state) | `self.A_log.float()`, `hidden_states.float()`, `B.float()`, `C.float()` | `Cast(to=1)` all inputs, `CastLike` output |
+| SSM recurrence (A, dt, exp, state) | `self.A_log.float()`, `hidden_states.float()`, `B.float()`, `C.float()` | `Cast(to=ir.DataType.FLOAT)` all inputs, `CastLike` output |
 | GatedRMSNorm (SiLU + variance) | `hidden_states.to(torch.float32)`, `gate.to(torch.float32)` | Explicit fp32 for both, `CastLike` output |
 | RMSNorm variance | `hidden_states.to(torch.float32)` | ONNX `RMSNormalization` handles via `stash_type=1` (default) |
 
@@ -783,7 +783,7 @@ that the ONNX graph must replicate.
 
 **Use `CastLike` for** parameters/constants that should match the *current*
 compute dtype (which is fp32 inside an upcast region, or the model dtype
-outside).  Use `Cast(to=1)` to explicitly enter an fp32 region.
+outside).  Use `Cast(to=ir.DataType.FLOAT)` to explicitly enter an fp32 region.
 
 ## Reference implementations
 
--- a/.github/skills/adding-a-new-model/SKILL.md
+++ b/.github/skills/adding-a-new-model/SKILL.md
@@ -733,42 +733,57 @@ for i in range(num_groups - 1):
     codec_sum += cp_embed[i, codes[i + 1], :]
 ```
 
-### 16. Hardcoded `Cast(to=float32)` breaks bfloat16/float16 models
+### 16. Precision-sensitive ops need fp32 upcast
 
-**Symptom:** `ONNXRuntimeError: Type parameter (T) of Optype (Mul) bound to
-different types (tensor(float) and tensor(bfloat16))` when loading a model
-built with `--dtype bf16` or from a bfloat16 checkpoint.
+**Symptom:** Type mismatch errors (`tensor(float) vs tensor(bfloat16)`)
+when loading a model built with `--dtype bf16`, or numerical drift compared
+to HuggingFace when running in fp16/bf16.
 
-**Root cause:** Component code uses `op.Cast(self.param, to=1)` (hardcoded
-to float32) to cast learned parameters before arithmetic.  When the graph is
-built in bfloat16, intermediate tensors are bf16, but the Cast forces float32,
-producing a type mismatch in the downstream `Mul`/`Add`.
+**Root cause:** Operations like `exp`, `softplus`, `sigmoid` (in gated norms),
+and RMSNorm variance are numerically sensitive and must run in float32 to
+match HuggingFace, which explicitly upcasts with `.float()` /
+`.to(torch.float32)`.
 
-**Example (bad — from SSM components):**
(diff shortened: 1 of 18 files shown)


gramalingam:
Per copilot investigation, the CI failures are
```
 1. Windows _weight_loading_test.py (3 failures per platform): UnicodeDecodeError — .read_text() without
encoding="utf-8" fails on Windows charmap. This affects ALL source files with Unicode (→, —), not just ours.
 2. Windows synthetic parity + rewrite rules + build_graph_test (many failures): Opset 24 is under development —
ORT on Windows CI doesn't support opset 24 yet. Failing on llama, mistral, qwen2, olmo, nemotron, etc. — core
models, not just ours.
 3. Windows arch_validation_test.py: No module named 'resource' — resource is Unix-only.
 4. Ubuntu jobs: All killed with SIGTERM (exit code 143 / "runner received shutdown signal") before completion —
likely a CI timeout or runner recycling issue.
 5. Integration tests: Same runner shutdown — tests were running but got terminated.
 ```
Will try merging with latest main in case they have been fixed already.

File: src/mobius/components/_rms_norm.py [hunk 1]
```diff
@@ -65,16 +65,20 @@ def __init__(self, hidden_size: int, eps: float = 1e-6):
...
+        # which does hidden_states.to(float32) and gate.to(float32).
+        h_f32 = op.Cast(hidden_states, to=1)
+        g_f32 = op.Cast(gate, to=1)
+        gate_activated = op.Mul(g_f32, op.Sigmoid(g_f32))
+        gated = op.Mul(h_f32, gate_activated)
+        # RMSNorm in fp32, then cast back to input dtype
+        normed = op.RMSNormalization(
             gated,
-            self.weight,
+            op.Cast(self.weight, to=1),
```
justinchuby:
```suggestion
            op.Cast(self.weight, to=ir.DataType.FLOAT),
```

src/mobius/components/_rms_norm.py
```diff
@@ -98,16 +102,20 @@ def __init__(self, hidden_size: int, eps: float = 1e-6):
...
             axis=-1,
         )
-        # Then apply gate: normed * SiLU(gate)
-        gate_activated = op.Mul(gate, op.Sigmoid(gate))
-        return op.Mul(normed, gate_activated)
+        # Apply gate in fp32: normed * SiLU(gate), then cast back.
+        # Matches HF Qwen3_5RMSNormGated which does gate.to(float32).
+        g_f32 = op.Cast(gate, to=1)
+        gate_activated = op.Mul(g_f32, op.Sigmoid(g_f32))
+        result = op.Mul(op.Cast(normed, to=1), gate_activated)
```
justinchuby:
```suggestion
        result = op.Mul(op.Cast(normed, to=ir.DataType.FLOAT), gate_activated)
```

src/mobius/components/_rms_norm.py
```diff
@@ -98,16 +102,20 @@ def __init__(self, hidden_size: int, eps: float = 1e-6):
...
             epsilon=self.variance_epsilon,
+            stash_type=1,
             axis=-1,
         )
-        # Then apply gate: normed * SiLU(gate)
-        gate_activated = op.Mul(gate, op.Sigmoid(gate))
-        return op.Mul(normed, gate_activated)
+        # Apply gate in fp32: normed * SiLU(gate), then cast back.
+        # Matches HF Qwen3_5RMSNormGated which does gate.to(float32).
+        g_f32 = op.Cast(gate, to=1)
```
justinchuby: ditto

justinchuby (commented):
Should we generate L4-L5 reference test data for it too?

src/mobius/components/_rms_norm.py
```diff
@@ -65,16 +62,19 @@ def __init__(self, hidden_size: int, eps: float = 1e-6):
...
 
     def forward(self, op: builder.OpBuilder, hidden_states: ir.Value, gate: ir.Value):
-        # Gate first: hidden * SiLU(gate)
-        gate_activated = op.Mul(gate, op.Sigmoid(gate))
-        gated = op.Mul(hidden_states, gate_activated)
-        # Then normalize the gated result
-        return op.RMSNormalization(
+        # Upcast to fp32 for SiLU gating + RMSNorm variance, matching HF
+        # which does hidden_states.to(float32) and gate.to(float32).
+        h_f32 = op.Cast(hidden_states, to=ir.DataType.FLOAT)
```
xadupre, line 80: This should not be useful, RMSNormalization is supported to do the computation in float by default: stash_type=1

.github/skills/adding-a-new-model/SKILL.md
```diff
@@ -733,6 +733,58 @@ for i in range(num_groups - 1):
...
 ```
 
+### 16. Precision-sensitive ops need fp32 upcast
+
+**Symptom:** Type mismatch errors (`tensor(float) vs tensor(bfloat16)`)
+when loading a model built with `--dtype bf16`, or numerical drift compared
+to HuggingFace when running in fp16/bf16.
+
+**Root cause:** Operations like `exp`, `softplus`, `sigmoid` (in gated norms),
+and RMSNorm variance are numerically sensitive and must run in float32 to
```
xadupre, line 743: RMSNorm is supposed to run in float32 (stash_type=1 by default).

src/mobius/components/_rms_norm.py
gramalingam: What about the sigmoid 2 lines below? Are you recommending lower-precision for these ops? My recommendation to copilot was to mimic the precision used in HF transformer version of the model

src/mobius/components/_rms_norm.py
```diff
@@ -84,25 +84,46 @@ def forward(self, op: builder.OpBuilder, hidden_states: ir.Value, gate: ir.Value
...
                 op.Constant(value_ints=[0, n_groups, self.group_size]),
             )
-            weight_grouped = op.Reshape(
-                self.weight,
-                op.Constant(value_ints=[n_groups, self.group_size]),
+            # variance = mean(x^2, axis=-1, keepdims=True)
+            variance = op.ReduceMean(
+                op.Mul(grouped, grouped), axes=[-1], keepdims=True,
             )
-            normed = op.RMSNormalization(
```
justinchuby: I think we should retain rmsnorm and move the workaround to the ort rewrites.

src/mobius/components/_gated_deltanet.py
```diff
@@ -214,23 +214,24 @@ def forward(
...
-        #   g = -self.A_log.float().exp() * F.softplus(a.float() + self.dt_bias)
-        # Without .float(), exp() on fp16 A_log can produce -inf.
-        a_f32 = op.Cast(a, to=ir.DataType.FLOAT)
-        dt_bias_f32 = op.Cast(self.dt_bias, to=ir.DataType.FLOAT)
-        a_log_f32 = op.Cast(self.A_log, to=ir.DataType.FLOAT)
-        softplus_val = op.Softplus(op.Add(a_f32, dt_bias_f32))
-        neg_a = op.Neg(op.Exp(a_log_f32))
-        g = op.Mul(neg_a, softplus_val)  # (B, S, num_v_heads), float32
+        # decay: (B, S, num_v_heads)
+        # Upcast to fp32 for softplus/exp (numerically sensitive).
```
gramalingam: @justinchuby : is this GatedDeltaNet used in Qwen3.5 ? I want to avoid causing any regression in Qwen3.5, and not sure if we have adequate CI tests to ensure that. But, otherwise, would be good to merge the PR sometime soon to avoid divergence

src/mobius/components/_rms_norm.py
gramalingam: Ok, emitting RMSNormalization for now. Will figure out workaround strategy separately.

src/mobius/components/_gated_deltanet.py
justinchuby: Yes - it is used. I can coordinate this with the qwen 3.5 changes.

src/mobius/components/_rms_norm.py
gramalingam: I think if this is fused into an op, the kernel can do this the right way, without overheads. An unfused op may need this cast overhead for precision?

src/mobius/components/_rms_norm.py
```diff
@@ -132,7 +133,7 @@ def forward(self, op: builder.OpBuilder, hidden_states: ir.Value, gate: ir.Value
             # Standard RMSNorm over the full dimension
             normed = op.RMSNormalization(
                 gated,
-                self.weight,
+                op.Cast(self.weight, to=ir.DataType.FLOAT),
```
justinchuby: Is this right?

gramalingam: Ok, fixed it. Just noticed that there might be a typo in the ONNX spec https://onnx.ai/onnx/operators/onnx__RMSNormalization.html#l-onnx-doc-rmsnormalization ... the output type should be T, not V ?

justinchuby: Agreed. It's a typo

justinchuby (approved):

PR #25, opened 2026-03-24
Display commit SHAs in CI outputs and add Qwen3.5 benchmarks

justinchuby:
### Architecture Diff
- Resolve `--base-ref` and `--head-ref` to short commit SHAs via `git rev-parse --short`
- Display compared SHAs in the markdown header (e.g., "Comparing `abc1234` → `def5678`")
- Graceful fallback to raw ref string if SHA resolution fails (shallow clones, invalid refs)

### Performance Comparison
- Add Qwen3.5 models to benchmarks (text, moe, vl) with custom VL builder
- Display commit SHAs in benchmark comparison markdown header
- Update perf_baseline.json with Qwen3.5 baseline values

### Testing
- Added tests for SHA display (present and absent cases)
- All existing tests pass

### DIFF
--- a/tests/benchmark_compare.py
+++ b/tests/benchmark_compare.py
@@ -40,6 +40,9 @@ def compare(current_path: str, baseline_path: str) -> tuple[str, bool]:
     rows: list[tuple[str, str, str, str, str, str]] = []
     has_blocker = False
 
+    base_sha = baseline.get("_metadata", {}).get("commit", "unknown")
+    head_sha = current.get("_metadata", {}).get("commit", "unknown")
+
     for model, metrics in sorted(current["models"].items()):
         base = baseline.get("models", {}).get(model, {})
         for metric in THRESHOLDS:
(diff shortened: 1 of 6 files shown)


tests/benchmark_compare.py
```diff
@@ -40,6 +42,9 @@ def compare(current_path: str, baseline_path: str) -> tuple[str, bool]:
     rows: list[tuple[str, str, str, str, str, str]] = []
     has_blocker = False
 
+    base_sha = baseline.get("_metadata", {}).get("commit", "unknown")
+    head_sha = current.get("_metadata", {}).get("commit", "unknown")
```
titaiwangms, line 46: What is the input/output (prefill or decode) of benchmarking here?

justinchuby: This is currently just checking the model size. Will create actual perf benchmarks later.

PR #38, opened 2026-03-25
Add Linux CUDA CI workflow with 5-tier testing pipeline

tadani3:
Added capability to run L4 and L5 tests on the ONNXRuntime A10 GPU.


justinchuby:
Don't run all tiers in a single workflow. Instead, run them in parallel. We already have workflows for all tiers. I think we just need to correctly assign machine pools for them.

justinchuby (approved):

PR #54, opened 2026-03-28
Memory-mapped safetensors loading with lazy tensor optimization

justinchuby:
## Summary
Replace `safetensors.torch.load_file()` with a custom parser that memory-maps weights using `torch.UntypedStorage.from_file()`. The returned `torch.Tensor` objects are backed by mmap'd storage — the OS pages in data lazily on demand, no custom descriptor needed.

## How it works
`torch.UntypedStorage.from_file()` sets up a virtual memory mapping without reading any data (0 KB RSS). Sub-storage slicing (`storage[start:end]`) returns a view (not copy) into the same mapping. The resulting tensors are inherently lazy via the OS virtual memory system.

## Features
- **Custom safetensors parser**: Parses header only (8-byte prefix + JSON), creates mmap'd tensors
- **Drop-in replacement**: Returns `dict[str, torch.Tensor]` — same interface as `safetensors.torch.load_file()`
- **Security**: Header size cap (100MB), offset validation, byte-size validation, dtype validation
- **Multi-shard support**: Handles model.safetensors.index.json

## Memory improvements

| Model | Before (eager) | After (mmap) | Savings |
|-------|---------------|--------------|----------|
| Qwen2.5-0.5B (1.26 GB, 1 shard) | 8,071 MB VmPeak | 7,128 MB VmPeak | -943 MB (-12%) |
| Mistral-7B (14.5 GB, 2 shards) | 25,303 MB VmPeak | 20,151 MB VmPeak | -5,152 MB (-20%) |

mmap-backed pages are file-backed and can be evicted by the OS without causing OOM.

## Integration tests
24 integration tests pass (prefill + decode + generation). All 11 failures are pre-existing.

## New files
- `src/mobius/integrations/_safetensors.py` — parser + mmap loader (~170 lines)
- `src/mobius/integrations/_safetensors_test.py` — 34 tests


justinchuby:
Closing: safetensors.torch.load_file() already returns mmap-backed tensors internally, so a custom parser provides no memory benefit. The investigation was valuable — confirmed that safetensors uses mmap and that torch.UntypedStorage.from_file() is pure virtual memory mapping with zero RSS cost.

PR #57, opened 2026-03-28
Fix ORT CUDA EP Scan shape mismatch in LinearAttention

justinchuby:
## Problem

ORT CUDA EP fails to resolve symbolic `dim_params` (H, d_k, d_v) in Scan body graphs inside inlined functions. It falls back to `dim_value=0→1`, creating undersized carry buffers (e.g. `{1,16,1,1}` instead of `{1,16,128,128}`). CPU EP handles this correctly.

## Fix

Make H (`kv_num_heads`), d_k (`head_k_dim`), and d_v (`head_v_dim`) concrete integers in the Scan body shapes. These values are all known at graph build time from the model config. Only batch (B) remains symbolic.

## Changes

- `linear_attention()`: add required `head_k_dim` and `head_v_dim` params
- `_build_recurrence_body()`: accept concrete dims, use them in `ir.Shape` instead of symbolic dim_param strings
- Update task call site to pass dims from `LinearAttentionDims`
- Update all test calls with concrete head dimensions

## Testing

- All 17 LinearAttention unit tests pass
- All 14 qwen3.5/qwen3_next build graph tests pass
- Full test suite: 1694 passed, 1 skipped (2 pre-existing golden_test failures unrelated)

### DIFF
--- a/.github/skills/debugging-scan-shapes/SKILL.md
+++ b/.github/skills/debugging-scan-shapes/SKILL.md
@@ -1,0 +1,221 @@
+# Debugging ORT CUDA EP Scan Shape Mismatches
+
+## When to use this skill
+
+Use this when you see a **shape mismatch error in a Scan node** that
+only occurs on CUDA EP (CPU works fine). The typical error message:
+
+```
+Shape mismatch attempting to re-use buffer. {1,16,1,1} != {1,16,128,1}
+```
+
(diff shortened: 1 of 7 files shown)


justinchuby:
@gramalingam

justinchuby:
Closing because this is an ORT issue. Workaround at https://github.com/onnxruntime/mobius/pull/59

PR #69, opened 2026-03-31
Remove manual dtype/shape setting for graph outputs

justinchuby:
## Summary

The shape inference pass (`SymbolicShapeInferencePass`) already infers output dtype/shape automatically during model optimization. This removes redundant manual shape/type assignments from output registration functions in task files.

### Task files (10 files, -71 lines net)

- **`_register_kv_cache_outputs()`** (`_base.py`): Removed `past_key_values` parameter and shape derivation logic. Now only names and registers outputs.
- **`_register_hybrid_cache_outputs()`** (`_base.py`): Removed `past_key_values` parameter and all shape/type copying for full_attention, linear_attention, mamba, and lightning_attention layer types.
- **`_register_static_cache_outputs()`** (`_causal_lm.py`): Removed `dtype`, `batch`, `max_seq_len`, `kv_hidden` parameters and explicit shape/type setting.
- **`_seq2seq.py`**: Removed inline shape/type assignments for self-attention and cross-attention KV cache outputs.
- Updated all callers in `_multimodal.py`, `_phi4mm_multimodal.py`, `_speech_language.py`, `_speech_to_text.py`, `_tts.py`, `_vision_language.py`, `_vision_language_3model.py`.

### Test infrastructure

- `tests/build_graph_test.py`: Run `SymbolicShapeInferencePass` before the ONNX checker in `_run_onnx_checker()` to match the real pipeline (checker requires output shapes).

## Test Results

1860 passed, 9 skipped, 0 failed. Lint clean.

### DIFF
--- a/src/mobius/tasks/_base.py
+++ b/src/mobius/tasks/_base.py
@@ -138,33 +138,16 @@ def _register_kv_cache_outputs(
     graph: ir.Graph,
     present_key_values: list[tuple[ir.Value, ir.Value]],
     *,
-    past_key_values: list[tuple[ir.Value, ir.Value]] | None = None,
     prefix: str = "present",
 ) -> None:
     """Name and register KV cache outputs on the graph.
 
-    When ``past_key_values`` is provided, output shapes are derived from
-    the corresponding past inputs (replacing the sequence-length dimension
-    with a ``total_sequence_len`` symbolic dim).
+    Output shapes and dtypes are inferred by the shape inference pass
+    that runs during model optimization.
     """
-    total_seq_len = ir.SymbolicDim("total_sequence_len")
     for i, (present_key, present_value) in enumerate(present_key_values):
         present_key.name = f"{prefix}.{i}.key"
         present_value.name = f"{prefix}.{i}.value"
-        if past_key_values is not None:
-            past_k, past_v = past_key_values[i]
-            if past_k.shape is not None and len(past_k.shape) >= 3:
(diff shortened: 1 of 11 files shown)


justinchuby:
@copilot show me an example of the model outputs and what is changed before and after this PR

gramalingam:
Doesn't this interact with the other question: of which shape-inference method to use? Specifically, use of ONNX shape inference may or may not produce the same results. In particular, if we expect outputs with symbolic dims like "past_length + current_length", that may not happen with ONNX shape inference ... at least, not until ONNX is updated to do this. Of course, this is an advantage of the other symbolic-shape-inference: but it would be better to make a conscious decision of whether we are going to dependence on that alternative shape-inference before doing this instead of having that decision forced on us by this refactoring.

justinchuby:
> Doesn't this interact with the other question: of which shape-inference method to use? Specifically, use of ONNX shape inference may or may not produce the same results. In particular, if we expect outputs with symbolic dims like "past_length + current_length", that may not happen with ONNX shape inference ... at least, not until ONNX is updated to do this. Of course, this is an advantage of the other symbolic-shape-inference: but it would be better to make a conscious decision of whether we are going to dependence on that alternative shape-inference before doing this instead of having that decision forced on us by this refactoring.

I think that if we use any other methods this refactoring is still ok? As long as we have some names for the shapes. We don't need to provide a contract on _how_ the output shapes are named.

PR #70, opened 2026-03-31
Comparison: SymbolicShapeInferencePass vs onnx_ir.passes.common.ShapeInferencePass

justinchuby:
## Summary

This PR adds a structured comparison analysis between the two shape inference approaches available in mobius, implemented as a test module that both documents and verifies the findings.

**File added**: `tests/shape_inference_comparison_test.py` (508 lines, 11 tests)

---

### 1. `SymbolicShapeInferencePass` (mobius primary)

Wraps `onnx_shape_inference.infer_symbolic_shapes` from the `onnx-shape-inference` package. Works natively on `ir.Model`.

### 2. `onnx_ir.passes.common.ShapeInferencePass`

Wraps `onnx.shape_inference.infer_shapes` (ONNX C++ engine). Round-trips through protobuf serialization.

---

## Comparison Results

| Property | `SymbolicShapeInferencePass` | `onnx_ir ShapeInferencePass` |
|---|---|---|
| Starting state required | Any | Fully typed model |
| Shape coverage (fully stripped model) | **65/65 (100%)** | **0/65 (0% — fails)** |
| Shape coverage (raw model from onnxscript) | 65/65 | 61/65 |
| Dtype coverage (fully stripped model) | **65/65 (100%)** | **0/65 (0% — fails)** |
| Produces symbolic dimensions | **Yes** (`batch`, `sequence_len`, `past_seq_len + seq_len`) | No (concrete integers only) |
| `com.microsoft` contrib op support | **Yes (51 ops)** | No |
| Standard ONNX + opset-23 op support | Yes (199 ops) | Yes |
| Serialization round-trip required | No | Yes |
| Can run after optimization pipeline | Yes | Yes (adds nothing) |

---

### Finding 1: Bootstrap failure of `ShapeInferencePass`

`ShapeInferencePass` **cannot bootstrap** shape inference on fresh mobius models. When intermediate values have no dtype (as happens after stripping all type info), `onnx_ir.serde.serialize_value_into` raises `SerdeError` because protobuf `ValueInfoProto.type` requires a concrete `elem_type`. Similarly, when `initializer.const_value` is `None`, the serialization path asserts.

The pass catches all exceptions and returns `modified=False`, leaving the model unchanged. This means it is **not viable as the primary inference pass** — it can only run after another pass has already established full type coverage.

### Finding 2: Symbolic vs concrete dimensions

Our `SymbolicShapeInferencePass` propagates **named symbolic dimensions**:
```
input_ids: [batch, sequence_len]
attention_mask: [batch, past_seq_len + seq_len]
Gather output: [batch, sequence_len, 64]
Attention output: [batch, sequence_len, 64]
```

`onnx.shape_inference.infer_shapes` can only substitute concrete integers when data values are statically known. It has no concept of named symbolic dimensions, so dynamic axes appear as unnamed `?`. This is critical for:
- ORT GenAI (reads dimension names to set up dynamic cache)
- Model analysis tools that need to distinguish batch vs. sequence axes
- Downstream graph transformations that pattern-match on dimension names

### Finding 3: Custom op coverage

`SymbolicShapeInferencePass` registers handlers for 51 `com.microsoft.*` contrib ops (RotaryEmbedding, GroupQueryAttention, Attention, MoE, etc.). `ShapeInferencePass` has no knowledge of these ops.

**Current state**: Mobius models use ONNX opset-23 standard ops (`Attention`, `RMSNormalization`, `RotaryEmbedding` — all in the `ai.onnx` domain), not `com.microsoft.*`. Both passes cover our current models. The gap matters for any future path targeting ORT contrib ops directly.

### Finding 4: No complementary benefit

After `SymbolicShapeInferencePass` achieves 100% coverage, running `ShapeInferencePass` reports `modified=False` — it adds nothing. The passes are compatible as a pipeline but there is no use case where combining them improves coverage.

---

## Op Coverage

```
onnx-shape-inference (our symbolic engine):
  Standard ONNX (ai.onnx /  domain): 199 ops
  com.microsoft contrib ops:              51 ops
  ai.onnx.ml:                             4 ops
  
ONNX C++ (onnx 1.20.1, opset 25):
  Standard ONNX + opset 23:              all standard ops
  com.microsoft contrib ops:             0
```

Opset-23 ops used by all mobius models (`Attention`, `RMSNormalization`, `RotaryEmbedding`) are registered in both engines.

---

## Recommendation

Keep `SymbolicShapeInferencePass` as the primary (and only) shape inference pass in the build pipeline. `onnx_ir.passes.common.ShapeInferencePass` is not a drop-in replacement and provides no additional coverage after symbolic inference runs.

If ONNX C++ shape validation is ever desired (separate from the existing `CheckerPass`), it can be added as a final post-optimization step after all dtypes are filled in — but the current `CheckerPass` already covers this use case.

---

## Test Summary

11 tests covering:
1. `SymbolicShapeInferencePass` achieves full coverage from scratch
2. `ShapeInferencePass` fails when dtypes are unknown
3. `ShapeInferencePass` succeeds after symbolic pass (adds nothing)
4. Symbolic pass produces named symbolic dimensions
5. Graph inputs carry semantic dimension names (`batch`, `sequence_len`)
6. Op coverage comparison (counts and assertions)
7. Full coverage on 4 representative model architectures (llama, qwen2, gemma, mistral)
8. Structured comparison summary printed to test output

### DIFF
--- a/tests/shape_inference_comparison_test.py
+++ b/tests/shape_inference_comparison_test.py
@@ -40,7 +40,6 @@ from mobius._builder import SymbolicShapeInferencePass, build_from_module
 from mobius.models import CausalLMModel
 from mobius.tasks import CausalLMTask
 
-
 # ---------------------------------------------------------------------------
 # Helpers
 # ---------------------------------------------------------------------------
@@ -325,8 +324,8 @@ def test_op_coverage_comparison():
     graph variant targeting ORT contrib ops directly.
     """
     import onnx
-    import onnx_shape_inference._ops  # trigger registration
     import onnx_shape_inference as osi
+    import onnx_shape_inference._ops  # trigger registration
 
     reg = osi.registry
 
@@ -358,17 +357,17 @@ def test_op_coverage_comparison():
             )
 
     # Summary for human readers (always printed)
-    print(f"\n=== Op Coverage Summary ===")
+    print("\n=== Op Coverage Summary ===")
     print(f"onnx version: {onnx.__version__} (opset {onnx_opset})")
     print(
         f"SymbolicShapeInferencePass: {standard_count} standard ops, "
         f"{ms_contrib_count} com.microsoft contrib ops"
     )
     print(
-        f"ShapeInferencePass (ONNX C++): covers standard ONNX + opset 23, "
-        f"NO com.microsoft contrib ops"
+        "ShapeInferencePass (ONNX C++): covers standard ONNX + opset 23, "
+        "NO com.microsoft contrib ops"
     )
-    print(f"Mobius model ops: all standard ONNX + opset 23 (no contrib ops currently)")
+    print("Mobius model ops: all standard ONNX + opset 23 (no contrib ops currently)")
 
     assert standard_count >= 190, f"Expected 190+ standard ops, got {standard_count}"
     assert ms_contrib_count >= 40, f"Expected 40+ ms contrib ops, got {ms_contrib_count}"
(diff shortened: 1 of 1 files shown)


gramalingam:
I don't understand Finding 2: surely ONNX shape inference is capable of propagating symbolic dims. (But it won't create new meaningful names, but if we start with symbolic dims in shapes of inputs, they are propagated.)

As for Finding 1: that feels like a bug in ir serialization?

justinchuby:
Right. I am having it fix findings 1

justinchuby:
Clarification on the review feedback about task file reverts: PR #70 contains **only** `tests/shape_inference_comparison_test.py` (+508 -0, 1 commit). There are no task file changes in this branch — `git diff main justinchu/shape-inference-comparison --name-only` shows only the new test file.

PR #69's simplifications (removing manual shape/type assignments from tasks) are fully intact. This PR is already option (a): comparison-test-only documentation of the two inference approaches, with no contradictory code changes.

justinchuby:
Good analysis. Closed

PR #72, opened 2026-03-31
Fix weight alignment roundtrip for ctrl, xlm, and deepseek

justinchuby:
Fixes 3 weight alignment CI failures:

## ctrl and xlm
`preprocess_weights()` called a rename function that returned `None` for unrecognized names, silently dropping them. Added a passthrough fallback so already-ONNX-aligned parameter names are preserved in the output dict.

## deepseek v2/v3
Removed an incorrect rename: `.mlp.shared_experts.` → `.mlp.moe.shared_experts.`

The ONNX module structure is `_DeepSeekMoEFFN.shared_experts` (alongside `_DeepSeekMoEFFN.moe`), so the ONNX parameter path is `mlp.shared_experts.*` — which matches HF directly. The rename was incorrectly transforming already-correct ONNX names to a non-existent path, causing 3 shared expert parameters to be dropped in layer 1 (MoE layer).


justinchuby:
@copilot check if there is an segfault happening before this PR on the main CI tests. Show me your findings

justinchuby:
@copilot if that's the case (getting SIGTERM), should we set fail fast to false?

pyproject.toml
```diff
@@ -10,7 +10,8 @@ authors = [
     { name = "ONNX Contributors", email = "onnx-technical-discuss@lists.lfaidata.foundation" },
 ]
 readme = "README.md"
-requires-python = ">=3.10"
+# Python 3.10 dropped: onnxruntime no longer supports it.
+requires-python = ">=3.11"
```
justinchuby:
```suggestion
requires-python = ">=3.10"
```

PR #85, opened 2026-04-02
Add Microsoft VL models: phi3_v (Phi-3-Vision) + phi4-siglip (Phi-4-Reasoning-Vision)

justinchuby:
## Summary
Adds two Microsoft vision-language models.

### phi3_v (Phi-3-Vision)
- CLIP ViT-L/14 vision encoder (336px) with HD crop support
- CumSum+Gather+Where image token fusion in embedding model
- Standard VisionLanguageTask 3-model split (decoder, vision, embedding)
- HF stores vision config in img_processor dict (non-standard)

### phi4_siglip (Phi-4-Reasoning-Vision)
- SigLIP-2 SO400M vision encoder (384px)
- Same CumSum+Gather+Where fusion pattern as phi3_v
- HF vision_config omits image_size and patch_size (extracted from sub-config)

### Key fixes
- IMAGE_TOKEN_ID clamping: op.Where(image_mask, 0, input_ids) before embed_tokens to prevent negative token IDs reaching ONNX Gather

## Testing
4 new tests (2 per model), all pass. Full suite passes.


justinchuby:
Addressed all six Copilot findings: Phi-4 vision now preserves batch dimension; Phi-3 decoder drops vision-only weights; invalid chat templates fall back; headers match repository convention; and Phi-3 vision extraction always supplies CLIP defaults. Added coverage for decoder filtering.

PR #97, opened 2026-04-02
[DRAFT] Nemotron_h (Mamba) changes

gramalingam:
* Add scan-based support to Mamba block to handle sequence of tokens (of length > 1)
* Add a patch to HF code for a Nemotron bug that was causing the HF version to use a random bias value. This helps reduce some of the numeric divergence observed between ONNX and HF versions.

(But there is still some numeric divergence that builds up over the sequence: TBD)

### DIFF
--- a/src/mobius/components/_mamba_block.py
+++ b/src/mobius/components/_mamba_block.py
@@ -29,6 +29,7 @@ from onnxscript import nn
 from onnxscript._internal import builder
 from onnxscript.onnx_types import FLOAT
 
+from mobius._flags import flags
 from mobius.components._common import INT64_MAX, Linear
 from mobius.components._rms_norm import GatedRMSNorm
 from mobius.components._ssm import Mamba2Scan, SelectiveScan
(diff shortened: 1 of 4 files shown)


src/mobius/components/_mamba_block.py
```diff
@@ -226,60 +226,133 @@ def __init__(
         self.norm = GatedRMSNorm(d_inner, eps=eps, group_size=norm_group_size)
         self.out_proj = Linear(d_inner, d_model, bias=proj_bias)
 
-    def forward(
+    @staticmethod
+    def _realize_submodule(
```
justinchuby, line 230: Is this needed? Ideally we don't need to touch onnxscript internal apis

gramalingam: This is not needed if and when we use onnxscript with this PR: https://github.com/microsoft/onnxscript/pull/2877 ... the problem is that the older mechanism for parameter realization will end up adding initializers to attribute subgraphs if the component's forward method is called from within a Scan/Loop's body subgraph, while we want it to be in main graph. (The onnxscript PR fixes it. But I left this in to avoid breaking things for devs unnecessarily.)

gramalingam:
Closing this in favor of a newer PR with more changes.

src/mobius/components/_mamba_block.py
justinchuby: We already target the latest dev of onnxscript so we should be ok. Just need all devs to run pip update

PR #109, opened 2026-04-06
[DRAFT] Add native GQA attention support for ORT GenAI compatibility

gramalingam:
Note: This is a copilot-authored PR. I haven't yet checked it. Creating the PR for easy review and collaboration.

Add a new GQAAttention component that emits com.microsoft::GroupQueryAttention directly, and a gqa=True mode on CausalLMTask that produces models compatible with the onnxruntime-genai runtime.

Key changes:
- New GQAAttention component with GQAContext NamedTuple
- DecoderLayer accepts pluggable attention_class parameter
- CausalLMTask(gqa=True) builds graphs with:
  - No position_ids input (RoPE fused inside GQA via do_rotary=1)
  - seqlens_k/total_seq_len computed from attention_mask
  - cos_cache/sin_cache as graph initializers
  - com.microsoft opset import
- CLI --gqa flag and build() module_kwargs parameter
- 10 new GQA-specific tests in build_graph_test.py

### DIFF
--- a/src/mobius/tasks/_causal_lm.py
+++ b/src/mobius/tasks/_causal_lm.py
@@ -1,7 +1,7 @@
 # Copyright (c) ONNX Project Contributors
 # SPDX-License-Identifier: Apache-2.0
 
-"""Causal language model tasks with internal and static KV cache."""
+"""Causal language model tasks with internal, static, and GQA KV cache."""
 
 from __future__ import annotations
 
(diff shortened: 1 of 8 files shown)


src/mobius/tasks/_causal_lm.py
```diff
@@ -23,10 +24,18 @@
 )
 
 
+def _find_rotary_emb(module: nn.Module) -> nn.Module | None:
```
gramalingam, line 27: This doesn't look good (searching for a module by name and attribute). I wonder why this is necessary.

justinchuby:
Superseded by https://github.com/onnxruntime/mobius/pull/134

PR #110, opened 2026-04-07
Some fixes for NemotronH

gramalingam:
* Fix some issues in HF transformer side that causes some numeric divergence with the ONNX implementation
* Add a couple of versions (using scan and chunked-SSD) of Mamba2Block (for dealing with sequence of length > 1).
* A few refinements to example script to run model


justinchuby:
Tests are failing

justinchuby:
## Can `com.microsoft.LinearAttention` Express Mamba2 SSD?

**TL;DR: Yes — with `update_rule="gated"` — and it would collapse all three forward modes in this PR into a single op call.**

---

### The LinearAttention Op

`com.microsoft.LinearAttention` (ORT ContribOps, opset 1) is a unified recurrent-attention op that supports both decode (T=1) and prefill (T>1) in one call, with a chunk-parallel implementation for GPU efficiency. Its `update_rule="gated"` recurrence is:

```
S_t = exp(g_t) * S_{t-1} + k_t ⊗ v_t
o_t = scale * q_t^T S_t
```

Inputs: `query, key, value, [past_state], [decay], [beta]`
Outputs: `output, present_state`

The `decay` input supports per-head scalar shape `(B, T, H_kv)` — exactly what Mamba2 needs.

---

### Mamba2 SSD Recurrence (from `Mamba2Scan.forward`)

```
dA_t = exp(A * dt_t)                     # scalar per head
h_t  = dA_t * h_{t-1} + dt_t * B_t ⊗ x_t   # outer product, B_t: d_state
y_t  = einsum(h_t, C_t) + D * x_t       # readout + skip
```

where `A_log` is `(num_heads,)`, `dt_t` is per-head, `B_t` / `C_t` are `(n_groups, d_state)` expanded to all heads.

---

### Exact Tensor Mapping

| Mamba2 tensor | LinearAttention arg | Packed shape | Notes |
|---|---|---|---|
| `C_mat` (readout) | `query` | `(B, T, H_q × d_state)` | `q_num_heads = num_heads` |
| `B_mat` (input matrix) | `key` | `(B, T, H_kv × d_state)` | `kv_num_heads = n_groups` |
| `dt * x` (discretized) | `value` | `(B, T, H_kv × d_head)` | absorb `dt` into value |
| `A * dt` (log-decay) | `decay` | `(B, T, H_kv)` | per-head scalar |
| `ssm_state` | `past_state` | `(B, H, d_state, d_head)` | requires transpose (see below) |

Attributes: `update_rule="gated"`, `q_num_heads=num_heads`, `kv_num_heads=n_groups`, `scale=1.0`

The `n_groups > 1` case maps naturally onto LinearAttention's GQA grouping (`q_num_heads / kv_num_heads = heads_per_group`).

---

### Three Items Outside the Op

1. **`dt` discretization** — `dt = softplus(dt_proj(x) + dt_bias)` is computed before calling LA; `dt * x` is passed as `value` and `A * dt` as `decay`. No structural change needed.

2. **D skip connection** — `y += D * x` is a plain `Add` after the LA output. One line.

3. **State shape transpose** — Current Mamba2 state is `(B, H, d_head, d_state)` but LA expects `(B, H, d_k, d_v) = (B, H, d_state, d_head)`. Need `Transpose([0,1,3,2])` on `past_state` input and `present_state` output. Minor.

---

### What This Means for This PR

This PR implements three modes — `Mamba2BlockSingle` (T=1), `Mamba2BlockScan` (ONNX Scan, sequential), `Mamba2BlockChunkedSSD` (parallel within chunks) — because ORT doesn't have a native op that handles both decode and prefill. **LinearAttention removes that need entirely.** A single LA call handles both paths, and ORT's implementation is chunk-parallel for prefill internally.

The entire `Mamba2BlockChunkedSSD._chunked_ssd()` method (~180 ONNX ops: segment_sum, CumSum, Trilu, cross-chunk propagation, 5-stage SSD) would collapse to:

```python
# Outside: compute dt, pack tensors
decay = op.Mul(a_2d, dt)                      # A * dt, shape (B,T,H_kv)
value = op.Mul(x_disc_packed, dt_packed)      # dt * x, shape (B,T,H_kv*d_head)

# The op
output, new_state = op.LinearAttention(
    C_packed,                     # query  (B,T,H_q*d_state)
    B_packed,                     # key    (B,T,H_kv*d_state)
    value,                        # value  (B,T,H_kv*d_head)
    past_state_transposed,        # (B,H_kv,d_state,d_head)
    decay,                        # (B,T,H_kv) — per-head scalar decay
    domain="com.microsoft",
    update_rule="gated",
    q_num_heads=self.num_heads,
    kv_num_heads=self.n_groups,
    scale=1.0,
)

# Outside: D skip
y = op.Add(output, op.Mul(D_skip, x))
```

---

### Recommendation

**Consider gating this PR on LinearAttention availability.** Since LinearAttention is already in `com.microsoft` ContribOps (confirmed in [ContribOperators.md](https://github.com/microsoft/onnxruntime/blob/main/docs/ContribOperators.md)), adding a rewrite rule that replaces the `Mamba2BlockChunkedSSD` ONNX subgraph with a single `LinearAttention` node (similar to how GQA rewrites work for attention) would:
1. Remove ~300 lines of chunked SSD ops from the ONNX graph
2. Let ORT choose the optimal chunk size at runtime
3. Unify decode and prefill into one code path

If LinearAttention isn't available in the minimum supported ORT version, the current ONNX-native fallback in this PR is still valuable — they can coexist as primary and fallback paths.

---

*Investigated by reading the PR diff, `Mamba2Scan.forward()` in `_ssm.py`, `Mamba2BlockChunkedSSD._chunked_ssd()`, and the LinearAttention spec in [ContribOperators.md](https://github.com/microsoft/onnxruntime/blob/main/docs/ContribOperators.md).*

gramalingam:
> Can `com.microsoft.LinearAttention` Express Mamba2 SSD?

Thanks for the suggestion. This version is working correctly at last. I suggest merging it. I will look at using LinearAttention as a follow-up PR. (Sometime copilots do a few dumb things that make it hard to get back to an earlier state. Better to get this version in, so that I can fallback to this easily.)

A second follow-on relates to the issue that was causing the numeric-precision: this model doesn't use RoPE (in HF), its config doesn't have rope_type, but we end up using a default RoPE instead of no RoPE. Not sure if we need to change anything about the way the default works in this setting. The current approach (the model class explicitly makes that decision) might be fine, but worth thinking about.

PR #114, opened 2026-04-07
Added GPU workflow for running existing validation scripts

tadani3:
This pull request introduces a new GitHub Actions workflow to validate GPU-based example scripts. The workflow is designed to run various example scripts covering multiple models and modalities (text, vision, audio) on a self-hosted GPU runner, ensuring that these examples execute successfully in a CI environment.

Key additions in the new workflow:

**CI/CD Automation:**

* Added a new workflow file `.github/workflows/validation_examples_gpu.yml` that runs a suite of example scripts on a self-hosted GPU runner using CUDA-enabled PyTorch.

**Environment Setup:**

* Installs Python 3.12, caches pip packages and HuggingFace models, and installs necessary dependencies including PyTorch with CUDA support.

**Model Example Coverage:**

* Runs a comprehensive set of example scripts for Phi, Qwen (text, vision, audio), Nemotron, and static cache generation, each with appropriate test data and timeouts.


.github/workflows/validation_examples_gpu.yml
```diff
@@ -0,0 +1,131 @@
...
+    name: Run Validation Example Scripts (GPU)
+    runs-on:
+      - self-hosted
+      - "1ES.Pool=onnxruntime-ep-mobius-github-linux-a10"
+      - "JobId=gpu-examples-${{ github.run_id }}-${{ github.run_number }}-${{ github.run_attempt }}"
+    timeout-minutes: 120
+    env:
+      HF_TOKEN: ${{ secrets.HF_TOKEN }}
+    steps:
+      - uses: actions/checkout@v6
```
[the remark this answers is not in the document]
tadani3: @copilot apply changes based on [this feedback](https://github.com/onnxruntime/mobius/pull/114#discussion_r3048166712)

.github/workflows/validation_examples_gpu.yml
```diff
@@ -0,0 +1,131 @@
...
+        run: pip install torch torchaudio --index-url https://download.pytorch.org/whl/cu124
+
+      - name: Install dependencies
+        run: |
+          pip install -r requirements/ci/requirements.txt
+          pip install -e '.[testing,transformers,ort-genai]'
+          pip install soundfile librosa ml_dtypes
+
+      # ── Phi ──────────────────────────────────────────────────────
+
```
justinchuby: Would be good to make sure if one script fails, the others will still run

.github/workflows/validation_examples_gpu.yml
```diff
@@ -0,0 +1,131 @@
...
+        run: >
+          python examples/phi4mm_ort_genai.py
+          --image testdata/pipeline-cat-chonk.jpeg
+        timeout-minutes: 20
+
+      # ── Qwen text ───────────────────────────────────────────────
+
+      - name: "qwen35_text_generation"
+        run: >
+          python examples/qwen35_text_generation.py
```
justinchuby: We can update the script to have a `--ci` option so that it exits with a non-zero code when failing

.github/workflows/validation_examples_gpu.yml
```diff
@@ -0,0 +1,131 @@
...
+          --device cuda --prompt "What is 2+2?"
```
justinchuby: Need to compare with HF

justinchuby (approved):

.github/workflows/validation_examples_gpu.yml
tadani3: Resolved. The workflow uses a run_example shell wrapper that catches each script's exit code individually, increments a failures counter, and continues to the next script. All examples always run regardless of prior failures. The job only exits non-zero at the end if failures > 0, so the CI summary shows exactly which scripts failed.

.github/workflows/validation_examples_gpu.yml
tadani3: Done, added --compare_hf to all models that have the flag as an option and removed models that do not from the script.

.github/workflows/validation_examples_gpu.yml
tadani3: All 7 scripts now accept a --ci flag. When we invoke it, any mismatch in output between ONNX/HuggingFace will trigger a sys.exit(1) which gets handled by the script.

.github/workflows/validation_examples_gpu.yml
tadani3: @justinchuby the qwen25_vl_ort_genai.py, qwen3_vl_ort_genai.py, and phi4mm_ort_genai.py script had a bug where their generate_with_image() and generate_multimodal methods() returned None instead of being stored for comparison with HuggingFace. I fixed in [[dd97c31]](https://github.com/onnxruntime/mobius/pull/114/commits/dd97c31026fca36b93ffe8a8100d7c2ac4d5d93a) but might be worth doing a quick review again.

justinchuby: Looks good to me, thanks!

.github/workflows/validation_examples_gpu.yml
```diff
@@ -0,0 +1,111 @@
+name: GPU Examples
+
+on:
+  workflow_dispatch:
```
justinchuby, line 6: should this be triggered on pr label or when there is a push to main, too?

tadani3: For now I would only run on push to main. Not every PR will affect the models in these validation scripts so it seems unnecessary to have it run that often. We can add specific labels to trigger this workflow more regularly in the future.

PR #121, opened 2026-04-08
fix: propagate compute dtype in components to prevent SIGABRT with --dtype f16

justinchuby:
## Problem

Running `python examples/qwen35_text_generation.py --model Qwen/Qwen3.5-0.8B --dtype f16 --device cuda` (or building with `execution_provider='cuda'`) crashes with:

```
stl_vector.h:1130 assertion '__n < this->size()' failed
SIGABRT
```

ORT accesses a node output vector by out-of-bounds index because the ONNX graph contains a type inconsistency that CUDA EP cannot recover from gracefully (CPU ORT raises a clean error; CUDA crashes).

## Root Cause

Two related bugs, both caused by `nn.Parameter()` defaulting to `ir.DataType.FLOAT` type annotation regardless of the build dtype:

### Bug 1: Type annotation mismatch at component boundaries
Components like `Embedding`, `OffsetRMSNorm`, `RotaryEmbedding`, `_DepthwiseConv1d` used float32-annotated parameters with float16 hidden states, causing type inference failures.

### Bug 2: Constant folding eliminates intentional upcasts
`GatedDeltaNet.dt_bias` and `A_log` are intentionally Cast to float32 for numerical stability (`Cast(param, to=FLOAT)`). But when `param.type=FLOAT` (annotation) and the target is also FLOAT, the constant folding pass sees `input_type == output_type` and eliminates the Cast as an identity operation. This leaves float16 `const_value` data with a float32 type annotation in the proto — ORT's CUDA EP crashes when it encounters this inconsistency.

The same pattern affects `A_log`, `D`, and `dt_bias` in the SSM family (`SelectiveScan`, `Mamba2Scan`) and `GatedRMSNorm.weight` in the CUDA workaround path.

## Fix

**Commit 1** — Component boundary fixes (CastLike pattern):
- `Embedding.forward()`: set Gather output type to build dtype
- `OffsetRMSNorm.forward()`: `CastLike(weight, hidden_states)` and `CastLike(1.0, hidden_states)`
- `apply_rotary_pos_emb()`: `CastLike(cos, x)` and `CastLike(sin, x)`
- `_DepthwiseConv1d.forward()`: `CastLike(weight, input_val)` and `CastLike(bias, input_val)`

**Commit 2** — Parameter dtype declaration (preserves intentional float32 upcasts):
- Declare `dt_bias`, `A_log`, `D` with `dtype=config.dtype` / `dtype` parameter so that `Cast(param_f16, to=FLOAT)` is seen as a type change (FLOAT16→FLOAT) and preserved — not eliminated as identity.
- Affected: `GatedDeltaNet`, `SelectiveScan`, `JambaSelectiveScan`, `Mamba2Scan`, `GatedRMSNorm`, and all model files that create these components (`mamba.py`, `bamba.py`, `nemotron_h.py`, `jamba.py`, `granitemoehybrid.py`).

## Verification

- ORT CPU loads and runs Qwen3.5 float16 graph correctly after both fixes
- All 2286 tests pass
- After fix: `Cast_101` (dt_bias upcast result) has `dtype=FLOAT` in proto — no more float16 data labeled as float32

### DIFF
--- a/src/mobius/components/_common.py
+++ b/src/mobius/components/_common.py
@@ -8,8 +8,6 @@ import onnx_ir as ir
 from onnxscript import nn
 from onnxscript._internal import builder
 
-from mobius._build_context import get_build_dtype
-
 # Used as Slice "end" to mean "all remaining elements along this axis".
 INT64_MAX = 9223372036854775807
 
--- a/src/mobius/components/_common.py
+++ b/src/mobius/components/_common.py
@@ -55,17 +55,7 @@ class Embedding(nn.Module):
         self.padding_idx = padding_idx
 
     def forward(self, op: builder.OpBuilder, input_ids: ir.Value):
-        result = op.Gather(self.weight, input_ids)
-        # Propagate the compute dtype so downstream ops see the correct type.
-        # nn.Parameter creates float32 initializers by default; if the model is
-        # built with a non-default dtype (e.g. float16) the Gather output must be
-        # annotated with that dtype so ORT's type checker and graph transformers
-        # see a consistent type throughout the graph.  This mirrors how Linear
-        # propagates its input dtype via `result.type = x.type`.
-        dtype = get_build_dtype()
-        if dtype != ir.DataType.FLOAT:
-            result.type = ir.TensorType(dtype)
-        return result
+        return op.Gather(self.weight, input_ids)
 
 
 class LayerNorm(nn.Module):
--- a/src/mobius/components/_common.py
+++ b/src/mobius/components/_common.py
@@ -8,6 +8,8 @@ import onnx_ir as ir
 from onnxscript import nn
 from onnxscript._internal import builder
 
+from mobius._build_context import get_build_dtype
+
 # Used as Slice "end" to mean "all remaining elements along this axis".
 INT64_MAX = 9223372036854775807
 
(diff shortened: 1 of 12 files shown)


src/mobius/components/_ssm.py
```diff
@@ -24,6 +24,38 @@
...
+
+    ``get_build_dtype()`` returns the correct dtype because :func:`build` wraps
+    module construction in :func:`build_context` before calling
+    ``module_class(config)``. For direct ``build_from_module`` callers who
+    construct modules outside a build context, ``_cast_module_dtype`` provides
+    the same guarantee by updating FLOAT annotations to the target dtype before
+    graph building.
+
+    Used by: A_log, D (always upcast to fp32 for numerical stability), dt_bias,
+    _RMSNorm.weight (Jamba), GatedRMSNorm.weight.
```
justinchuby:
```suggestion
HuggingFace reference: ``MambaMixer`` (SSM portion).
```

src/mobius/components/_common.py
```diff
@@ -53,7 +55,17 @@ def __init__(
...
+        # Propagate the compute dtype so downstream ops see the correct type.
+        # nn.Parameter creates float32 initializers by default; if the model is
+        # built with a non-default dtype (e.g. float16) the Gather output must be
+        # annotated with that dtype so ORT's type checker and graph transformers
+        # see a consistent type throughout the graph.  This mirrors how Linear
+        # propagates its input dtype via `result.type = x.type`.
+        dtype = get_build_dtype()
+        if dtype != ir.DataType.FLOAT:
+            result.type = ir.TensorType(dtype)
+        return result
```
justinchuby:
```suggestion
        return op.Gather(self.weight, input_ids)
```

justinchuby:
Does not fix the issue

PR #129, opened 2026-04-08
refactor: replace FusedMatMul with Transpose+MatMul, add initializer-folding passes

justinchuby:
## Summary

Four graph optimization changes for better downstream tooling compatibility.

### Change 1: Linear emits Transpose+MatMul instead of FusedMatMul
- `Linear.forward()` now emits `Transpose(weight, perm=[1,0])` + `MatMul(x, w_t)`
- `LoRALinear.forward()` updated consistently  
- For EPs that support FusedMatMul (default), `fused_matmul_rules()` runs as a pre-fusion step **before** `register_function_bodies()` to avoid a cyclic dependency where the FusedMatMul function body itself contains Transpose+MatMul

### Change 2: FoldTransposedInitializerPass (`src/mobius/passes/`)
- Post-weights pass: detects `Transpose(initializer, perm=[1,0])` patterns
- Pre-transposes weight data using `ir.LazyTensor` (deferred evaluation, avoids peak memory)
- Stores as `{name}_t` initializer; removes the runtime Transpose node
- 6 unit tests

### Change 3: FoldConcatInitializersPass (`src/mobius/passes/`)
- Post-weights pass: detects `Concat` nodes where ALL inputs are initializers
- Pre-concatenates using `ir.LazyTensor`; handles QKV packing at build time
- 6 unit tests

### Change 4: Reduce constant-fold size limit to default
- `_FOLD_OUTPUT_SIZE_LIMIT` reduced from `256 * 1024 * 1024` (256M elements) to `DEFAULT_CONSTANT_FOLD_OUTPUT_SIZE_LIMIT` (262144)
- Large weights are now handled explicitly by the two new passes

## Testing
- 2315 passed, 29 skipped
- Lintrunner: clean


titaiwangms (approved):

src/mobius/_passes/_fold_concat.py
```diff
@@ -0,0 +1,172 @@
...
+                continue
+
+            # Build a name for the packed initializer from the input names and axis.
+            # The concatenated value depends on both, so include axis to avoid
+            # collisions when the same inputs are concatenated along different axes.
+            input_names = [v.name for v in inputs]  # type: ignore[union-attr]
+            packed_name = "__".join(input_names) + f"__axis_{axis}__concat"
+
+            # If an equivalent packed initializer already exists, reuse it and still
+            # remove this Concat node so all eligible Concat nodes are folded.
```
justinchuby: How do we know it is the same?

src/mobius/_passes/_fold_concat.py
```diff
@@ -0,0 +1,172 @@
...
+                        np.concatenate([v.const_value.numpy() for v in ins], axis=ax)  # type: ignore[union-attr]
+                    ),
+                    dtype=dtypes[0] or ir.DataType.FLOAT,
+                    shape=out_shape,
+                    name=new_val.name,
+                )
+            else:
+                # No data yet — leave const_value=None and record the source names
+                # so fold_initializers_after_weights() can fill it in later.
+                new_val.metadata_props["_fold_sources"] = ",".join(input_names)
```
justinchuby: Use a namespaced key

src/mobius/_passes/_fold_transpose.py
```diff
@@ -0,0 +1,143 @@
...
+
+            out_val = node.outputs[0]
+
+            if inp.name not in folded:
+                new_name = f"{inp.name}_t"
+
+                if new_name in model.graph.initializers:
+                    # An equivalent transposed initializer already exists (e.g. from a
+                    # previous pass invocation).  Reuse it rather than silently overwriting.
+                    folded[inp.name] = model.graph.initializers[new_name]
```
justinchuby: Why will we see that initializer again if folded?

src/mobius/_optimizations.py
```diff
@@ -13,18 +13,39 @@
...
+   not yet loaded the folded initializers record their source in
+   ``metadata_props`` so that :func:`fold_initializers_after_weights` can
+   materialise the values later.
+
+   .. important::
+       Stage 5 does **not** run :class:`~onnx_ir.passes.common.RemoveUnusedNodesPass`
+       after folding.  The original source initializers must remain registered
+       so that :func:`apply_weights` can later populate their ``const_value``.
+       :func:`fold_initializers_after_weights` materialises deferred values and
+       then removes the (now-redundant) source initializers.
```
justinchuby: We should actually remove this. Weightless model should match model with weights.

src/mobius/_optimizations.py
```diff
@@ -83,9 +105,13 @@ def call(self, model: ir.Model) -> ir.passes.PassResult:
...
-# Set high enough to fold packed-QKV weights (Llama-3.2-1B packed QKV is
-# ~6.3M elements per layer; 256M gives ample headroom for large models).
-_FOLD_OUTPUT_SIZE_LIMIT = 256 * 1024 * 1024
+# Set to onnxscript's standard default.  Large weight tensors (Transpose,
+# Concat/QKV packing) are handled by FoldTransposedInitializerPass and
+# FoldConcatInitializersPass after weight loading, so the general constant-fold
+# pass no longer needs a high limit.
+_FOLD_OUTPUT_SIZE_LIMIT = (
+    onnxscript.optimizer._constant_folding.DEFAULT_CONSTANT_FOLD_OUTPUT_SIZE_LIMIT
+)
```
justinchuby: Just set a concrete number

PR #134, opened 2026-04-09
feat: emit GroupQueryAttention directly in Attention when EP supports it

justinchuby:
## Summary

Eliminates the post-hoc `RotaryAttentionToGQA` rewrite rule as the primary path for GQA-capable EPs, replacing it with direct `com.microsoft::GroupQueryAttention` emission at graph construction time.

### New: `GQAContext` NamedTuple (`components/_attention.py`)
A typed bundle of per-graph scalars passed through the decoder stack as `attention_bias`:
- `seqlens_k`: per-batch KV sequence length `[batch]` INT32
- `total_seq_len`: scalar INT32
- `cos_cache` / `sin_cache`: full RoPE tables (not gathered slices)

### Modified: `Attention.forward()`
Checks `isinstance(attention_bias, GQAContext)` at the top and dispatches to `_forward_gqa()` which emits `GroupQueryAttention` directly with `do_rotary=1`, bypassing the external `RotaryEmbeddingBase.forward()` + `apply_rotary_pos_emb()` path.

### Modified: `TextModel.forward()` (`models/base.py`)
Adds an EP-driven dispatch at graph construction time:
```python
caps = ep_capabilities()
use_gqa = (
    attention_mask is not None
    and get_build_dtype() in caps.gqa_dtypes
    and caps.supports_fused_rope
    and isinstance(self.rotary_emb, BaseRope)
)
```
When True: calls `self.rotary_emb(op, position_ids)` to realize `cos_cache`/`sin_cache` as ONNX initializers (discards result), then builds `GQAContext` with the raw parameter tensors and computes `seqlens_k`/`total_seq_len` from `attention_mask`.

## What stays unchanged

- `RotaryAttentionToGQA` rewrite rule — kept as fallback for:
  - Qwen3.5 with `_MRopeBase` 3D mRoPE (`isinstance(..., BaseRope)` is True but `supports_fused_rope=False` on affected EPs excludes it)
  - DML EP (`supports_fused_rope=False`)  
  - Any model not using `TextModel` (VLMs, Qwen3.5, etc.)
- Graph I/O: `position_ids` remains in the graph inputs for consistency with the rewrite-rule path (also leaves `position_ids` as a dead input post-optimization in both paths)

## Tests

Five new tests in `components/_attention_test.py::TestGQAContextDispatch`:
- `test_gqa_context_emits_group_query_attention`: component-level, verifies GQA is emitted
- `test_gqa_context_respects_rotary_interleaved`: checks `rotary_interleaved` attribute
- `test_standard_attention_when_no_gqa_context`: standard path unaffected
- `test_build_with_cuda_ep_emits_gqa_directly`: CUDA EP + f16 → GQA present, Attention absent
- `test_build_with_default_ep_uses_standard_attention`: default EP → standard Attention

All 2327 tests pass.

### DIFF
--- a/src/mobius/components/_attention.py
+++ b/src/mobius/components/_attention.py
@@ -33,8 +33,9 @@ class GQAContext(NamedTuple):
     (e.g. Qwen3.5 with 3D mRoPE).
 
     Fields:
-        seqlens_k: Per-batch actual KV sequence length ``[batch]`` INT32.
-            Computed as ``ReduceSum(attention_mask, axis=1) - 1``.
+        seqlens_k: Per-batch last valid KV index ``[batch]`` INT32.
+            Computed as ``ReduceSum(attention_mask, axis=1) - 1``; this is a
+            0-based index into the valid KV tokens, not the KV length itself.
         total_seq_len: Scalar total sequence length INT32.
             Computed as ``Shape(attention_mask)[1]``.
         cos_cache: Full cosine RoPE table ``[max_seq_len, rotary_dim]`` FLOAT.
--- a/src/mobius/components/_attention.py
+++ b/src/mobius/components/_attention.py
@@ -4,8 +4,9 @@
 from __future__ import annotations
 
 import math
-from typing import TYPE_CHECKING, NamedTuple
+from typing import NamedTuple
 
+import onnx_ir as ir
 from onnxscript import nn
 from onnxscript._internal import builder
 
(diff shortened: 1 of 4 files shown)


src/mobius/components/_attention.py
```diff
@@ -14,8 +15,38 @@
...
+    GQA for the current build dtype.  Passed through DecoderLayer as the
+    ``attention_bias`` argument so that :class:`Attention` can detect it and
+    emit ``GroupQueryAttention`` directly instead of the generic
+    ``Attention + RotaryEmbedding`` sequence.
+
+    Using this context skips the post-hoc
+    :class:`~mobius.rewrite_rules._group_query_attention.RotaryAttentionToGQA`
+    rewrite rule for models that use the standard :class:`TextModel` backbone.
+    The rewrite rule remains as a fallback for models with non-standard RoPE
+    (e.g. Qwen3.5 with 3D mRoPE).
```
justinchuby, line 33: Why do we need to rewrite for 3D mRoPE? What's the difference? @copilot

justinchuby: @kunal-vaishnavi does this sound reasonable?

justinchuby: @copilot would it be a good idea to just set do_rotary=0 here? What's missing?

justinchuby: Let's do it in a separate PR.

PR #172, opened 2026-04-16
Fix Gemma4 CUDA EP support and audio encoder accuracy

justinchuby:
### Changes

**CUDA EP compatibility**
- Lower ONNX opset 24 → 23 for CUDA/TRT EPs (ORT ≤1.24.x lacks opset-24 kernels). Gated by `ort_lower_opset_for_ep` feature flag (default: on).
- Restructure Gemma4 per-layer embedding from one `[V, L*D]` table (262144×8960 = 2.35B elements) to L separate `[V, D]` tables (262144×256 = 67M each). Avoids ORT CUDA Gather int32 overflow ([onnxruntime#28107](https://github.com/microsoft/onnxruntime/issues/28107)) without any sharding workaround — each layer does its own small Gather directly.
- Add generic `ort_shard_large_gathers` flag + `Embedding.shard_weight_dict()` safety net for other models that may hit the same int32 limit.

**Audio encoder accuracy**
- Implement `ClippableLinear` component with learned input/output clamping buffers (HF `Gemma4ClippableLinear`). Fixes audio encoder divergence: max diff 52.68 → 0.0003.
- Fix off-by-one in `_build_causal_window_mask`: `context_left → context_left - 1` to match HF behavior.

**Example script (`examples/gemma4_multimodal.py`)**
- Fix audio feature extraction: use `Gemma4AudioFeatureExtractor` (not `WhisperFeatureExtractor`)
- Add `--compare-hf` support for audio and vision+audio modes
- Align `--ep`/`--device` flags with established CLI pattern

### Testing
- All modes (text, vision, audio, vision+audio) verified on CUDA
- Audio `--compare-hf --device cuda`: max diff ~0.0003 (float32)
- Unit tests: 2329 passed (no regressions)
- Lint clean


examples/gemma4_multimodal.py
```diff
@@ -962,12 +965,13 @@ def main() -> int:
     # and returns a ModelPackage containing all sub-models.
     # ------------------------------------------------------------------
     load_weights = not args.no_weights
-    print(f"Building ONNX models from {args.model_id!r} (dtype={args.dtype}) ...")
+    ep = args.ep or args.device
```
justinchuby: Device and ep should be separate

PR #189, opened 2026-04-22
Refactor ir.Function builders to use build_function from onnxscript 0.7.0

gramalingam:
## Summary

Replaces manual `ir.Graph` + `GraphBuilder` + `ir.Function` construction boilerplate with the new `builder.build_function()` utility from onnxscript 0.7.0. This eliminates ~25 lines of setup per function and automatically handles initializer lifting to Constant nodes.

## Changes

- **pyproject.toml**: Bump onnxscript requirement to `>=0.7.0`
- **causal_conv.py**: Extract body into trace function; remove resolved TODO
- **linear_attention.py**: Extract body into trace function; preserve per-variant input arity (see note below)
- **packed_multi_head_attention.py**: Extract body; use `op.Attention(...)` with `ir.Attr` kwargs for `ref_attr_name` forwarding (eliminates raw `ir.Node` construction)
- **skip_layer_normalization.py**: Both functions refactored; use `op.LayerNormalization`/`op.RMSNormalization` with `ir.Attr` kwargs for `ref_attr_name` (eliminates raw `ir.Node` construction and outdated docstring note)

## Key findings

- **OpBuilder supports `ir.Attr` with `ref_attr_name` as kwargs** — it flows through to `ir.node()` which accepts `ir.Attr` in `SupportedAttrTypes`. This eliminates the need for raw `ir.Node` construction in all three files that previously required it.
- **LinearAttention keeps per-variant arity** because `onnx-shape-inference` (Python package) does a strict `len(f.inputs) != len(node.inputs)` check. ONNX's own C++ shape inference (`onnx.shape_inference.infer_shapes`) accepts trailing optional inputs fine, even in strict mode. A TODO is added to investigate relaxing this upstream.

## Testing

All 2558 tests pass, 39 skipped (unchanged from baseline).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,7 +17,7 @@ dependencies = [
     "numpy>=1.24.0",
     "onnx_ir>=0.1.0",
     "onnx-shape-inference==0.1.7",
-    "onnxscript>=0.6.0.dev",
+    "onnxscript>=0.7.0",
     "torch>=2.1.0",
     "tqdm",
 ]
--- a/src/mobius/functions/causal_conv.py
+++ b/src/mobius/functions/causal_conv.py
@@ -20,9 +20,6 @@ from mobius._constants import OPSET_VERSION
 DOMAIN = "com.microsoft"
 
 
-# TODO(justinchuby): Simplify function creation boilerplate
-
-
 def causal_conv_nd_with_state(
     *,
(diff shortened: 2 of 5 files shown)


justinchuby (approved):
Thanks! This reminds me we can also update the call sites now.

PR #207, opened 2026-04-24
feat: add input_features_mask to Gemma4 audio encoder

justinchuby:
Add bool mask input (`input_features_mask: BOOL [B, T]`) to the audio encoder ONNX graph for proper variable-length audio handling in batched inference.

## Changes

- **Conv subsampling** (`Gemma4ConvSubsampling`): Accepts `input_features_mask`, zeros out padded positions before each conv layer, downsamples the mask by stride 2 after each stage (matching HF `Gemma4AudioSubSampleConvProjectionLayer`)
- **Attention** (`Gemma4Attention`): Accepts optional `attention_mask` and applies additive padding bias to prevent attending to padded key positions
- **Encoder layer/encoder** (`Gemma4AudioLayer`, `Gemma4AudioEncoder`): Thread the mask through the full Conformer stack
- **Task** (`Gemma4Task._build_audio`): Adds `input_features_mask` as a graph input
- **Tests**: Updated to verify the new input

Matches the onnx-community/gemma-4-E2B-it-ONNX audio encoder interface.


apsonawane (approved):

src/mobius/tasks/_gemma4.py
```diff
@@ -315,13 +316,25 @@ def _build_audio(
...
 
-        audio_features = audio(op, input_features)
+        audio_features = audio(
+            op,
+            input_features,
+            input_features_mask=input_features_mask,
+        )
         audio_features.name = "audio_features"
         graph.outputs.append(audio_features)
+        # TODO: For batched variable-length audio, expose the downsampled
```
justinchuby: This sounds useful?

justinchuby: @apsonawane what do you think

src/mobius/components/_gemma4_audio.py
```diff
@@ -191,11 +191,64 @@ def _conv_norm_relu(
...
+        """
+        if mask is None:
+            return x, None
+        # Broadcast mask [B, T] → [B, 1, T, 1] over C and F dims
+        mask_4d = op.Unsqueeze(mask, [1, 3])  # [B, 1, T, 1]
+        x = op.Mul(x, op.CastLike(mask_4d, x))
+        # Downsample mask by conv stride (2): mask[:, ::2]
+        mask = op.Slice(
+            mask,
+            op.Constant(value_ints=[0]),  # starts
```
gramalingam: nit: op.Constant not required, I think. Just `[0]` should work. I think it should also work as a kwarg `starts=[0]`.

gramalingam:
I think it would be helpful to clarify whether the mask can be arbitrary or if there is an assumption/requirement that it is all Trues up to some length and all False afterwards.

src/mobius/components/_gemma4_audio.py
```diff
@@ -539,6 +606,16 @@ def forward(self, op: builder.OpBuilder, x: ir.Value):
         window_mask = self._build_causal_window_mask(op, seq_len)
         scores = op.Add(scores, window_mask)
 
+        # Padding mask: block attention to padded key positions
+        if attention_mask is not None:
+            # attention_mask [B, T] bool → additive bias [B, 1, 1, T]
+            pad_bias = op.Where(
+                op.Unsqueeze(attention_mask, [1, 2]),
+                op.Constant(value_float=0.0),
```
gramalingam, line 614: Is `scores` guaranteed to be float? If not, we might need a `CastLike(0.0, scores)` here. nit: `op.Constant` is not needed.

justinchuby: Let me test different dtypes

src/mobius/components/_gemma4_audio.py
justinchuby: Done

src/mobius/components/_gemma4_audio.py
justinchuby: Confirmed that the audio part is float32 to preserve accuracy. Same as HF

justinchuby:
Examples validated

PR #208, opened 2026-04-24
refactor: split per_layer_inputs to embedding, remove input_ids from decoder

justinchuby:
Match onnx-community Gemma4 interface: embedding outputs per_layer_inputs, decoder consumes it instead of input_ids. Multimodal fusion moves to runtime.

## Changes

**Embedding model** (input_ids → inputs_embeds + per_layer_inputs):
- Removed multimodal fusion (image_features, audio_features inputs)
- Added per-layer embedding computation when `hidden_size_per_layer_input > 0`
- Outputs `per_layer_inputs [B, S, L, D]` alongside `inputs_embeds`

**Decoder model** (inputs_embeds + per_layer_inputs → logits):
- Removed `input_ids` from graph inputs
- Added `per_layer_inputs [B, S, L, D]` as input (when per-layer enabled)
- Slices per_layer_inputs tensor and passes per-layer list to text model

**Weight routing** in `Gemma4Model.preprocess_weights`:
- Per-layer weights now route to `embedding.*` instead of `decoder.model.*`

**Tests**:
- Updated all gemma4 test assertions for new I/O signatures
- Added `test_gemma4_per_layer_inputs_graph` for per-layer enabled config
- Updated ORT GenAI config tests

### DIFF
--- a/src/mobius/models/gemma4.py
+++ b/src/mobius/models/gemma4.py
@@ -1370,13 +1370,17 @@ class Gemma4TextModel(nn.Module):
         position_ids: ir.Value,
         past_key_values: list | None = None,
         inputs_embeds: ir.Value | None = None,
+        per_layer_inputs: list[ir.Value] | None = None,
     ) -> tuple[ir.Value, list]:
         if inputs_embeds is not None:
             hidden_states = inputs_embeds
         else:
             hidden_states = self.embed_tokens(op, input_ids)
 
-        per_layer_inputs = self._compute_per_layer_inputs(op, input_ids, hidden_states)
+        # Use pre-computed per-layer inputs when provided (multimodal decoder
+        # path); otherwise compute from input_ids + hidden_states (text-only).
+        if per_layer_inputs is None:
+            per_layer_inputs = self._compute_per_layer_inputs(op, input_ids, hidden_states)
 
         # Determine whether to emit GroupQueryAttention directly.
         # GQA fuses RoPE + attention + KV cache into a single op, and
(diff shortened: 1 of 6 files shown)


apsonawane (approved):

src/mobius/models/gemma4.py
```diff
@@ -1809,72 +1809,47 @@ def __init__(self, config: Gemma4Config):
...
-
-        # Scatter audio features at audio-token positions (only for AnyToAny models)
-        if audio_features is not None:
-            assert self.audio_token_id is not None, (
-                "Gemma4EmbeddingModel received audio_features but audio_token_id is not set. "
-                "Ensure config.audio is provided."
+        # Mask multimodal token IDs to pad_token_id (0) before embedding.
+        # HF's Gemma4Model replaces image/audio tokens with pad_token_id
+        # before calling embed_tokens, so inputs_embeds at multimodal
+        # positions contain PAD embeddings (later replaced by runtime
```
justinchuby, line 1828: What do you mean

justinchuby:
@apsonawane I wonder if it is better to keep the audio/image embedding fusion logic in the model. This way genai doesn't need to know how to do this?

src/mobius/models/gemma4.py
```diff
@@ -1809,72 +1809,47 @@ def __init__(self, config: Gemma4Config):
...
+        # fusion).
+        masked_ids = input_ids
+        pad = op.Constant(value_int=0)
+        if self._image_token_id:
+            masked_ids = op.Where(
+                op.Equal(
+                    masked_ids,
+                    op.Constant(value_int=self._image_token_id),
+                ),
+                pad,
```
gramalingam, line 1838: nit: using `0` instead of `op.Constant(value_int=0)` will cast it to same type as masked_ids/input_ids. (In case it can be int value of different precision). Is it always int64?

gramalingam:
> @apsonawane I wonder if it is better to keep the audio/image embedding fusion logic in the model. This way genai doesn't need to know how to do this?

Sounds useful if doable without much cost. When/where does it happen in genai? Is that extra logic between invocation of two onnx models?

gramalingam (approved):

justinchuby:
I am closing this PR after a discussion with @apsonawane . To match what hf is doing for transformers.js, the embedding merge logic needs to be hard coded in the runtime, which is not desirable. Instead, we keep the original architecture where this is handled in the model.

PR #216, opened 2026-04-24
Update onnx-shape-inference dependency version and add transformers

justinchuby:
Also add the transformers dependency


justinchuby:
@titaiwangms

titaiwangms (approved):

PR #223, opened 2026-04-27
Update gemma4 genai audio config logic

apsonawane:
### Summary

Adds audio encoder export support for Gemma4 multimodal models in the ORT-GenAI integration pipeline. This enables Gemma4 any-to-any (vision+audio+text) models to be exported with the correct audio processor configuration and genai_config.json speech section.

#### `auto_export.py`

- **`_write_audio_processor_config()`** — New function that writes `audio_feature_extraction.json` for Gemma4 models using the `feature_extraction.sequence` schema required by `OrtxCreateSpeechFeatureExtractor`. The pipeline consists of `AudioDecoder` → `Gemma4LogMel` with USM-style 128-dim log-mel spectrogram parameters.
- **`_write_genai_config()`** — Gemma4 speech section overrides:
  - `filename`: `audio_encoder/model.onnx`
  - `config_filename`: `audio_feature_extraction.json` (instead of default `audio_processor.json`)
  - `input_names` mapping: `audio_embeds` → `input_features`, `attention_mask` → `input_features_mask`
- **`_write_vision_processor_config()`** — Renamed from `_write_processor_config` for clarity. Gemma4 vision path now writes `image_processor.json` with `DecodeImage` → `Gemma4ImageTransform` transforms pipeline, checked before `has_speech` to ensure vision config is always written for VLM+audio models.
- **`write_ort_genai_config()`** — Calls `_write_audio_processor_config()` and includes result in the output dict under `"audio_processor"`.

#### `auto_export_test.py`

Added 6 new tests covering audio export:

| Test | What it verifies |
|------|-----------------|
| `test_audio_no_audio_returns_none` | `_write_audio_processor_config` returns `None` when config has no `audio` attribute |
| `test_audio_non_gemma4_returns_none` | Returns `None` for non-Gemma4 model types |
| `test_audio_gemma4_writes_feature_extraction_json` | Writes `audio_feature_extraction.json` with correct `feature_extraction.sequence` schema, `AudioDecoder` + `Gemma4LogMel` ops, and expected attrs |
| `test_gemma4_audio_processor_json_written` | End-to-end: `write_ort_genai_config` produces `audio_feature_extraction.json` |
| `test_gemma4_audio_processor_not_written_without_audio` | No audio config written when config lacks `audio` |
| `test_gemma4_genai_config_speech_section` | `genai_config.json` speech section has correct `config_filename`, `filename`, and input mapping (`input_features` / `input_features_mask`) |

### Notes

[...]

PR #225, opened 2026-04-28
Add unit tests and simplify condition

apsonawane:
follow up from this PR: https://github.com/onnxruntime/mobius/pull/223


src/mobius/integrations/ort_genai/auto_export.py
```diff
@@ -353,14 +353,13 @@ def _write_genai_config(
...
             model_type = getattr(config, "model_type", "")
-            if model_type in _GEMMA4_MODEL_TYPES:
-                vision_cfg = getattr(config, "vision", None)
-                sms = getattr(vision_cfg, "spatial_merge_size", 2)
-                vision_kwargs["spatial_merge_size"] = sms
-                vision_kwargs["config_filename"] = "image_processor.json"
-            elif has_speech:
-                vision_kwargs["spatial_merge_size"] = None
+            if model_type in _GEMMA4_MODEL_TYPES or has_speech:
                 vision_kwargs["config_filename"] = "image_processor.json"
```
justinchuby: Is this implying that the image config file name is different when the model has speech? Should it be standardized and stay the same whether we have speech or not?

justinchuby (approved):

rui-ren (approved):

src/mobius/integrations/ort_genai/auto_export.py
apsonawane: Yeah true, we do not need to set config_filename, it defaults to image_processor updated it to differ from speech

PR #231, opened 2026-05-01
Migrate task files to new GraphBuilder input/output APIs

justinchuby:
## Summary

Migrate all 24 task files under `src/mobius/tasks/` to use the new onnxscript `GraphBuilder.input()` and `GraphBuilder.add_output()` APIs, replacing manual `ir.Value()` construction and `graph.outputs.append()` patterns.

## Changes

- **`_base.py`**: `_make_graph()` no longer takes an inputs parameter. Inputs are now created via `builder.input(name, dtype, shape)` which auto-registers them on the graph.
- **`_cache_utils.py`**: Cache helper functions (`_make_kv_cache_inputs`, `_make_hybrid_cache_inputs`, `_register_kv_cache_outputs`, `_register_hybrid_cache_outputs`) now take `builder` as first parameter. Return types simplified — flat input lists removed since `builder.input()` auto-registers.
- **All 24 task files**: `ir.Value(name=..., shape=ir.Shape(...), type=ir.TensorType(...))` → `builder.input(name, dtype=..., shape=[...])`; `value.name = name; graph.outputs.append(value)` → `builder.add_output(value, name)`
- **Import cleanup**: `GraphBuilder` imported from public `onnxscript` namespace instead of `onnxscript._internal.builder`
- **Naming consistency**: Standardized `builder` variable name across all files

## Impact

- 24 files changed, 571 insertions, 903 deletions (net -332 lines)
- Eliminates dual-bookkeeping anti-pattern (separate input list + graph registration)
- Input ordering structurally guaranteed by call order
- Output registration is atomic (single call vs two-step name+append)
- All 2662 tests pass

## Note

`op.call()` migration for module/function invocations is out of scope — `op.call()` currently only accepts `ir.Function`, not `nn.Module`, and treats kwargs as ONNX attributes rather than tensor inputs. This requires upstream onnxscript changes.

### DIFF
--- a/src/mobius/tasks/_adapter.py
+++ b/src/mobius/tasks/_adapter.py
@@ -23,33 +23,23 @@ class AdapterTask(ModelTask):
         module,
         config,
     ) -> ModelPackage:
+        graph, builder = _make_graph()
+        op = builder.op
+
         # Determine input shape based on adapter type
         if hasattr(config, "in_channels"):
             # T2I-Adapter: conditioning image input
-            condition = ir.Value(
-                name="condition",
-                type=ir.TensorType(ir.DataType.FLOAT),
-                shape=ir.Shape(("batch", config.in_channels, "height", "width")),
-            )
+            condition = builder.input("condition", dtype=ir.DataType.FLOAT, shape=["batch", config.in_channels, "height", "width"])
         else:
             # IP-Adapter: image embedding input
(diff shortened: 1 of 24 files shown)


justinchuby:
@titaiwangms @gramalingam

titaiwangms (approved):

gramalingam (approved):

PR #232, opened 2026-05-01
Migrate custom op callsites to op.call() with ir.Function

justinchuby:
## Summary

Migrate 7 custom ONNX op callsites to use `op.call(ir_function, *args, **attrs)` instead of `op.CustomOpName(..., _domain="com.microsoft")`. This auto-registers function definitions in the graph, eliminating manual function registration.

### Component callsites migrated (7 total)
- **PackedMultiHeadAttention** (2 sites): `_qwen25_vl_vision.py`, `_qwen3_vl_vision.py`
- **CausalConvWithState** (2 sites): `_gated_deltanet.py`, `_mamba_block.py`
- **LinearAttention** (3 sites): `_gated_deltanet.py`, `_mamba_block.py`, `_lightning_attention.py`

### Pattern
All functions are cached at `__init__` time and reused across forward calls. ONNX attributes are passed as `op.call()` kwargs so they appear on the calling node:
- **LinearAttention**: `scale`, `q_num_heads`, `kv_num_heads`, `update_rule`
- **CausalConvWithState**: `activation`
- **PackedMultiHeadAttention**: `num_heads`, `scale`

### Infrastructure
- `_make_model()` accepts `functions: Iterable[ir.Function]` to transfer auto-registered functions from `builder.functions.values()` to the model
- Removed manual function registration (`_register_linear_attention_functions`, `_register_linear_attention_functions_for_ssm2`) — ~145 lines of boilerplate deleted
- All 42 `_make_model()` callsites updated to pass `builder.functions.values()`

### Unit tests (build_graph_test.py)
```
1185 passed, 40 skipped in 11.97s
```

### Full test suite (build_graph + co-located + CLI)
```
2662 passed, 41 skipped, 74 warnings in 17.84s
```

### Generation tests (integration_test.py -k generation)
```
18 passed, 5 skipped in 1254.55s
```
Key results:
- Qwen2.5-VL-3B (uses PackedMHA via op.call): **ONNX matches Torch exactly**
- Qwen3-VL-2B (uses PackedMHA via op.call): **ONNX matches Torch exactly**
- LLaMA, Qwen2, Phi3, Gemma2, Mistral generation with KV cache: all pass

### Example scripts
- `build_and_save.py --model Qwen/Qwen2.5-0.5B --no-weights` ✅
- `build_and_save.py --model Qwen/Qwen2.5-VL-3B-Instruct --no-weights` ✅ (uses PMHA)
- `build_and_save.py --model meta-llama/Llama-3.2-1B --no-weights` ✅
- `text_generation.py --model Qwen/Qwen2.5-0.5B` ✅ (coherent output)

## Design decisions

- **Parametric functions cached at init**: CausalConvWithState and LinearAttention are parametric (kernel_size, channels, num_heads baked into the function body) because these values are used as graph-structural constants that cannot be ONNX attribute references. Functions are built once at `__init__` and reused.
- **PackedMultiHeadAttention is static**: Its attributes forward to the inner Attention op via `ref_attr_name`. Also cached at `__init__` for consistency.
- **op.call() does not mutate ir.Function objects**: Verified empirically. Caching is safe; each model gets its own copy via `_make_model()`.


justinchuby:
@gramalingam

titaiwangms (commented):
Mechanical refactor that's overall a clear win — deleting ~145 lines of bespoke registration in favor of `op.call()` auto-collecting from `builder.functions` is the right direction. Verified against `microsoft/onnxscript@1f43db38` (post-0.7.1 main): `op.call` registers in `builder.functions`, `call_inline` does not (PR #2851 made the split explicit), `build_function` already lifts initializers (PR #2889), and the removed `Opset.function_defs` API (PR #2901) is not used by mobius. Local `tests/build_graph_test.py` + `src/mobius/functions/` runs clean (1225 passed). No incoming onnxscript breakage.

A few real concerns and some smaller items.

### 🟡 Major

1. **Identifier-collision dedup is a silent footgun for parametric functions.** `_make_model` dedups by `fn.identifier()` = `(domain, name, overload)`, but `linear_attention` and `causal_conv_nd_with_state` bake `q_num_heads`, `kv_num_heads`, `kernel_size`, `channels`, `update_rule` into the function body as Python literals while exposing the same identifier `("com.microsoft", "LinearAttention", "")` regardless of params. Verified locally: two `linear_attention` instances with different head counts produce different bodies (26 vs 28 nodes) but identical identifiers — the first one wins, the rest are dropped, and any call node referencing them gets the wrong body. No current model triggers this (per-architecture configs are uniform), but a hybrid model that mixes — say — DeltaNet layers with different head counts, or two `CausalConvWithState` instances with different `kernel_size`, will silently produce a model whose function body matches only the first instance. ORT itself doesn't read the body for these custom ops, so the call attributes will save you at runtime today, but the *exported model* is misleading and any tooling that does inspect the function body will get wrong results. Either:
   - encode the parametric choices into `overload` (e.g. `overload=f"h{q}_kv{kv}_{rule}"`), or
   - assert in `_make_model` when two functions share an identifier but differ structurally, or
   - hoist function caching to the task layer with a lookup keyed on the parameter tuple.

2. **Duplicated source of truth on parametric attrs.** Component callsites pass `scale`, `q_num_heads`, `kv_num_heads`, `update_rule` to `op.call()` *and* the cached function bakes those same values into its body. Today they always agree because both come from the same `__init__`, but if anyone later edits one without the other (e.g., a bugfix that changes `head_k_dim` only at the call site), the body silently won't follow. Worth a comment at each callsite saying "must match the cached function's parameters" — or, if these values are only meant to be readable by ORT and never by the function body, drop the bake-in approach and rely on `ref_attr_name` (PMHA does this correctly).

3. **Per-instance function rebuild is wasteful.** `GatedDeltaNet`, `Mamba2Block`, `LightningAttention`, etc. each rebuild their full `linear_attention` function (Scan body, recurrence subgraph, ~25–30 nodes) at every layer's `__init__`. For a 32-layer model that's 32 redundant ir.Function constructions per build. Functionally fine because dedup keeps one — but slow on big models, and this is exactly what (1) papers over. A small per-process LRU on the function factories (or moving the cache one level up to the task) would address both.

### 🟢 Minor

4. **PMHA loses the named output.** `_outputs=["packed_attn_out"]` was removed in `_qwen25_vl_vision.py` and `_qwen3_vl_vision.py`. Auto-generated output names will now appear in the graph. If any rewrite rule, golden, or downstream test pinned that name, it'll break. Worth a quick check.

5. **`_make_model` API drift.** The previous note (and an older reviewer comment) said tasks should pass the *builder* into `_make_model(graph, builder)` so `builder.functions` is collected internally. This PR keeps the explicit `builder.functions.values()` at all 42 callsites. Both are fine but the `(graph, builder)` signature would be one less thing to forget at new callsites — easy to miss in a future task and silently emit a model with no functions. Consider an overload that accepts the builder directly.

6. **Already raised by Copilot, worth fixing before merge:**
   - SKILL.md examples (`adding-a-new-model`, `custom-functions`, `reusable-components`) call `linear_attention(...)` without the required `q_num_heads`/`kv_num_heads` — won't run as-is.
   - `self._kernel_size` is assigned but unused in `_DepthwiseConv1d` (`_gated_deltanet.py:63`) and `_Mamba2DepthwiseConv1d` (`_mamba_block.py:197`).

### ✅ Good

- Generation test evidence is solid (Qwen2.5-VL, Qwen3-VL exact ONNX↔Torch match — these are the ones that exercise PMHA via `op.call`).
- `_make_model` does the right "no overwrite" check (`if op_id not in model.functions`) so multiple sub-models in a `ModelPackage` won't trample each other.
- Removing `_register_linear_attention_functions` + `_for_ssm2` is a real boilerplate win; the `layer_types`/`Mamba2` switch logic was a maintenance hazard.
- Opset import for `com.microsoft` is still set by `_make_graph`, so dropping the explicit `model.graph.opset_imports[_FUNCTIONS_DOMAIN] = 1` is safe.

### Recommendation

**Approve with comments after the SKILL.md examples are fixed and (1)/(2) are at minimum acknowledged in code comments.** The dedup-collision behavior should be tracked even if it's not blocking — please file a follow-up issue if you'd rather not address it in this PR.

PR #236, opened 2026-05-04
Add Fun-ASR-Nano and SenseVoiceSmall support

justinchuby:
This pull request introduces support for the SenseVoiceSmall speech recognition model, including a new ASR example script and shared audio preprocessing utilities. It also expands the model registry and configuration to accommodate Fun-ASR and SenseVoiceSmall models, and exposes new SANM encoder components. The most important changes are grouped below:

**New ASR Example and Utilities:**

* Added `examples/sensevoice_small.py`, a complete CLI script for running ONNX-based speech recognition with SenseVoiceSmall, including CTC decoding, language control, and chunked inference.
* Introduced `examples/asr_utils.py` with reusable audio preprocessing functions: audio loading, log-mel fbank computation, LFR stacking, CMVN loading, and full frontend pipeline.

**Model Registry and Configuration Updates:**

* Registered `FunASRForConditionalGeneration` and `SenseVoiceSmallModel` in the model registry, with proper task types and default HuggingFace model IDs. [[1]](diffhunk://#diff-8d5c42d97d876ab3b86d0698ef084e04c65db26af0cea79f0d9dd002cf2ebf69R97) [[2]](diffhunk://#diff-8d5c42d97d876ab3b86d0698ef084e04c65db26af0cea79f0d9dd002cf2ebf69R121) [[3]](diffhunk://#diff-8d5c42d97d876ab3b86d0698ef084e04c65db26af0cea79f0d9dd002cf2ebf69R512-R522) [[4]](diffhunk://#diff-8d5c42d97d876ab3b86d0698ef084e04c65db26af0cea79f0d9dd002cf2ebf69R857-R858)
* Updated the registry to assign the "fun_asr" and "sensevoice_small" models to the "qwen" organization for fallback detection.
* Extended `AudioConfig` with additional fields for Fun-ASR/SenseVoice encoder configuration (e.g., `tp_num_blocks`, `adaptor_proj_dim`).

**Component Exposure:**

* Exported SANM encoder components (`SANMAttention`, `SANMEncoderLayer`, `SANMFFN`) in `mobius.components.__init__` and imported them from the internal module. [[1]](diffhunk://#diff-8eacab4d5682bac795dd69c12af6cfb6e6ac9198de7bf37ed06fc748a4682d0cR87-R89) [[2]](diffhunk://#diff-8eacab4d5682bac795dd69c12af6cfb6e6ac9198de7bf37ed06fc748a4682d0cR247-R251)


titaiwangms (changes requested):
Reviewed locally. Solid Fun-ASR scaffolding (SANM components, 3-model split task, golden tests, example script), but several issues and significant scope concerns.

### 🔴 Critical

1. **Stale rebase regresses Gemma4 KV head logic.** Main (#234) intentionally made `num_global_key_value_heads` apply whenever `is_sliding == False` (independent of `attention_k_eq_v`) so Gemma4 GGUF — which sets per-layer KV heads but never sets `attention_k_eq_v` — works. This PR reverts that to gate on `_use_alternative_attention`, which will silently break Gemma4 GGUF inference (full layers will fall back to `num_key_value_heads` and KV cache shapes will mismatch the weights). Same regression in `tasks/_gemma4.py::_make_gemma4_kv_cache_inputs`. The gemma4 hunks shouldn't be part of this PR.

2. **PR is `mergeable_state: dirty`** — there are merge conflicts with main. Rebase is required, and that rebase is the same operation that should drop the gemma4 reverts above.

### 🟡 Major

3. **Scope creep — SenseVoiceSmall is a separate model.** The PR title/body advertise Fun-ASR only, but it also adds `models/sensevoice_small.py`, `tasks/_audio_ctc.py`, `examples/sensevoice_small.py`, two golden files, and registry entries. Two unrelated models in one PR makes review and revert harder; recommend splitting SenseVoiceSmall into a follow-up PR.

4. **Multi-batch audio fusion is silently incorrect.** `FunASREmbeddingModel.forward` builds gather indices via `CumSum(axis=1)` over a per-batch mask, but `audio_features` is a single flat `(num_audio_tokens, hidden)` table shared across the batch. With `batch > 1`, each row's cumsum restarts at 1, so all rows index the *same* prefix of `audio_features` — wrong audio is scattered into rows ≥ 1. Either assert `batch == 1`, or document the constraint, or compute per-batch offsets. Today it works only because ASR inference is single-utterance; nothing in the graph enforces it.

### 🟢 Minor / Nits

5. **`_skip_layer_norm.py` rank guard skips unknown shapes.** `if inp.shape is not None` means an Add input with no static shape passes the guard and can still produce a 1-D skip that ORT rejects at runtime. Either fail-closed when the shape is unknown, or document the assumption.

6. **`SANMAttention` and `AdaptorAttention` ignore attention masks.** No padding mask is plumbed through the encoder; for the example pipeline this is OK because audio is right-padded zeros and SANM is local enough, but if anyone later batches variable-length audio the encoder will attend across pad frames silently.

7. **Hardcoded fallback `audio_token_id = 151676`** in `FunASREmbeddingModel.__init__` — fine as a default for this checkpoint, but worth a `# TODO` or making the absence loud.

8. `_FSMNBlock` uses `(kernel - 1) // 2` left-pad — asymmetric for even kernels. Default 11 is odd so it's symmetric; just a footgun if kernel becomes configurable.

### ✅ Good

- SANM components are clean, well-commented, model-agnostic, and follow the codebase's `op`-as-first-arg + `_test.py` colocation conventions.
- Weight-name alignment is mostly structural (`nn.ModuleList`, matching attribute names) with only a small `preprocess_weights` rename map.
- FSMN conv-weight transpose `(C, K, 1) → (C, 1, K)` is detected by shape rather than name — robust.
- Golden L4/L5 tests + 20 unit tests with deterministic seeds is good coverage.
- Architecture-diff bot reports zero changes to existing model graphs ✅.

### Recommendation

**Request changes.** Rebase onto current main, drop the gemma4 + `_gemma4` task hunks (they regress GGUF), decide whether SenseVoiceSmall should be a separate PR, and either fix or document the batch>1 audio-fusion limitation.

titaiwangms (commented):
Update LGTM on the items I raised earlier — thanks for the quick turnaround. A few remaining concerns from this revision:

### Confirmed addressed ✅
- KV-head regression in Gemma4 attention/task is gone — diff vs main no longer touches the `_use_alternative_attention` gating logic. 👍
- batch=1 limitation now in `FunASREmbeddingModel` class docstring.
- `audio_token_id=0` collision documented in the module docstring.
- SkipLayerNorm rank-guard now has a comment explaining unknown-shape Adds intentionally pass through.

### Still open

1. **Gemma4 MoE refactor still slipping in.** The new `_remap_moe_expert_weights` helper (gemma4.py +33 / -38) is unrelated to Fun-ASR — Fun-ASR uses a Qwen3 decoder. It also introduces a behavioral change for `Gemma4CausalLMModel.preprocess_weights`: the old code's `if ".router.scale" in key:` now becomes `... and ".per_expert_scale" not in key`, so any `.router.per_expert_scale` weights are no longer multiplied by `hidden_size^-0.5`. That may well be the right fix (matches the path `Gemma4Model` already had after #239), but it should land in a separate PR with its own test rather than ride along here. If you do keep it, please add a one-line note in the PR body and a test that pins the new behavior for `Gemma4CausalLMModel`.

2. **Title typo:** "Add Fun-ASR-Nano and SneseAudio-ASR support" → presumably "SenseVoice-ASR" (or "SenseVoiceSmall").

3. **batch=1 is documented but not enforced.** The new docstring is good, but a runtime guard would prevent a future caller from silently producing wrong outputs:
   ```python
   # In FunASREmbeddingModel.forward, after computing batch dim
   batch = op.Shape(input_ids, start=0, end=1)
   # ... add a check or rely on an op that fails for batch > 1
   ```
   At minimum, add an `assert` in the example pipeline (`examples/fun_asr.py`) and a unit test that verifies the model raises/returns wrong shapes for batch>1 so the constraint is mechanically enforced somewhere.

### Recommendation
Drop or split out the Gemma4 hunk and fix the title typo; then I'm comfortable approving. The batch>1 enforcement is nice-to-have, not blocking.

justinchuby:
@titaiwangms Thanks - will handle batch>1 in a follow up. The gemma4 changes should have already been removed.

titaiwangms (commented):
@justinchuby You're right — apologies, my local `main` was stale at `353dcdfb`. After `git fetch origin main` (now at `d39b13b8` from #240), `git diff origin/main..pr-236 -- src/mobius/models/gemma4.py src/mobius/tasks/_gemma4.py` is empty. No Gemma4 changes in this PR. False alarm on my part.

Remaining items from my last review reduce to:
- Title typo ("SneseAudio" → "SenseVoice").
- batch>1 enforcement — acknowledged as follow-up. 👍

Otherwise LGTM.

titaiwangms (approved):

PR #237, opened 2026-05-04
Fix Gemma4 weight mapping for 26b MoE and 31b alternative attention

justinchuby:
## Summary

Two weight mapping fixes for Gemma4 models that use `attention_k_eq_v=True` (26b-a4b, 31b):

### Fix 1: Alternative attention (V=K) for full_attention layers
When `attention_k_eq_v=True`, full-attention layers derive V from K (no separate `v_proj`), matching HF `Gemma4TextAttention`. Changes:
- Skip `v_proj` creation for alternative-attention layers
- In forward: V = raw `k_proj` output (before k_norm/RoPE), then v_norm applied
- Per-layer KV head count: full_attention uses `num_global_key_value_heads`
- Update task KV cache inputs with per-layer head counts

### Fix 2: MoE expert weight mapping in Gemma4Model
`Gemma4Model` (multimodal) had its own `preprocess_weights` that was missing:
- Expert weight rename (`experts.gate_up_proj` → `fc1_experts_weights`)
- Router scale folding (`hidden_size^-0.5`)

These were only in `Gemma4CausalLMModel.preprocess_weights`.

### Testing
- 12/12 Gemma4 L1 tests pass
- 2662/2662 full suite pass
- Verified: decoder exports successfully for both 26b-a4b and 31b models

### Models affected
| Model | Issue | Fixed |
|-------|-------|-------|
| gemma-4-26b-a4b[-it] | MoE expert weights + V=K | ✅ |
| gemma-4-31b[-it] | V=K alternative attention | ✅ |
| gemma-4-e2b[-it] | No issue (no k_eq_v) | ✅ Unaffected |
| gemma-4-e4b[-it] | No issue (no k_eq_v) | ✅ Unaffected |

### DIFF
--- a/src/mobius/_configs.py
+++ b/src/mobius/_configs.py
@@ -1774,6 +1774,7 @@ class Gemma4Config(VisionLanguageConfig):
     final_logit_softcapping: float = 0.0
     attn_logit_softcapping: float = 0.0
     enable_moe_block: bool = False
+    attention_k_eq_v: bool = False
     boa_token_id: int | None = None
 
     @classmethod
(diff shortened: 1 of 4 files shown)


justinchuby:
Replaced by https://github.com/onnxruntime/mobius/pull/239

PR #241, opened 2026-05-04
Add gqa_max_head_dim feature flag for configurable GQA limit

justinchuby:
## Summary

Make the Attention→GroupQueryAttention rewrite rule head_dim limit configurable via a feature flag instead of a hardcoded constant.

## Problem

The GQA rewrite rule had a hardcoded `_MAX_GQA_HEAD_DIM = 256` that prevented GQA fusion for Gemma4 global attention layers (head_dim=512). PR #240 intended to change this to 512 but the value was not actually updated.

## Solution

Replace the constant with `flags.gqa_max_head_dim` (default: 512), configurable via:
- Environment variable: `MOBIUS_GQA_MAX_HEAD_DIM=256`
- Programmatic: `override_flags(gqa_max_head_dim=256)`

## Changes

- **`_flags.py`**: Add `_env_int` helper, `gqa_max_head_dim` flag (default 512), update `override_flags` type hint
- **`_group_query_attention.py`**: Replace `_MAX_GQA_HEAD_DIM` constant with `flags.gqa_max_head_dim`
- **`_flags_test.py`**: 6 new tests (default value, env var, override, integration with GQA rewrite)

## Testing

2674 tests pass, lint clean.


src/mobius/_flags_test.py
```diff
@@ -108,3 +108,51 @@ def test_returns_snapshot_not_live_view(self):
         original = _flags.flags.suppress_dedup_warning
         with _flags.override_flags(suppress_dedup_warning=not original):
             assert snapshot["suppress_dedup_warning"] == original
+
+
+class TestGqaMaxHeadDim:
+    """Tests for the gqa_max_head_dim integer flag."""
+
+    def test_default_is_512(self, monkeypatch):
+        monkeypatch.delenv("MOBIUS_GQA_MAX_HEAD_DIM", raising=False)
```
titaiwangms: We set it as ENV? Why not just hard code it to 512 since it's a "max" value anyway.

justinchuby:
Closing this PR. After further discussion, the feature flag is not needed — we will hardcode MAX_HEAD_DIM=512 since older ORT versions have other CUDA issues that prevent usage anyway, and CPU EP is unaffected by head_dim. The 512 limit is already set in #240.

PR #242, opened 2026-05-04
Add Foundry Local deployment skill

justinchuby:
Document how to deploy mobius-exported ONNX models to Microsoft Foundry Local.

## What this adds

New skill at `.agents/skills/foundry-local/SKILL.md` covering:

- **Custom model registration** — cache directory structure, file copying, verification
- **inference_model.json format** — field reference + common prompt templates (ChatML, Llama, Gemma)
- **Running models** — CLI (`foundry model run`), OpenAI-compatible API, Python SDK
- **Known limitations** — bundled GenAI version, no CUDA in pip SDK, multimodal requirements, model_type workaround
- **End-to-end example** — complete workflow from `mobius build` to local inference

Reference: https://techcommunity.microsoft.com/blog/educatordeveloperblog/deploying-custom-models-with-microsoft-olive-and-foundry-local/4489002

PR #243, opened 2026-05-04
Add Foundry Local deployment skill

justinchuby:
Document how to deploy mobius-exported ONNX models to Microsoft Foundry Local.

## What this adds

New skill at `.agents/skills/foundry-local/SKILL.md` covering:

- **Custom model registration** — cache directory structure, file copying, verification
- **inference_model.json format** — field reference + common prompt templates (ChatML, Llama, Gemma)
- **Running models** — CLI (`foundry model run`), OpenAI-compatible API, Python SDK
- **Known limitations** — bundled GenAI version, no CUDA in pip SDK, multimodal requirements, model_type workaround
- **End-to-end example** — complete workflow from `mobius build` to local inference

Reference: https://techcommunity.microsoft.com/blog/educatordeveloperblog/deploying-custom-models-with-microsoft-olive-and-foundry-local/4489002

Replaces #242 (closed due to wrong branch name).

### DIFF
--- a/.agents/skills/building-ort-genai/SKILL.md
+++ b/.agents/skills/building-ort-genai/SKILL.md
@@ -29,19 +29,19 @@ Use this skill when:
 ## Step 1: Install CUDA toolkit
 
 ```bash
-# CUDA 12.8
-wget https://developer.download.nvidia.com/compute/cuda/12.8.1/local_installers/cuda_12.8.1_570.124.06_linux.run
-sudo sh cuda_12.8.1_570.124.06_linux.run \
-  --toolkit --toolkitpath=$HOME/cuda12.8 \
+# CUDA 13.0
+wget https://developer.download.nvidia.com/compute/cuda/12.8.1/local_installers/cuda_13.0.1_575.51.03_linux.run
+sudo sh cuda_13.0.1_575.51.03_linux.run \
+  --toolkit --toolkitpath=$HOME/cuda13.0 \
   --silent --override --no-man-page
 
-# Or CUDA 13.0 if available — adjust the URL and path accordingly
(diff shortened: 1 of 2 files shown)


titaiwangms (commented):
Nice skill doc! A few suggestions from our Pixtral/Ministral-3 VLM work:

**1. Missing `processor_config.json` in copy step**
Pixtral/Ministral-3 VLM exports use `processor_config.json` (not `image_processor.json`) for the vision preprocessing pipeline. Without copying this file, VLM inference will fail — the PixtralImageSizes and other preprocessing transforms won't be loaded. Suggest adding:
```bash
cp output/processor_config.json "${CACHE_DIR}/models/Custom/${MODEL_NAME}/" 2>/dev/null
```

**2. Missing single-component model copy**
Text-only LLMs export as `model.onnx` + `model.onnx.data` at the output root (not in a `decoder/` subdirectory). The copy commands only handle multi-component subdirs. Suggest adding:
```bash
cp output/model.onnx* "${CACHE_DIR}/models/Custom/${MODEL_NAME}/" 2>/dev/null
```

**3. `model_type` path is nested**
The workaround says to set `model_type: "decoder"` but the actual key path in genai_config.json is `model.type` (nested under the `model` object), not a top-level key.

**4. Mistral3 prompt template (nice-to-have)**
The skill includes ChatML, Llama, and Gemma templates but not Mistral3's `[INST]`/`[/INST]` format. Since Ministral-3 is now supported via mobius, this would be a helpful addition.

justinchuby:
@titaiwangms Done

titaiwangms (approved):

justinchuby:
Will improve build instruction if needed

PR #246, opened 2026-05-04
Add static KV cache (TensorScatter) support for Gemma4

justinchuby:
## Summary

Adds `--static-cache` support for Gemma4 models, enabling pre-allocated KV cache buffers with TensorScatter ops (opset 24).

### `src/mobius/models/gemma4.py`
- Added `StaticCacheState` dispatch to `Gemma4DecoderLayer.forward()` — extracts static cache from `past_key_value` and passes to attention
- Added `static_cache` parameter to `Gemma4TextAttention.forward()` — threads through to `_apply_attention()`
- Added `static_cache_mode` flag in `Gemma4TextModel.forward()` — skips GQA, fallback mask construction, and uses `None` attention bias when `attention_mask` is `None`

### `src/mobius/tasks/_gemma4.py`
- New `_make_gemma4_static_cache_inputs()` — creates per-layer static cache with correct dual head_dim (sliding=256, full=512) and skips KV-shared layers
- Updated `Gemma4TextCausalLMTask` with `static_cache`/`max_seq_len` params and static vs dynamic cache branching

### `src/mobius/tasks/_causal_lm.py`
- Added `Gemma4DecoderLayer` to `_validate_static_cache_support()`
- Scoped validation to skip encoder sub-modules (vision_encoder, audio_encoder, etc.)

### `src/mobius/__main__.py`
- Updated `--static-cache` CLI to resolve the correct task class for Gemma4 (uses text model type for multimodal models)

## Testing

- ✅ All 1186 existing tests pass
- ✅ `mobius build --model google/gemma-4-e2b-it --dtype f16 --static-cache --max-seq-len 4096` produces correct model with 30 TensorScatter + 35 Attention + 0 GQA ops
- ✅ Tiny config unit test verified programmatically

## Phase 1 Scope / Limitations

This is Phase 1 — full-attention and sliding-attention layers both work, but:
- **Sliding window masking**: Static cache path uses `is_causal=1` without sliding window constraint (standard causal attention). This matches the generic DecoderLayer static cache behavior.
- **KV-shared layers**: Correctly excluded from cache allocation. They continue to borrow K,V from source layers via the existing shared_kv_states mechanism.
- **Requires ORT ≥ 1.25.0** for TensorScatter opset 24 kernel support.

### DIFF
--- a/src/mobius/__main__.py
+++ b/src/mobius/__main__.py
@@ -268,9 +268,7 @@ def _cmd_build(args: argparse.Namespace) -> None:
             hf_config = transformers.AutoConfig.from_pretrained(
                 model_id_or_path, trust_remote_code=trust_remote_code
             )
-            task = _resolve_static_cache_task(
-                getattr(hf_config, "model_type", "")
-            )
+            task = _resolve_static_cache_task(getattr(hf_config, "model_type", ""))
 
         pkg = build(
             model_id_or_path,
--- a/src/mobius/__main__.py
+++ b/src/mobius/__main__.py
@@ -253,21 +253,25 @@ def _cmd_build(args: argparse.Namespace) -> None:
                 state_dict = model_module.preprocess_weights(state_dict)
             pkg.apply_weights(state_dict)
     else:
-        # Resolve static cache task if needed (requires model_type detection)
         model_id_or_path = args.model
+        extra_build_kwargs: dict = {}
         if static_cache_params is not None:
+            # Static cache: detect model type to resolve the correct task.
+            # For multimodal models, override to use the text-only model class.
             import transformers
 
             hf_config = transformers.AutoConfig.from_pretrained(
                 model_id_or_path, trust_remote_code=trust_remote_code
             )
             mt = getattr(hf_config, "model_type", "")
-            # For multimodal models (e.g. gemma4), use the text sub-config
-            # so build() resolves to the text-only model class for static cache.
             if hasattr(hf_config, "text_config"):
                 text_mt = getattr(hf_config.text_config, "model_type", "")
-                if text_mt:
+                if text_mt and text_mt != mt:
                     mt = text_mt
+                    # Override module_class so build() uses the text-only model
+                    from mobius._registry import registry as _registry
+
+                    extra_build_kwargs["module_class"] = _registry.get(mt)
             task = _resolve_static_cache_task(mt)
 
         pkg = build(
@@ -277,7 +281,8 @@ def _cmd_build(args: argparse.Namespace) -> None:
             load_weights=load_weights,
             trust_remote_code=trust_remote_code,
             execution_provider=execution_provider,
-            text_only=args.text_only or static_cache_params is not None,
+            text_only=args.text_only,
+            **extra_build_kwargs,
         )
 
     _save_package(pkg, output_dir, args, optimize, component_filter)
--- a/src/mobius/__main__.py
+++ b/src/mobius/__main__.py
@@ -121,7 +121,14 @@ def _cmd_build(args: argparse.Namespace) -> None:
 
     def _resolve_static_cache_task(model_type: str) -> ModelTask:
         """Create the correct static cache task for the given model type."""
-        if model_type in ("gemma4", "gemma4_text"):
+        if model_type == "gemma4":
+            from mobius.tasks._gemma4 import Gemma4Task
+
+            return Gemma4Task(
+                static_cache=True,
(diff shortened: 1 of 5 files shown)


titaiwangms:
## Code Review

### 🔴 Critical — Sliding-window masking silently disabled in static cache mode
**`src/mobius/components/_attention.py:156-158`** (and the static-cache branch of `_apply_attention`)

The static-cache path calls Attention with `is_causal=1` and no `local_window_size` / `attn_mask`. For Gemma4 sliding layers (e.g. `sliding_window=512`), once the sequence exceeds the window the model will attend to **all** past tokens instead of just the last 512 — diverging from HuggingFace and from the dynamic-cache path in this same model.

The PR description calls this a "Phase 1 limitation matching the generic DecoderLayer," but the dynamic-cache GQA path here *does* enforce sliding via `GQAContext.local_window_size` (`gemma4.py:1488-1495`), so the two cache modes will produce different outputs for long sequences. This reads as a latent correctness bug, not just a limitation.

**Suggested fix:** either (a) reject `--static-cache` in `_validate_static_cache_support` when any layer has `sliding_window` set, or (b) re-document as a known correctness bug with a runtime warning so users aren't silently surprised.

### 🟡 Major — Encoder-skip heuristic in `_validate_static_cache_support` is fragile
**`src/mobius/tasks/_causal_lm.py:377-385`**

Validation skips ModuleLists by string-prefix matching the module path against `vision_encoder` / `audio_encoder` / etc. A future module nested under a differently-named path would silently bypass validation. Prefer iterating `module.named_modules()` and explicitly whitelisting instances of `DecoderLayer | MoEDecoderLayer | Gemma4DecoderLayer` rather than blacklisting by path prefix.

### 🟢 Minor — `attention_mask: ir.Value | None` overloads `None`
**`src/mobius/models/gemma4.py:1429,1451-1455`**

`static_cache_mode = attention_mask is None` conflates "no mask provided" with "static-cache mode is on" and weakens the type contract of `Gemma4TextModel.forward`. Prefer an explicit `static_cache_mode: bool` parameter (or detect from the `past_key_values` type via `isinstance(..., StaticCacheState)`), keeping `attention_mask: ir.Value` non-optional.

### Notes
- Out-of-scope diff in `integrations/gguf/_config_mapping.py` and `rewrite_rules/_group_query_attention.py` aren't mentioned in the PR body — worth a one-line note for reviewers.
- Dual head_dim handling (256 sliding / 512 full), KV-shared layer exclusion, and per-layer cache index alignment look correct.
- Test coverage covers the happy path well; consider adding a test that exercises a sliding layer with `seq_len > sliding_window` to make the Critical issue above visible.

justinchuby:
Thanks for the thorough review @titaiwangms! All three findings addressed in commit f50b2b8:

### 🟡 Major — Fragile encoder-skip heuristic
Replaced the path-prefix blacklist (`vision_encoder`, `audio_encoder`, ...) with a whitelist `isinstance` check against `_supported = (DecoderLayer, MoEDecoderLayer, Gemma4DecoderLayer)`. Vision/audio encoder layers naturally skip validation since they use different classes (e.g. `Gemma4VisionEncoderLayer`). No more fragile string matching.

### 🟢 Minor — `attention_mask is None` overload
Now detects `static_cache_mode` by checking `isinstance(kv, StaticCacheState)` in `past_key_values`, keeping `attention_mask: ir.Value | None` with its natural semantics. The `GQAContext` import was updated to also import `StaticCacheState` for this check.

titaiwangms (approved):

justinchuby:
Nooo what's that

justinchuby:
**Update**: The hybrid per-layer approach (commit 0ebf291) fully addresses the Critical finding:

### 🔴 Sliding-window — RESOLVED
Instead of disabling sliding window or just warning, we now use **per-layer hybrid cache**:
- **Full-attention layers** → TensorScatter + standard Attention (static cache)
- **Sliding-attention layers** → GQA with `local_window_size` (dynamic cache)

This ensures sliding window constraints are correctly enforced via GQA's native `local_window_size` attribute, while full-attention layers benefit from static pre-allocated buffers. The two cache types coexist in the same model.

### 🟡 Fragile validation — RESOLVED  
Whitelist `isinstance` check (from f50b2b8) remains in place.

### 🟢 `None` overload — RESOLVED
`use_gqa` now depends on `attention_mask is not None` (which is always True in hybrid mode since sliding layers need it). Per-layer dispatch uses `isinstance(past_kv, StaticCacheState)` to route each layer to the correct path.

justinchuby:
Will wait after the standard paths land

PR #249, opened 2026-05-05
Add Gemma4 GenAI text generation example with benchmarking

justinchuby:
## Summary

Add `examples/gemma4_genai.py` — a streamlined example for Gemma4 text generation via ORT GenAI with performance benchmarking.

## Features

- **Build + export**: Uses `mobius build --runtime ort-genai` to create ONNX model with genai_config.json
- **Multi-step generation**: Prefill + autoregressive decode with streaming output
- **Performance metrics**: Tokens/sec, time-to-first-token (TTFT), decode throughput
- **Device selection**: `--device cpu` or `--device cuda`
- **Precision**: `--dtype f32/f16/bf16`
- **Model sizes**: `--model google/gemma-4-e2b-it` (default) or any Gemma4 variant
- **Pre-built models**: `--model-dir` to skip export, `--save-to` for export-only

## Usage

```bash
# CPU (default)
python examples/gemma4_genai.py

# CUDA with FP16
python examples/gemma4_genai.py --device cuda --dtype f16

# Different model
python examples/gemma4_genai.py --model google/gemma-4-e4b-it --max-new-tokens 200
```

## Relation to existing examples

This is a simplified, benchmark-focused version of `gemma4_ort_genai.py`. Key differences:
- Focused on text-only generation (no VLM/image support)
- Adds performance reporting (tok/s, TTFT)
- Uses mobius CLI for export instead of inline config generation
- Simpler code (~250 lines vs ~550 lines)

### DIFF
--- a/examples/gemma4_genai.py
+++ b/examples/gemma4_genai.py
@@ -24,6 +24,12 @@ Usage::
     # Custom prompt:
     python examples/gemma4_genai.py --prompt "Explain quantum computing"
 
+    # Image + text (multimodal):
+    python examples/gemma4_genai.py --image photo.jpg --prompt "Describe this image"
+
+    # Audio + text (multimodal):
+    python examples/gemma4_genai.py --audio speech.wav --prompt "Transcribe this audio"
+
     # CUDA generation:
     python examples/gemma4_genai.py --device cuda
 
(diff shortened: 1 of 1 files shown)


justinchuby:
@copilot please fix lint

titaiwangms (approved):

PR #250, opened 2026-05-05
Fix Gemma4 CUDA EP: GQA head_dim guard, provider_options, opset lowering

justinchuby:
## Summary
Fixes Gemma4 CUDA EP inference pipeline:

### 1. GQA head_dim guard (per-layer)
ORT's GQA CUDA kernel max head_dim is 256. Gemma4 full_attention layers (4, 9, 14) have head_dim=512. Added per-layer gate in `Gemma4TextModel.forward()` so only compatible sliding_attention layers use GQA.

**Model**: 12 GQA + 23 Attention (was 15 GQA + 20 Attention)

### 2. Remove CUDA provider_options from genai_config
Explicit CUDA provider_options in genai_config.json conflicted with GenAI's internal session setup, causing NaN. Empty `provider_options: []` lets GenAI handle CUDA configuration internally.

### 3. Float16 masks for fallback layers
Use float16 additive masks instead of bool masks for KV-shared and head_dim>256 layers. More robust on CUDA.

### 4. Opset 24→23 lowering
Implements the `ort_lower_opset_for_ep` flag (was declared but never wired up). Lowers opset from 24 to 23 for non-default EPs to avoid missing CUDA kernel registrations. Workaround until ORT adds opset 24 registrations (PR microsoft/onnxruntime#28365).

## Testing
All 1214 build graph tests pass. GenAI generation: 12.5 tok/s with 4 memcpy.

### DIFF
--- a/src/mobius/integrations/ort_genai/ep_config.py
+++ b/src/mobius/integrations/ort_genai/ep_config.py
@@ -45,7 +45,6 @@ def make_provider_options(
 
     Returns:
         A list with a single dict mapping the EP name to its options.
-        A list with a single dict mapping the EP name to its options.
         Empty list for CPU (no provider_options needed).
     """
     if ep == "cpu":
--- a/src/mobius/integrations/ort_genai/ep_config.py
+++ b/src/mobius/integrations/ort_genai/ep_config.py
@@ -45,7 +45,9 @@ def make_provider_options(
 
     Returns:
         A list with a single dict mapping the EP name to its options.
-        Empty list for CPU (no provider_options needed).
+        Empty list when no provider-specific options are needed (CPU,
+        or CUDA without explicit options — GenAI handles CUDA EP setup
+        internally).
     """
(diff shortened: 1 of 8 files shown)


src/mobius/integrations/ort_genai/ep_config.py
```diff
@@ -44,6 +44,7 @@ def make_provider_options(
         enable_webgpu_graph: Enable graph capture for WebGPU EP.
 
     Returns:
+        A list with a single dict mapping the EP name to its options.
```
justinchuby:
```suggestion
```

justinchuby:
@titaiwangms could you help with reviewing the logic that handles the attention bias? I wasn't sure it this is the right fix.

justinchuby:
Split into individual PRs: #253 (V norm FP16 fix), #254 (opset lowering), #255 (provider_options + float masks).

PR #252, opened 2026-05-05
Fix BF16 vision encoder: keep OneHot values as float32

justinchuby:
## Summary

ONNX OneHot op does not support bfloat16 for the values tensor. When building Gemma4 with `--dtype bf16`, the vision encoder fails to load because `CastLike` casts the OneHot values to bf16 before the op.

## Fix

Keep OneHot values as float32 and cast the output to model dtype afterward for the subsequent MatMul.

## Testing

- 13 gemma4 tests pass
- BF16 build succeeds: `mobius build --model google/gemma-4-e2b-it --dtype bf16 --no-weights`


titaiwangms (approved):
Did copilot miss op spec when it's designing model? Maybe we need to re-enforce into SKILL.md about op spec?

justinchuby:
> Did copilot miss op spec when it's designing model? Maybe we need to re-enforce into SKILL.md about op spec?

Onehot does support it. I think it is missing support in ort: https://onnx.ai/onnx/operators/onnx__OneHot.html

PR #253, opened 2026-05-05
Fix Gemma4 scale-free V norm FP16 overflow on CUDA

justinchuby:
Gemma4's parameterless V normalization (`v / sqrt(mean(v²) + ε)`) squared FP16 values directly. V projection outputs reach ~888, and 888² overflows FP16 max (65504) → inf → 0. This caused all-zero V outputs on CUDA (CPU uses FP32 internally).

**Fix**: Cast to FP32 before squaring, compute full RMSNorm in FP32, CastLike back.

**Result**: F16 CUDA 151.5 tok/s on H200 (was NaN).

### DIFF
--- a/src/mobius/models/gemma4.py
+++ b/src/mobius/models/gemma4.py
@@ -109,9 +109,7 @@ class _Gemma4ScaleFreeRMSNorm(nn.Module):
         self.dim = dim
         self.eps = eps
         # Constant all-ones scale (not a learnable parameter from HF).
-        self.weight = nn.Parameter(
-            [dim], data=ir.Tensor(np.ones(dim, dtype=np.float32))
-        )
+        self.weight = nn.Parameter([dim], data=ir.Tensor(np.ones(dim, dtype=np.float32)))
 
     def forward(self, op: builder.OpBuilder, hidden_states: ir.Value) -> ir.Value:
         # stash_type=1 means accumulate variance in float32, avoiding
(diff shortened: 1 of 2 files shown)


titaiwangms (approved):
Is there a way to avoid copilot doing this again?

titaiwangms:
It's concerning especially mobius would be heavily relying on ai if we don't have a good steering.

justinchuby:
I think we can have unit tests to test the overflow behavior to harden this. What do you think?

titaiwangms:
> I think we can have unit tests to test the overflow behavior to harden this. What do you think?

Sure. But also I think it would be good to have some ort knowledge in mobius as well. ORT is the major backend.

justinchuby:
> > I think we can have unit tests to test the overflow behavior to harden this. What do you think?
> 
> Sure. But also I think it would be good to have some ort knowledge in mobius as well. ORT is the major backend.

Maybe have it as a skill? Or we can use onnx-doctor to detect unsupported dtypes? What do you think?

titaiwangms:
> > > I think we can have unit tests to test the overflow behavior to harden this. What do you think?
> > 
> > 
> > Sure. But also I think it would be good to have some ort knowledge in mobius as well. ORT is the major backend.
> 
> Maybe have it as a skill? Or we can use onnx-doctor to detect unsupported dtypes? What do you think?

Maybe it's sufficient that we point it to https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/providers/cuda/cuda_execution_provider.cc for registered kernels (as well as contrib op and cpu ones). Just for copilot to understand what options it has when it's building models.

PR #254, opened 2026-05-05
Lower opset 24→23 for non-default EPs to eliminate memcpy

justinchuby:
ORT CUDA EP registers kernels up to opset 23. Opset 24 models have Reshape/Cast fall to CPU → 280 memcpy. Wire up the ort_lower_opset_for_ep flag to lower opset after optimization. Result: 282→4 memcpy. Workaround until ORT PR #28368 adds opset 24 registrations.

### DIFF
--- a/src/mobius/_builder.py
+++ b/src/mobius/_builder.py
@@ -37,6 +37,7 @@ from mobius._configs import (
     BaseModelConfig,
 )
 from mobius._execution_providers import ep_registry
+from mobius._flags import flags
 from mobius._model_package import ModelPackage
 from mobius._optimizations import optimize_model
 from mobius._registry import registry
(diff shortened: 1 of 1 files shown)


titaiwangms (commented):
It would be better when this trigger, it sends a message to users to remind us of adding opset to ort.

justinchuby:
Addressed the feedback — added an `INFO` log message when opset lowering triggers. It now logs the original opset, model name, EP, and links to the ORT tracking PR (https://github.com/microsoft/onnxruntime/pull/28368) so users know the workaround and when it can be removed.

Commit: 69fc529

src/mobius/_builder.py
```diff
@@ -208,6 +209,27 @@ def forward(self, op, input_ids, attention_mask,
...
+    # etc.). Without this, those ops fall to CPU and produce ~280 memcpy
+    # nodes that destroy performance. The flag defaults to True; set
+    # MOBIUS_ORT_LOWER_OPSET_FOR_EP=0 to disable for EPs that support
+    # opset 24 natively.
+    if flags.ort_lower_opset_for_ep and execution_provider != "default":
+        for name, model in pkg.items():
+            if "" in model.graph.opset_imports:
+                original = model.graph.opset_imports[""]
+                model.graph.opset_imports[""] = 23
+                logger.info(
```
justinchuby:
```suggestion
                logger.warning(
```

src/mobius/_builder.py
```diff
@@ -208,6 +209,27 @@ def forward(self, op, input_ids, attention_mask,
...
+                    "Lowered opset %d→23 for '%s' (EP=%s). "
+                    "ORT does not yet register opset %d kernels for this EP. "
+                    "Track https://github.com/microsoft/onnxruntime/pull/28368",
```
justinchuby:
```suggestion
                    "Track https://github.com/microsoft/onnxruntime/issues/27729",
```

PR #255, opened 2026-05-05
Fix CUDA EP provider_options and use float masks for fallback layers

justinchuby:
1. Remove CUDA provider_options from EP defaults — explicit options conflict with GenAI's internal session setup, causing NaN. 2. Use float16 additive masks for all fallback Attention layers instead of bool masks (more robust on CUDA). 3. Return empty provider_options list when no options needed.

### DIFF
--- a/src/mobius/_execution_providers.py
+++ b/src/mobius/_execution_providers.py
@@ -207,10 +207,12 @@ def _register_builtins() -> None:
                 {ir.DataType.FLOAT, ir.DataType.FLOAT16, ir.DataType.BFLOAT16}
             ),
             supports_packed_multi_head_attention=True,
-            provider_options={
-                "enable_cuda_graph": "0",
-                "enable_skip_layer_norm_strict_mode": "1",
-            },
+            # provider_options intentionally empty for CUDA EP.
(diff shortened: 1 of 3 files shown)


titaiwangms:
This PR is superseded by #262, which goes further by omitting masks entirely for full-attention layers (enabling MEA for head_dim=512) and switching sliding-window to float additive bias. The provider_options fix in this PR is also included in #262.

Recommend closing in favor of #262.

justinchuby:
Closing: this PR is superseded by other changes:

1. **CUDA provider_options removal** (changes 1 & 2): This was found to be incorrect — GenAI derives its providers list from `provider_options` names (config.cpp:1763). Empty `provider_options` means CPU-only. The correct fix is `[{"cuda": {}}]`, which is already on main via PR #247.

2. **Float masks for fallback attention** (change 3): Superseded by PR #262 which goes further — omits the mask entirely for full_attention layers to enable MEA (CUTLASS FMHA) for head_dim=512.

Relevant PRs:
- #247 (merged): Provider name format fix
- #253 (merged): V norm FP16 overflow fix  
- #262 (open): Omit full_attention mask for MEA

PR #258, opened 2026-05-05
Clarify Gemma4 GQA dispatch comments for KV-shared layers

justinchuby:
Update comments in `Gemma4TextModel.forward()` to clarify that all non-shared layers use GQA when available. Only KV-shared layers fall back to standard Attention.


rui-ren (approved):

justinchuby:
Closing: trivial comment-only change that can be folded into #255 or #262 when they merge. Not worth a separate PR.

PR #260, opened 2026-05-05
Use Shape instead of ReduceSum for seqlens_k computation

justinchuby:
Replace `ReduceSum(attention_mask, axis=1) - 1` with `Shape(attention_mask)[1] - 1` for computing `seqlens_k` in GQA.

`ReduceSum` on an empty dimension triggers an ORT CUDA assertion. `Shape` is simpler, avoids the empty-tensor issue, and is equivalent for non-padded sequences (GenAI uses dense attention masks).

Applied to `base.py`, `gemma4.py`, and both GQA rewrite rules.

### DIFF
--- a/.agents/skills/building-ort-genai/SKILL.md
+++ b/.agents/skills/building-ort-genai/SKILL.md
@@ -1,0 +1,255 @@
+---
+name: building-ort-genai
+description: >
+  Use this skill when building OnnxRuntime and onnxruntime-genai from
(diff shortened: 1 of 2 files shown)


justinchuby:
Closed because Shape(mask)[1]-1 is incorrect for padded batches (batch>1). GenAI sets attention_mask=0 for pad tokens, so ReduceSum correctly counts actual tokens per batch while Shape gives the padded length. ReduceSum is the correct approach.

PR #261, opened 2026-05-05
Omit mask for full_attention layers to enable MEA for head_dim=512

justinchuby:
Omit explicit mask for full_attention KV-shared layers (is_causal=1 handles causality). This enables ORT's MEA (CUTLASS FMHA) for head_dim=512 layers without bias alignment issues. Also uses float16 masks for sliding layers and includes V norm FP32 fix.

Result: **165.0 tok/s** (up from 159.3) on H200.


justinchuby:
Replaced with clean PR — had unrelated changes mixed in.

PR #262, opened 2026-05-05
Omit mask for full_attention layers to enable MEA for head_dim=512

justinchuby:
Omit explicit mask for full_attention KV-shared layers (is_causal=1 handles causality). Enables MEA for head_dim=512 layers. Also switches sliding-window fallback to float additive bias. Result: 165 tok/s (up from 159). Single file change.


titaiwangms (approved):

justinchuby:
Recreating as clean PR with squashed commit.

PR #265, opened 2026-05-05
Cast vision/audio inputs f32 to model dtype for GenAI compatibility

justinchuby:
## Summary

Keep vision/audio encoder graph inputs as float32 (matching GenAI image/audio processor output) and add a Cast(f32→model_dtype) at the graph entry. Encoder weights use the requested f16/bf16 dtype for memory efficiency.

## Problem

GenAI multimodal processor outputs float32 pixel values and audio features. When building with `--dtype f16` or `--dtype bf16`, the vision/audio encoder graph inputs were also cast to f16/bf16, creating a type mismatch: GenAI sends f32, encoder expects f16, ORT produces all-zero output (silent failure).

## Solution: Cast at graph entry

Instead of keeping entire vision/audio encoders at f32 (wasteful), add a `Cast` op at the start of each encoder graph:

```
pixel_values (f32, from GenAI) → Cast(f32→f16) → vision encoder (f16 weights) → image_features
```

This gives both **GenAI compatibility** (f32 input) and **memory efficiency** (f16/bf16 weights).

## Changes

- `_gemma4.py`: Vision/audio encoder inputs always `ir.DataType.FLOAT`, with `op.Cast` to `config.dtype` when dtype ≠ f32
- `_builder.py`: No changes needed — all parameters cast to requested dtype as before

## Testing

13 gemma4 tests pass. Verified f16 build: vision input=f32, first node=Cast, weights=f16.

### DIFF
--- a/src/mobius/_builder.py
+++ b/src/mobius/_builder.py
@@ -191,6 +191,9 @@ def build_from_module(
     if hasattr(config, "validate"):
         config.validate()
     dtype = getattr(config, "dtype", ir.DataType.FLOAT)
+    # Cast all parameters to the target dtype. Vision/audio encoder weights
+    # are included — their graph inputs are kept at f32 (matching GenAI's
+    # image processor output) with a Cast at the graph entry.
     _cast_module_dtype(module, dtype)
     resolved_task = get_task(task)
     capabilities = ep_registry.require(execution_provider)
--- a/src/mobius/tasks/_fun_asr_speech_language.py
+++ b/src/mobius/tasks/_fun_asr_speech_language.py
@@ -91,14 +91,19 @@ class FunASRSpeechLanguageTask(ModelTask):
         input_dim = (config.audio.input_size if config.audio else None) or 560
 
         graph, builder = _make_graph(name="audio_encoder")
+        op = builder.op
 
+        # Audio encoder input is always f32 (matching audio processor output).
+        # Cast at graph entry for f16/bf16 builds.
         input_features = builder.input(
             "input_features",
-            dtype=config.dtype,
+            dtype=ir.DataType.FLOAT,
             shape=[batch, seq_len, input_dim],
         )
+        if config.dtype and config.dtype != ir.DataType.FLOAT:
+            input_features = op.Cast(input_features, to=config.dtype)
 
-        audio_features = audio_encoder(builder.op, input_features)
+        audio_features = audio_encoder(op, input_features)
 
         builder.add_output(audio_features, "audio_features")
         return _make_model(graph)
--- a/src/mobius/tasks/_gemma4.py
+++ b/src/mobius/tasks/_gemma4.py
@@ -298,11 +298,16 @@ class Gemma4Task(ModelTask):
         graph, builder = _make_graph(name="vision_encoder")
         op = builder.op
 
+        # Vision encoder input is always f32 (matching GenAI's image processor
+        # output). When the model uses f16/bf16, add a Cast at the graph entry
+        # so weights can stay at the requested dtype for memory efficiency.
         pixel_values = builder.input(
             "pixel_values",
-            dtype=config.dtype,
+            dtype=ir.DataType.FLOAT,
             shape=[batch, num_patches, pixel_dim],
         )
+        if config.dtype and config.dtype != ir.DataType.FLOAT:
+            pixel_values = op.Cast(pixel_values, to=config.dtype)
         pixel_position_ids = builder.input(
             "pixel_position_ids",
             dtype=ir.DataType.INT64,
@@ -352,9 +357,11 @@ class Gemma4Task(ModelTask):
 
         input_features = builder.input(
             "input_features",
-            dtype=config.dtype,
+            dtype=ir.DataType.FLOAT,  # Always f32 (matching audio processor output)
             shape=[batch, time, input_size],
         )
+        if config.dtype and config.dtype != ir.DataType.FLOAT:
+            input_features = op.Cast(input_features, to=config.dtype)
         input_features_mask = builder.input(
             "input_features_mask",
             dtype=ir.DataType.BOOL,
--- a/src/mobius/tasks/_phi4mm_multimodal.py
+++ b/src/mobius/tasks/_phi4mm_multimodal.py
@@ -85,19 +85,24 @@ class Phi4MMMultiModalTask(ModelTask):
         image_size = (config.vision.image_size if config.vision else None) or 448
 
         graph, builder = _make_graph(name="vision_encoder")
+        op = builder.op
 
+        # Vision encoder input is always f32 (matching GenAI's image processor
+        # output). Cast at graph entry for f16/bf16 builds.
         pixel_values = builder.input(
             "pixel_values",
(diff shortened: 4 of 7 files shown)


titaiwangms (commented):
## Review summary

Reviewed by Opus 4.7 + GPT-5.3-Codex (code) + GPT-5.5 (adversarial) + Claude Sonnet 4.6 (readability). Consensus: **the direction is right, but the fix is incomplete and may not actually produce a loadable model in BF16/F16 multimodal builds.**

The PR keeps selected encoder *parameters* at float32, but does not change the encoder graph's *I/O* dtype or the downstream *embedding* model's input dtype. That's three independent problems.

---

### 🔴 Major — Encoder graph I/O still uses `config.dtype`, mixing types with f32 weights

The tasks declare encoder graph inputs with `dtype=config.dtype` (e.g. `_vision_language_3model.py:86-91`, `_phi4mm_multimodal.py:122-125`, `_gemma4.py:301-305`). After this PR, in a bf16 build:

- `pixel_values` / `input_features` graph input → **bf16**
- First Conv/MatMul/Gemm weight in the encoder → **float32**

ONNX numeric op type constraints generally require matching tensor types. The likely failure mode is an invalid graph or ORT load-time type-check failure — not a safe fallback to f32 compute.

Suggested direction: make precision a per-component contract, not a parameter-only exclusion. Encoder graph inputs, constants, and outputs should all be FLOAT, with an explicit Cast at the next-component boundary.

### 🔴 Major — Embedding model still expects bf16/f16 multimodal features

Even if the encoder is fully corrected to float32, ORT GenAI passes its outputs as `image_features` / `audio_features` into a separate embedding session whose input is declared at `config.dtype` (`_phi4mm_multimodal.py:166-174`, `_gemma4.py:394-407`). `genai_config.json` (`integrations/ort_genai/genai_config.py:279-296`) only wires names; it does not insert dtype conversions.

Suggested fix: either (a) declare embedding `*_features` inputs as FLOAT and Cast to decoder dtype inside the embedding graph before token-merge, or (b) keep embedding in float32 and Cast `inputs_embeds` at the decoder boundary.

### 🔴 Major — Hardcoded `("vision_encoder", "audio_encoder")` misses real models

The exclude list only matches literal `nn.Module` attribute names. Several existing multimodal/speech models bypass the fix:

| Model | Module attr | Excluded? |
|---|---|---|
| Gemma4 / Qwen-VL / Gemma3 / MLLaMA / LLaVA / DeepSeek-OCR2 / InternVL / BLIP2 | `vision_encoder`, `audio_encoder` | ✅ |
| **Phi4-MM** | `speech_encoder` (`models/phi.py:981`, remapped via `tasks/_phi4mm_multimodal.py:59`) | ❌ |
| **Qwen3-ASR** | `audio_tower` (`models/qwen3_asr.py:348`, remapped via `tasks/_speech_language.py:51`) | ❌ |
| **FunASR** | `audio_tower` (`models/fun_asr.py:511`, remapped via `tasks/_fun_asr_speech_language.py:54`) | ❌ |

Suggested fix: derive exclusions from the resolved task's component-attribute mapping (or have each task declare a `component_dtypes` / encoder-attrs policy) instead of guessing module layout.

### 🟡 Minor — No tests covering the new behavior

Existing `_cast_module_dtype` tests (`_exporter_test.py:181`) only exercise full-module casting. Please add tests asserting (per affected model class) that encoder initializer dtypes remain FLOAT while decoder initializers are BF16/F16. Phi4-MM, Qwen3-ASR, and FunASR are the highest-value cases since they would have silently regressed.

### 🟡 Minor — Silent precision-policy override

Passing `--dtype bf16` previously implied uniform bf16. After this PR it silently means "mixed precision". Worth at least a `logger.info(...)` listing which sub-modules were kept at float32, and ideally a named policy flag (`keep_multimodal_encoders_fp32=True`) exposed in the public API + docs.

### 🟢 Minor / Nit — Readability

- Hardcoded tuple `("vision_encoder", "audio_encoder")` duplicates information already in `_MODEL_ROLE_MAP` (lines 111–115). Either co-locate as `_ENCODER_SUBMODULE_ATTRS = frozenset(...)` or derive from the role map.
- `exclude=exclude_from_cast or None` is a no-op: `_cast_module_dtype` already guards with `if exclude:` (truthy check passes both `None` and `set()`). Drop the `or None`, or change the signature to `exclude: set[str] = frozenset()`.
- Parameter name `exclude` is weaker than the call-site `exclude_from_cast`. Consider `excluded_attrs`.
- Docstring on `exclude` mixes contract + specific use case. Keep the rationale at the call site.
- `exclude` silently ignores typos in attr names — consider validating or warning.

### ✅ What's right
- Casting before graph construction (vs post-export mutation) is the right architectural layer.
- `id(param)` set is the correct way to identify shared parameters under weight tying.
- `keyword-only` (`*`) marker on the new `exclude` arg is good API hygiene.
- Inline comment at the call site cites the concrete ORT op gap (OneHot/Mul/ReduceMean) — exactly the "why" the project style guide asks for.
- Confirmed (`_optimizations.py:285-291`): GQA / QKV-packing fusions are gated to `model_role == "decoder"`, so passing bf16 to `optimize_model()` for the f32 vision graph does not trigger decoder-only fusions on it.

---

### Recommendation
**Request changes.** The two BF16 graph-boundary issues (encoder I/O dtype, embedding-input dtype) are likely to make exported BF16 multimodal models fail to load in ORT, which would silently invalidate the PR's "BF16 export succeeds" claim once the embedding/vision sessions are actually run end-to-end. Worth verifying with an ORT GenAI smoke test on Gemma4-bf16 (load all three sub-models in one session pipeline) before merging.

justinchuby:
Addressed the review feedback in commit 76e37fa:

### What was fixed

Extended the f32 encoder input pattern (FLOAT input + Cast at graph entry) to **all** multimodal task files:

| Task | Encoder inputs fixed |
|---|---|
| VisionLanguageTask (base) | `pixel_values` |
| QwenVLTask | `pixel_values` |
| PixtralVLTask | `pixel_values` |
| Phi4MMMultiModalTask | `pixel_values` + `audio_embeds` |
| SpeechLanguageTask | `input_features` |
| FunASRSpeechLanguageTask | `input_features` |
| Gemma4Task | (already fixed in original commit) |

### Embedding boundary dtype

Embedding model `image_features`/`audio_features` inputs remain at `config.dtype`. This is correct because:
1. Encoder weights are cast to `config.dtype` → encoder computation runs in `config.dtype` → encoder outputs are `config.dtype`
2. ORT GenAI passes encoder output directly to embedding input — dtypes match
3. No dtype conversion needed at the boundary

### Tests added

4 new tests verify encoder inputs stay FLOAT in f16/bf16 builds:
- `test_multimodal_encoder_inputs_are_float32[f16/bf16]` — generic VL (LLaVA 3-model split)
- `test_gemma4_encoder_inputs_are_float32[f16/bf16]` — Gemma4 (vision + audio encoders)

### Test results
- 1204 passed, 40 skipped, 0 failed
- All lint checks pass

justinchuby:
@titaiwangms PTAL

justinchuby:
Merged LMK if follow ups needed

PR #267, opened 2026-05-06
Add use_gqa_for_kv_shared flag for hybrid GQA/Attention CUDA EP dispatch

justinchuby:
## Summary

Adds a feature flag `MOBIUS_USE_GQA_FOR_KV_SHARED` (default `False`) to control whether KV-shared layers use GQA or standard ONNX Attention on CUDA EP.

### Background

Gemma4 has 35 decoder layers: 15 non-shared (layers 0-14) with their own KV cache, and 20 KV-shared (layers 15-34) that borrow K/V from source layers. On CUDA EP:

- **Non-shared layers**: Work correctly with GQA (fused RoPE + attention + KV cache)
- **KV-shared layers**: Cannot use GQA because they pass the full KV cache as K input with no `past_key` — the CUTLASS MEA aligned kernel crashes at sequence lengths that are multiples of 8 (tracked in microsoft/onnxruntime#28376)

### Changes

1. **`_flags.py`**: New `use_gqa_for_kv_shared` flag with `MOBIUS_USE_GQA_FOR_KV_SHARED` env var
2. **`gemma4.py`**:
   - Per-layer dispatch: GQA for normal layers, Attention for KV-shared (controlled by flag)
   - Always use float additive masks for fallback path (fixes CUDA EP output — bool masks caused incorrect results)
   - Remove unused `create_padding_mask`/`create_sliding_window_mask` imports

### Testing

- All 13 Gemma4 L1 tests pass
- Lint clean
- Pure ORT decoder test: all seq lengths 1-64 pass on CUDA EP
- GenAI end-to-end blocked by separate issue (Equal<int64> CPU kernel on CUDA input_ids, tracked in onnxruntime-genai PR #2123)

### Future

Set `MOBIUS_USE_GQA_FOR_KV_SHARED=1` once ORT GQA supports `new_kv_length=0` for KV-shared layers.

### DIFF
--- a/src/mobius/_flags.py
+++ b/src/mobius/_flags.py
@@ -84,6 +84,11 @@ class _Flags:
          - ``True``
          - Lower the ONNX opset declaration to 23 for non-CPU EPs
            (ORT ≤1.24.x workaround).
+       * - ``use_gqa_for_kv_shared``
+         - ``MOBIUS_USE_GQA_FOR_KV_SHARED``
+         - ``False``
+         - Use GQA for KV-shared layers on CUDA EP. Requires ORT GQA
+           new_kv_length=0 support.
     """
 
     suppress_dedup_warning: bool = dataclasses.field(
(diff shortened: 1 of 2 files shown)


justinchuby:
Consolidated into PR #262 which now includes the feature flag, hybrid dispatch, and float additive bias changes.

PR #268, opened 2026-05-06
[KV cache] support-past-present-share-buffer

rui-ren:
## Support past present share buffer

Add `supports_past_present_share_buffer=True`

Benefit: 

Save the `memcpy` overhead.


justinchuby (approved):

justinchuby:
Thanks!

PR #272, opened 2026-05-06
Add bool mask vs float additive bias guidance to debugging-memcpy skill

justinchuby:
Add a section to the debugging-memcpy skill documenting when to use float additive bias vs bool mask for attention.

## Key points

- **Float additive bias** recommended for complex patterns (sliding window, KV-shared layers, dual head_dim, padding)
- **Bool mask** only for simple causal-only patterns where `is_causal=1` suffices
- **Common misconception:** Bool masks do NOT enable Flash Attention — Flash requires `attn_mask=nullptr`
- Includes decision table and `create_attention_bias()` recommendation

Learned from Gemma4 debugging where bool mask construction for mixed sliding-window + KV-shared + dual head_dim was error-prone.


titaiwangms (commented):
Test from CLI

titaiwangms:
LGTM with three small suggestions. Verified the technical claims against `core/providers/cuda/llm/attention.cc` (`ConvertAttnMaskToBias`, Flash mask rejection at line 1354, MEA bias-alignment guard at 1402). Tried to leave inline suggestion blocks but the GitHub Reviews API keeps 422'ing — pasting them here.

---

**1. Scope the section to the ONNX Attention op (suggested replacement for the section header on line 332):**

````markdown
## Bool mask vs float additive bias for attention

> Applies to the ONNX-domain `Attention` op (opset 23/24) that mobius emits.
> Contrib `MultiHeadAttention` takes a pre-built float `attention_bias`;
> contrib `GroupQueryAttention` does not accept any bias.
````

The bool-mask + `ConvertAttnMaskToBias` claims are ONNX-Attention-specific. Mobius emits ONNX `Attention` so the audience is right, but stating it avoids confusion when this skill is read during contrib-op debugging.

---

**2. Wording nit on the "double-applies" bullet (lines 358–359):**

````markdown
- **`is_causal=1` + bool mask** is redundant and easy to get out of sync —
  the `is_causal` flag adds its own causal mask on top of the explicit one,
  so any off-by-one between them silently changes results
````

In practice MEA (`attention.cc:799`) and Unfused (`:1186`) apply both `p.causal` *and* the converted bias, which is idempotent if the bool mask is itself causal — output stays correct, just redundant. "Double-applies constraints" reads as if correctness breaks. The real footgun is an off-by-one between the two, or a non-causal bool pattern silently overridden by `is_causal=1`.

---

**3. Flash prerequisites + `nonpad_kv_seqlen` (suggested replacement for lines 374–377):**

````markdown
Bool masks do **NOT** enable Flash Attention. Flash Attention requires
`attn_mask=nullptr` (no mask at all). Both bool and float masks route
to Memory-Efficient Attention (MEA) or unfused attention. If you need
Flash Attention, use `is_causal=1` with no explicit mask. Note that
Flash also requires fp16/bf16, `head_size == v_head_size` (≤ 256),
and SM ≥ 8.0 (Ampere+); otherwise the op falls to MEA or unfused
regardless of how the mask is expressed. For variable-length padding,
prefer `nonpad_kv_seqlen` (opset 24) over a mask — it preserves Flash
eligibility, while a mask hard-disables Flash.
````

`is_causal=1` is necessary but not sufficient for Flash. And it's worth distinguishing **padding** (express via `nonpad_kv_seqlen` → Flash-eligible) from **attention pattern** (sliding window / KV-shared / dual head_dim → float bias → MEA/unfused). Without that split, "use float bias" might lead users to express padding via bias too, and silently lose Flash on causal LM workloads.

titaiwangms (approved):

PR #274, opened 2026-05-06
Fix Gemma4 GenAI regression: disable past_present_share_buffer for dual head_dim

justinchuby:
## Summary

Fix CPU regression where Gemma4 GenAI generation fails with:
```
inconsistent total_sequence_length (between attn_mask and past_key and past_value)
```

## Root Cause

`past_present_share_buffer` was set to `true` in genai_config.json because the `default` EP maps to `cpu` (which has `supports_past_present_share_buffer=True`). However, Gemma4 has dual head_dim (256 for sliding-window, 512 for full-attention), and shared buffers require uniform head_dim across all KV cache layers.

## Fix

Force `past_present_share_buffer=false` in genai_config.json for all Gemma4 model types.

## Testing

- 15 gemma4 tests pass
- 119 ort_genai tests pass
- Verified: Gemma4 CPU generation produces correct output with the fix

### DIFF
--- a/src/mobius/integrations/ort_genai/auto_export.py
+++ b/src/mobius/integrations/ort_genai/auto_export.py
@@ -655,6 +655,14 @@ def _write_genai_config(
             **audio_kwargs,
         )
 
+    # Disable past_present_share_buffer for models with dual head_dim
+    # (e.g., different head_dim for sliding vs full-attention layers).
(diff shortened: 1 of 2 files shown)


rui-ren (approved):

rui-ren:
👍

PR #275, opened 2026-05-06
Keep vision/audio inputs f32 in single-model tasks for f16/bf16 builds

justinchuby:
## Summary

Apply the f32+Cast-at-input pattern to 3 remaining task files that still used `config.dtype` for vision/audio graph inputs.

## Problem

When building with `--dtype bf16` or `--dtype f16`, image/audio processor outputs are always f32. If graph inputs are declared at `config.dtype` (e.g. bf16), ORT raises type mismatches at load time.

## Fix

Declare vision/audio graph inputs as `ir.DataType.FLOAT` and add a `Cast` to `config.dtype` at graph entry, matching the pattern already used in `_gemma4.py` and `_vision_language_3model.py`.

### Files changed:
- `_vision_language.py`: `pixel_values` (Qwen3-VL single-model task)
- `_multimodal.py`: `pixel_values` + `audio_features` (generic multimodal task)
- `_audio_ctc.py`: `input_features` (SenseVoiceSmall CTC task)

## Tests

1204 passed, 40 skipped — no regressions.

### DIFF
--- a/src/mobius/tasks/_audio_ctc.py
+++ b/src/mobius/tasks/_audio_ctc.py
@@ -22,7 +22,12 @@ import onnx_ir as ir
 
 from mobius._configs import ArchitectureConfig
 from mobius._model_package import ModelPackage
-from mobius.tasks._base import ModelTask, _make_graph, _make_model
+from mobius.tasks._base import (
+    ModelTask,
+    _cast_encoder_input,
+    _make_graph,
+    _make_model,
+)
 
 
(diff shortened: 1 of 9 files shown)


justinchuby:
Added DRY refactor: extracted `_cast_encoder_input(op, tensor, config)` helper to `_base.py` and replaced all 13 inline Cast blocks across 8 task files. Net -13 lines (39 added, 52 removed).

Commit: 26f1c0c

PR #276, opened 2026-05-06
Add benchmark methodology to profiling skill

justinchuby:
Add a comprehensive benchmark methodology section to the profiling skill.

## Benchmark dimensions covered

- **5 pipelines:** ORT CUDA, GenAI CUDA/CPU, HF PyTorch CUDA/CPU
- **5 dtypes:** F16, BF16, F32, Q4_K_M, NF4
- **3 EP variants:** default (MEA), cuda (GQA rewrite), onnx-standard
- **4 input modalities:** text, image+text, audio+text, multimodal
- **4 core metrics:** decode tok/s, TTFT, VRAM, model size
- **5 additional dimensions:** cache type, prompt length, batch size, padding strategy, KV buffer
- **6 best practices** for reproducible benchmarks


justinchuby:
Lowering opset is not the right thing to do. We should update onnxruntime kernel registration

PR #279, opened 2026-05-06
Enable GQA for KV-shared layers and fix dtype issues in vision and audio

apsonawane:
### Summary

Four changes to enable optimized Gemma4 multimodal inference:

1. **GQA for KV-shared layers** — use `GroupQueryAttention` instead of standard `Attention` for the 20 KV-shared decoder layers
2. **Vision encoder float32 input** — accept float32 `pixel_values` and cast inside the model to prevent NaN
3. **Audio encoder SkipNorm fix** — prevent incorrect fusion of `Add(MatMul, 1D_bias) + RMSNorm`
4. **Enable `past_present_share_buffer`** — avoid per-step KV cache allocation

---

### 1. GQA for KV-shared layers (`gemma4.py`)

KV-shared layers borrow K/V from a source layer's `present_key`/`present_value` (already in BNSH format). Instead of Transpose+Reshape into BSNH for the standard `Attention` op, we now pass:

- Empty K/V tensors `[B, 0, kv_hidden]` (`kv_sequence_length=0`)
- Source layer's present K/V directly as `past_key`/`past_value`

This eliminates Transpose+Reshape ops from the graph for all 20 shared layers and uses the optimized GQA flash attention kernel. Requires corresponding ORT changes to support `kv_sequence_length=0` (separate ORT PR).

**Decoder loop change:** All layers (shared and non-shared) now receive GQA context when the EP supports it. The previous `is_shared` exclusion and fallback attention bias/position embedding generation for shared layers are removed.

### 2. Vision encoder float32 input (`_gemma4.py`)

**Problem:** The image processor in ORT-GenAI outputs float32 `pixel_values`, but the vision encoder declared its input as `config.dtype` (float16). The ORT-GenAI trimming path did a raw `memcpy` from float32 source into an fp16 tensor using fp16 strides — reinterpreting float32 bit patterns as fp16, producing NaN.

**Fix:** Declare `pixel_values` as `ir.DataType.FLOAT` and add `op.Cast(pixel_values, to=config.dtype)` inside the ONNX model when needed. This keeps the host-side pipeline in float32 and lets the model handle the precision conversion safely.

### 3. Audio encoder SkipNorm fix (`_skip_norm.py`)

**Problem:** The `AddRMSNormToSkipNorm` rewrite rule fused `Add(MatMul_result, 1D_bias) + RMSNorm` into `SkipSimplifiedLayerNormalization`. The 1D bias `[hidden]` became the `skip` input, but the op requires both inputs to have the same shape (`[B, S, hidden]`). This caused:
- CPU: `skip is expected to have 3 or 2 dimensions, got 1`
- CUDA: `input and skip shall have same shape when enable_skip_layer_norm_strict_mode is True`

**Fix:** Add a rank equality check in the `check()` method. Fusion is skipped when the two Add inputs have different ranks (e.g., 3D tensor + 1D bias via broadcasting).

### 4. Enable `past_present_share_buffer` (`genai_config.py`)

Set `past_present_share_buffer` to `true` by default for all EPs. This pre-allocates KV cache buffers and aliases past/present pointers, avoiding per-step allocation and past→present copies.

---

### Testing

| Scenario | Status |
|----------|--------|
| Text-only generation (CPU) | ✅ |
| Text-only generation (CUDA) | ✅ |
| Multimodal image+text (CPU) | ✅ Clean output, no repetition |
| Multimodal image+text (CUDA) | ✅ Clean output, no crashes |
| Model graph | ✅ 35 GQA nodes (15 source + 20 shared), 0 Attention nodes |

### DIFF
--- a/src/mobius/models/gemma4.py
+++ b/src/mobius/models/gemma4.py
@@ -727,37 +727,79 @@ class Gemma4TextAttention(nn.Module):
             )
 
         if self.is_kv_shared_layer:
-            # KV-shared layers always use standard Attention path because
-            # they borrow K,V from a source layer (no own KV cache).
-            if use_gqa:
-                raise ValueError("KV-shared layers should not receive GQAContext")
-            # Borrow full-history K,V from source layer.
-            # present_key/value from the ONNX Attention op is 4D:
-            #   [batch, kv_heads, total_seq, head_dim]
-            # The Attention op expects key/value as 3D:
-            #   [batch, total_seq, kv_heads * head_dim]
-            # Transpose and reshape to match.
+            # KV-shared layers borrow K,V from a source layer (no own KV cache).
             src_key, src_value = shared_kv_states[self.kv_shared_layer_index]
 
-            # [B, kv_heads, total_seq, head_dim] → [B, total_seq, kv_heads*head_dim]
-            src_key = op.Transpose(src_key, perm=[0, 2, 1, 3])
-            src_key = op.Reshape(src_key, [0, 0, -1])
-            src_value = op.Transpose(src_value, perm=[0, 2, 1, 3])
-            src_value = op.Reshape(src_value, [0, 0, -1])
+            if use_gqa:
+                # GQA path for shared KV: pass empty K/V tensors and wire the
+                # source layer's present_key/value as past_key/past_value.
+                # The shared buffer is already in BNSH format, so GQA reads it
+                # directly — no Transpose/Reshape needed.
+                gqa_ctx = attention_bias
 
-            attn_output, present_key, present_value = _apply_attention(
-                op,
-                query_states,
-                src_key,
-                src_value,
-                attention_bias,
-                past_key=None,
-                past_value=None,
-                num_attention_heads=self.num_attention_heads,
-                num_key_value_heads=self.num_key_value_heads,
-                scale=self.scaling,
-                softcap=self.softcap,
-            )
+                # Create empty K/V tensors with kv_sequence_length=0.
+                # Shape: [batch, 0, kv_heads * head_dim]
+                batch_dim = op.Shape(query_states, start=0, end=1)
+                kv_hidden = self.num_key_value_heads * self.head_dim
+                empty_shape = op.Concat(
+                    batch_dim,
+                    op.Constant(value_ints=[0, kv_hidden]),
+                    axis=0,
+                )
+                empty_kv = op.CastLike(op.ConstantOfShape(empty_shape), query_states)
+
+                gqa_attrs: dict = {
+                    "num_heads": self.num_attention_heads,
+                    "kv_num_heads": self.num_key_value_heads,
+                    "scale": self.scaling,
+                    "do_rotary": 1,
+                    "rotary_interleaved": int(self._rope_interleave),
+                }
+                if self.softcap:
+                    gqa_attrs["softcap"] = self.softcap
+                if self.rotary_embedding_dim:
+                    gqa_attrs["rotary_embedding_dim"] = self.rotary_embedding_dim
+                if gqa_ctx.local_window_size > 0:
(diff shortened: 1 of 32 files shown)


src/mobius/models/gemma4.py
```diff
@@ -727,37 +727,79 @@ def forward(
...
+                # Shape: [batch, 0, kv_heads * head_dim]
+                batch_dim = op.Shape(query_states, start=0, end=1)
+                kv_hidden = self.num_key_value_heads * self.head_dim
+                empty_shape = op.Concat(
+                    batch_dim,
+                    op.Constant(value_ints=[0, kv_hidden]),
+                    axis=0,
+                )
+                empty_kv = op.CastLike(op.ConstantOfShape(empty_shape), query_states)
+
```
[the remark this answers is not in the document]
apsonawane: No additional changes needed, it is already safe

src/mobius/models/gemma4.py
```diff
@@ -35,7 +35,6 @@
 
 from mobius._build_context import ep_capabilities
 from mobius._configs import ArchitectureConfig, Gemma4Config
-from mobius._flags import flags
```
justinchuby, line 38: Could you also remove the flag definition?

src/mobius/tasks/_gemma4.py
```diff
@@ -311,6 +311,10 @@ def _build_vision(
             shape=[batch, num_patches, 2],
         )
 
+        # Cast float32 pixel_values from image processor to model dtype (e.g. fp16)
+        if config.dtype != ir.DataType.FLOAT:
+            pixel_values = op.Cast(pixel_values, to=config.dtype)
```
justinchuby: Could you update this to the shared utility: https://github.com/onnxruntime/mobius/blob/8542e9f855f9120b19537525f985200b58464971/src/mobius/tasks/_base.py#L133

src/mobius/models/gemma4.py
```diff
@@ -1477,11 +1514,10 @@ def forward(
...
         if use_gqa:
-            # Realize cos/sin caches as ONNX graph initializers.
-            # The returned gathered embeddings are saved for potential reuse
-            # by KV-shared layers that fall back to standard Attention.
-            local_pos_emb = self.rotary_emb_local(op, position_ids)
-            global_pos_emb = self.rotary_emb_global(op, position_ids)
+            # Realize cos/sin caches as ONNX graph initializers so that
+            # the GQA op can reference them.
+            self.rotary_emb_local(op, position_ids)
+            self.rotary_emb_global(op, position_ids)
```
justinchuby: Could you double check if this is needed? Supposedly the cache is computed at initialization. But I could be wrong

justinchuby (approved):

src/mobius/models/gemma4.py
apsonawane: Without these calls, the cos/sin caches are never referenced in the graph, so onnxscript doesn't emit them as initializers

src/mobius/tasks/_gemma4.py
apsonawane: Fixed

src/mobius/models/gemma4.py
justinchuby: Maybe then do `_ = self.rotary_emb_local(op, position_ids)` and `_ = self.rotary_emb_global(op, position_ids)` because these calls have return values. This makes it clear we are explicitly discarding them.

src/mobius/models/gemma4.py
justinchuby: @apsonawane the definition is in _flags. We can remove if unused

src/mobius/models/gemma4.py
apsonawane: Updated

src/mobius/models/gemma4.py
apsonawane: Removed unused code

PR #280, opened 2026-05-06
update-genai integration to generate the config file

rui-ren:
# Why this PR

Mobius currently does not ergonomically expose `genai_config.json` generation when users build a `ModelPackage` themselves.

---

# The Problem

The natural Path B workflow today produces an output directory that `onnxruntime-genai` cannot load:

```python
pkg = mobius.build("Qwen/Qwen3-0.6B", load_weights=True)
pkg.save("/output/qwen3/")

# /output/qwen3/ now contains:
#   - model.onnx
#   - model.onnx.data
#
# But ORT-GenAI also requires:
#   - genai_config.json
#   - tokenizer files
#
# which save() does not write.
```

To make it loadable, the user must remember a second call:

```python
write_ort_genai_config(
    pkg,
    "/output/qwen3/",
    hf_model_id="Qwen/Qwen3-0.6B",
    ep="cuda",
)
```

If they forget, they end up with a complete-looking directory that silently fails at load time.

Path A (Model Builder) does not have this footgun — it always writes the config alongside the ONNX artifacts.

---

# What This PR Adds

A new integration helper: `export_package`

This combines:

- `pkg.save()`
- `write_ort_genai_config()`

into a single call with a fail-fast preflight.

```python
from mobius.integrations.ort_genai import export_package

pkg = mobius.build("Qwen/Qwen3-0.6B", load_weights=True)

export_package(
    pkg,
    "/output/qwen3/",
    hf_model_id="Qwen/Qwen3-0.6B",
    ep="cuda",
)

# /output/qwen3/ now contains:
#   - model.onnx
#   - model.onnx.data
#   - genai_config.json
#   - tokenizer files
#
# Directly loadable by ORT-GenAI.
```

This becomes the recommended Path B equivalent of:

```bash
python builder.py -m ... -o out -p fp16 -e cuda
```

---

# Three-Tier API After This PR

| Use Case | API | Notes |
|---|---|---|
| ONNX already on disk, only need configs | `write_ort_genai_config(pkg, dir, …)` | unchanged |
| Built `pkg` in memory, want save + config in one call | `export_package(pkg, dir, …)` | new |
| Have only an HF model id | `auto_export("HF/id", dir, …)` | now delegates to `export_package` |
---

# Changes

- Added:
  ```python
  mobius.integrations.ort_genai.export_package(
      pkg,
      output_dir,
      *,
      hf_model_id,
      ep,
      context_length,
      local_config_dir,
      external_data,
      progress_bar,
  )
  ```

- Refactored `auto_export()` to delegate to `export_package` (DRY)
- Fixed bug:
  - `auto_export()` was annotated as `-> dict[str, str]`
  - but did not actually return the result
- Added doc note on `ModelPackage.save()` directing users to `export_package`
  for ORT-GenAI-compatible directories
- Added tests:
  - `TestExportPackage`
  - validates:
    - artifacts written
    - save kwargs forwarded
    - config kwargs forwarded
    - fail-fast preflight
    - no I/O on missing config
    - manifest returned

---

# Migration Impact

After this lands, the Path A → Path B documentation becomes:

```diff
- python onnxruntime-genai/src/python/py/models/builder.py \
-     -m Qwen/Qwen3-0.6B -o /output/qwen3 -p fp16 -e cuda

+ from mobius import build
+ from mobius.integrations.ort_genai import export_package
+
+ pkg = build(
+     "Qwen/Qwen3-0.6B",
+     load_weights=True,
+     dtype="f16",
+ )
+
+ export_package(
+     pkg,
+     "/output/qwen3",
+     hf_model_id="Qwen/Qwen3-0.6B",
+     ep="cuda",
+ )
```

One Path B call, one Path A call, same output directory layout.

---

# CI Note

The L4/L5 GPU jobs failed with the self-hosted-runner eviction signature:

- step stuck in `in_progress`
- post-cleanup skipped
- ~20 minute runtime
- no actual test failure logged

The docstring change to `_model_package.py` triggered:

```python
run_all = true
```

inside:

```python
detect_affected_models.py
```

because `_model_package.py` matches `_SHARED_INFRA_PATTERNS`.

This caused the full L4/L5 sweep instead of the affected subset.

Re-running the failed jobs should pass.

### DIFF
--- a/src/mobius/_model_package.py
+++ b/src/mobius/_model_package.py
@@ -73,6 +73,13 @@ class ModelPackage(UserDict[str, ir.Model]):
         ``model.onnx`` in *directory*.  When multiple models are present,
         each is saved in its own subfolder as ``{name}/model.onnx``.
 
+        .. note::
+            This method writes ONNX files only.  If you need a directory that
+            ``onnxruntime-genai`` can load (i.e. with ``genai_config.json`` and
+            tokenizer files), use
+            :func:`mobius.integrations.ort_genai.export_package` instead — it
+            wraps :meth:`save` with the ORT-GenAI config-generation step.
+
         Args:
             directory: Path to the output directory (created if needed).
             external_data: External data format. ``"onnx"`` (default) saves
--- a/src/mobius/integrations/ort_genai/__init__.py
+++ b/src/mobius/integrations/ort_genai/__init__.py
@@ -7,7 +7,10 @@ All onnxruntime-genai specific code lives here. The core model/task/component
 layers remain runtime-agnostic.
 """
 
-from mobius.integrations.ort_genai.auto_export import write_ort_genai_config
+from mobius.integrations.ort_genai.auto_export import (
+    export_package,
+    write_ort_genai_config,
+)
 from mobius.integrations.ort_genai.ep_config import (
     make_genai_decoder_config,
     make_kv_cache_dim_name,
@@ -20,6 +23,7 @@ from mobius.integrations.ort_genai.genai_config import (
 
 __all__ = [
     "GenaiConfigGenerator",
+    "export_package",
     "write_ort_genai_config",
     "make_genai_decoder_config",
     "make_kv_cache_dim_name",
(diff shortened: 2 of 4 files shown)


justinchuby (approved):
Thank you!

justinchuby:
Feel free to merge after lint is fixed. No need to worry about the L4/L5 tests.

PR #282, opened 2026-05-07
Refresh skill docs for current multimodal/CUDA behavior

justinchuby:
Align skill docs with current mobius behavior: encoder inputs are declared with config.dtype, KV-shared Gemma4 GQA emits empty K/V with runtime dependency on ORT KV-shared support, and quality checklist now points to search.past_present_share_buffer in genai_config.json.


titaiwangms (commented):
## Review summary

Direction is right (treating opset lowering as a workaround, not the default), and the new checklist items are concrete. A few accuracy/freshness items below.

### Verifications

- ✅ mobius#281 — MERGED 2026-05-06 ("Disable opset lowering by default") — matches PR description
- ⚠️ microsoft/onnxruntime#28242 — currently **OPEN** ("[GQA] Make present_key/present_value outputs optional and add Gemma4 support"). The new section in `debugging-multimodal/SKILL.md` reads as if it has landed.
- Cross-PR consistency: PR #283 (also under review) references microsoft/onnxruntime#28318 for the same shared-KV GQA feature. #28318 is closed as DUPLICATE; #28242 is the live PR. Both files should converge on a single canonical link.

### Major

1. **`debugging-multimodal/SKILL.md` "GQA for KV-shared layers"** cites microsoft/onnxruntime#28242 as if the support exists. That PR is still open as of this review. A reader who sets `kv_sequence_length=0` on stock ORT today will hit the failure modes documented in the (deduped) issue #28318. Either:
   - phrase as *"requires ORT built with PR #28242 (not yet merged as of \<date\>)"*, or
   - hold this section until the PR lands.

2. **Cross-skill drift with PR #283.** Once both PRs land, the repo will document the same shared-KV GQA mechanism with two different upstream references (#28242 here, #28318 in #283). #28318 is closed as DUPLICATE — make #28242 canonical and update both files to match. This is a tiny coordination cost now and a real grep-hazard later.

### Minor

3. **"Current ORT" / "newer ORT versions" needs a version pin.** Three places replace prescriptions with phrases like *"On current ORT this is unnecessary"* and *"This has been fixed in newer ORT versions."* In 6 months *"current"* will be wrong. Pin the version where the kernel registrations actually landed (e.g. *"Fixed in ORT ≥1.25"*) so a future reader can act without changelog spelunking.

4. **`debugging-memcpy/SKILL.md` commented-out code.** The diff comments out the opset-lowering snippet rather than deleting it:
   ```python
   # if model.opset_imports.get("", 0) > 23:
   #     model.opset_imports[""] = 23
   ```
   Commented-dead-code-with-explanation is a smell — readers tend to copy the whole block including the comment. Either delete the lines (and rely on the NOTE comment alone) or extract the snippet into a clearly-labeled *"If running on ORT ≤1.24.x"* sub-section so it's clear it's a workaround path, not the default.

5. **`past_present_share_buffer` checklist item.** *"must be `false` for dual head_dim models (e.g. Gemma4: 256 for sliding, 512 for full attention)"* — head_dim numbers are stated as fact but vary by Gemma4 variant (E2B vs E4B). If 256/512 is correct specifically for E2B, say so; otherwise verify against the variant you actually measured. (PR #283 makes a related `head_dim=512` claim — same caveat applies; resolve them together.)

6. **`_cast_encoder_input()` reference.** Naming the helper is good. A one-liner pointing to where it's called from (e.g. *"called as the first step inside each encoder task's `build()`"*) would help a reader adding a new multimodal task spot when it's needed without grepping.

### Nit

7. The MEA / float-additive-bias note ("prefer float additive bias for complex attention patterns") is good. One-line *"why"*: bool masks broadcast inconsistently across batch with MEA, float additive bias doesn't. That's the actionable takeaway and the only reason a reader will remember the rule.

### Praise

- The directional shift (lowering opset is a workaround, not the fix) is the right framing — encourages fixing ORT instead of permanently shimming around it.
- The env-var name `MOBIUS_ORT_LOWER_OPSET_FOR_EP=1` is repeated consistently across all three files, keeping the workaround discoverable.
- The new `past_present_share_buffer` checklist item captures a real footgun in `genai_config.json` that's easy to miss.

justinchuby:

PR #283, opened 2026-05-07
docs: refresh skills for latest Gemma4 state and references

justinchuby:
Updates 3 MEDIUM priority skills with latest codebase-aligned findings and references:

### profiling-onnx-models
- Flash Attention gives only 1-3% benefit for Gemma4 (7/35 layers eligible with float masks, 25/35 with all-GQA)
- Quantization benchmark reference standardized to Gemma4 E2B-IT
- Q4_K_M vs NF4 benchmark: Q4_K_M wins on both CPU (+138%) and CUDA (+7%) vs F16; NF4 is -11% on CUDA and -41% on CPU vs F16
- Quality wording softened to “no observed quality regression in spot checks”

### onnx-export-quantization
- Removed duplicated inference speed table and now cross-references the canonical benchmark table in `profiling-onnx-models`
- Recommendation remains Q4_K_M over NF4, with wording aligned to measured results and spot-check quality language

### multimodal-models
- GQA shared KV section updated to match current implementation state in mobius
- Terminology clarified: `new_kv_length=0` mapped to graph wiring with empty K/V (`kv_sequence_length=0`)
- Removed stale caveat linked to duplicate-tracked ORT issue
- Float-domain bicubic resize reference clarified to `microsoft/onnxruntime-extensions#1056`
- f32 cast-at-entry already documented (confirmed, no change needed)

Documentation only — no code changes.

### DIFF
--- a/.agents/skills/multimodal-models/SKILL.md
+++ b/.agents/skills/multimodal-models/SKILL.md
@@ -291,7 +291,7 @@ earlier source layers instead of projecting their own. With GQA
 - `value=None` (empty)
 - `past_key=source_layer_present_key` (borrowed from source)
 - `past_value=source_layer_present_value` (borrowed from source)
-- `do_rotary=0` (RoPE already applied by source layer)
+- `do_rotary=1` (RoPE applied to queries; source K already has RoPE)
 
 This eliminates ~40 Transpose/Reshape ops per shared layer that
 were previously needed to convert the source KV from BNSH
--- a/.agents/skills/multimodal-models/SKILL.md
+++ b/.agents/skills/multimodal-models/SKILL.md
@@ -281,6 +281,33 @@ If you're adding a new multimodal model, you don't need to handle this
 manually — mobius inserts the Cast automatically for all encoder graphs.
 If the model dtype is already f32, no Cast is needed.
 
+## GQA for KV-shared layers (Gemma4)
+
+Models with KV-shared layers (e.g. Gemma4) borrow key/value from
+earlier source layers instead of projecting their own. With GQA
+(PR #279), shared layers use `GroupQueryAttention` with:
+
+- `key=None` (empty, `kv_sequence_length=0`)
+- `value=None` (empty)
+- `past_key=source_layer_present_key` (borrowed from source)
+- `past_value=source_layer_present_value` (borrowed from source)
+- `do_rotary=0` (RoPE already applied by source layer)
+
+This eliminates ~40 Transpose/Reshape ops per shared layer that
+were previously needed to convert the source KV from BNSH
+`[B, kv_heads, seq, head_dim]` to 3D `[B, seq, kv_hidden]` format.
+
+**Caveat:** Requires ORT to support `kv_sequence_length=0` in the
+GQA op. Tracked in microsoft/onnxruntime#28318.
+
+## Vision preprocessing: float-domain bicubic resize
+
+For HF parity in vision preprocessing, ort-extensions provides a
+float-domain bicubic resize (PR #1056) that avoids uint8 clamping
+artifacts. This eliminates 0.03-0.13 pixel differences that can
+cause token divergence in vision encoders. The utility operates in
+packed RGB with fused uint8→float rescale.
+
 ## Cross-references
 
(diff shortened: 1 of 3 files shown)


titaiwangms (commented):
## Review summary

Documentation-only PR adding 64 lines across 3 skill files. Numbers are concrete and cross-referenced PRs verified. A few accuracy / freshness items below, nothing blocking.

### Verifications
- ✅ mobius#279 — MERGED ("Enable GQA for KV-shared layers and fix dtype issues in vision and audio")
- ✅ microsoft/onnxruntime-extensions#1056 — MERGED ("Fix Gemma4 image resize to use float-domain bicubic for HF parity")
- ⚠️ microsoft/onnxruntime#28318 — **closed as DUPLICATE** on 2026-05-03. See issue below.
- ✅ Internal consistency: the Q4_K_M / NF4 table appears identically in `onnx-export-quantization/SKILL.md` and `profiling-onnx-models/SKILL.md`.
- ✅ Math: 16.2/6.8 ≈ +138%, 4.0/6.8 ≈ −41%, 93/104 ≈ −10.6%. **One off:** 111/104 = +6.7%, reported as "+8%" — should be "+7%".

### Major

1. **`multimodal-models/SKILL.md`** — the "Caveat: Tracked in microsoft/onnxruntime#28318" link points to an issue that was closed as a duplicate. A future reader following the link will land on a dead end. Either:
   - replace the link with the canonical tracking issue/PR that 28318 was deduped against, or
   - if support has actually landed, drop the caveat and note the minimum ORT version that supports `kv_sequence_length=0` for shared KV.

### Minor

2. **Duplicated table.** The Q4_K_M vs NF4 benchmark table is copy-pasted into both `onnx-export-quantization/SKILL.md` and `profiling-onnx-models/SKILL.md`. Two sources of truth = drift hazard the next time numbers update. Pick one canonical home (`profiling-onnx-models` reads more natural for a benchmark) and have the other cross-reference it, or at least add a "keep in sync with X" note.

3. **Benchmark provenance is thin.** "Gemma4 E2B-IT on GenAI (text-only, 50+ token decode)" omits hardware (CPU model, GPU model), CUDA / ORT versions, batch size, and exact sequence length. A one-line environment footer below the table would make the numbers reproducible and let future readers spot when they've gone stale. Suggested form:

   > _Measured on `<CPU>` / `<GPU>`, ORT `<x.y>`, CUDA `<v>`, batch=1, prompt=`<n>` tokens, decoded `<n>` tokens, `<date>`._

4. **"NF4 is paradoxically slower than F16"** is a surprising claim presented without a mechanism. One line on the suspected cause (likely: no fused dequant kernel for NF4 on either EP, so each step pays per-tensor dequant cost) would let a future reader judge whether the claim still holds after a kernel fix or after upgrading ORT.

5. **"no measurable quality degradation"** for Q4_K_M needs either a citation (perplexity, token-match, N-prompt semantic check) or a softening to "no observed degradation in spot checks". The earlier section in this same skill file emphasizes verifying that generated text is "coherent and semantically correct" — reuse that wording so the bar is consistent.

6. **Rounding:** "+8%" should be "+7%" (or "≈+7%"); "−10%" should be "−11%" (the actual is −10.6%). The +138% / −41% entries are accurate. Minor, but the rest of the table is precise.

### Nit

7. **`do_rotary=1` shared-KV note.** Worth one line clarifying that this assumes the source layer's `present_key` was emitted *after* RoPE (i.e. the upstream GQA writes the rotated K to `present_key`). If any configuration emits pre-RoPE `present_key`, RoPE would double-apply on the borrowed past, producing silently wrong outputs.

8. **`head_dim=512` for Gemma4 global attention** — sanity-check this; my recollection is global head_dim is 256 for Gemma4. If 512 is right for the variant you measured (E2B?), worth saying which.

### Praise

- Concrete numbers with %-deltas instead of vague "Q4 is faster".
- Cross-references to the actual PRs (#279, ort-extensions #1056) so a future reader can trace where a behavior came from.
- The GQA shared-KV pattern (`key=None`, borrowed `past_key`/`past_value`, `do_rotary=1`) is a non-obvious skill-shaped finding that future agents will need.

justinchuby:

PR #284, opened 2026-05-07
docs: fix GQA MEA head_size constraints in attention skill

justinchuby:
Fixes incorrect GQA MEA eligibility claims in the attention-optimization skill, based on critical reviewer's source code verification.

**Changes:**
- Contrib GQA table: Added explicit `head_size ≤ 256` constraint (hard guard at group_query_attention.cc:480, shared memory overflow concern)
- ONNX Attention GQA table: Added `nonpad_kv_seqlen` head_size ≤ 256 note
- GQA + float bias table: Clarified this applies to ONNX Attention path (not Contrib GQA which rejects bias)
- Key takeaways: Added head_size constraint item

**Context:** The MEA kernel itself supports head_size up to 1024, but GQA's eligibility check blocks it at 256. ONNX Attention is less restrictive (≤ 1024 standard, ≤ 256 only with `nonpad_kv_seqlen`).


justinchuby:

PR #288, opened 2026-05-08
update lm head last token pruning

rui-ren:
# PR Description

Adds opt-in final-token LM-head pruning as the cargo-style build feature:

```bash
mobius build --model meta-llama/Llama-3.2-1B output/ \
  --features prune-lm-head
```

The same behavior is available programmatically through `build(prune_lm_head=True)` and `build_from_module(..., prune_lm_head=True)`.

When enabled, Mobius selects the final hidden-state position before the LM-head projection, changing logits from `[B, S, vocab]` to `[B, 1, vocab]`. This avoids computing unused per-token vocabulary logits during prefill.

The default remains disabled to preserve workflows requiring per-token logits, including logprob scoring, speculative decoding, and multi-token generation. Models using unsupported custom forward paths fail explicitly instead of silently emitting an unoptimized graph.

This branch is updated with the latest `main`.


rui-ren:
@copilot  merge main branch into this PR.

justinchuby:
Maybe we should consider creating an option like we do for use_static_cache, as the feature flags are designed to be temporary and would eventually be removed when the feature stabilizes. Do you think it would make sense to create a new task for this? Or maybe create an option to an existing task?

rui-ren:
> Maybe we should consider creating an option like we do for use_static_cache, as the feature flags are designed to be temporary and would eventually be removed when the feature stabilizes. Do you think it would make sense to create a new task for this? Or maybe create an option to an existing task?

ok, sure, make sense to create a new task for this.

justinchuby:
@copilot please address comments and update from main

justinchuby:
@copilot update from main and make this a feature

justinchuby (approved):

PR #294, opened 2026-05-08
Migrate op_multi_out to __getattr__ dispatch with _outputs

justinchuby:
## Problem

`op.op_multi_out()` is being removed from onnxscript/onnx_ir. 9 call sites across 5 rewrite rule files use this API.

## Fix

Replace all `op.op_multi_out()` calls with the standard `__getattr__` dispatch pattern using `_outputs=N`:

```python
# Before
outputs = op.op_multi_out(
    'GroupQueryAttention',
    inputs=[q, k, v, past_k, past_v, seqlens, total_seq, cos, sin],
    domain='com.microsoft',
    attributes=gqa_attrs,
    num_outputs=3,
)

# After
outputs = op.GroupQueryAttention(
    q, k, v, past_k, past_v, seqlens, total_seq, cos, sin,
    _domain='com.microsoft',
    _outputs=3,
    **gqa_attrs,
)
```

## Files Changed (5 files, 9 calls)
- `_group_query_attention.py`: 4 calls
- `_skip_layer_norm.py`: 2 calls
- `_skip_norm.py`: 1 call
- `_unpack_qkv.py`: 1 call
- `_separate_rope.py`: 1 call

## Testing
- ✅ All 2718 tests pass
- ✅ Verified GQA (24 ops), SkipNorm (48 ops), RoPE separation all produce correct output on Qwen2.5-0.5B with CUDA EP

### DIFF
--- a/src/mobius/rewrite_rules/_group_query_attention.py
+++ b/src/mobius/rewrite_rules/_group_query_attention.py
@@ -647,7 +647,13 @@ class AttentionToGQA(RewriteRuleClassBase):
             gqa_attrs["softcap"] = softcap
 
         outputs = op.GroupQueryAttention(
-            q, k, v, past_key, past_value, self._seqlens_k, self._total_seq_len,
+            q,
+            k,
+            v,
+            past_key,
+            past_value,
+            self._seqlens_k,
+            self._total_seq_len,
             _domain="com.microsoft",
             _outputs=3,
             **gqa_attrs,
--- a/src/mobius/rewrite_rules/_group_query_attention.py
+++ b/src/mobius/rewrite_rules/_group_query_attention.py
@@ -244,22 +244,19 @@ class RotaryAttentionToGQA(RewriteRuleClassBase):
         }
         if softcap:
             gqa_attrs["softcap"] = softcap
-        outputs = op.op_multi_out(
-            "GroupQueryAttention",
-            inputs=[
-                q_pre,
-                k_pre,
-                v,
-                past_key,
-                past_value,
-                self._seqlens_k,
-                self._total_seq_len,
-                self._cos_cache,
-                self._sin_cache,
-            ],
-            domain="com.microsoft",
-            attributes=gqa_attrs,
-            num_outputs=3,
+        outputs = op.GroupQueryAttention(
+            q_pre,
+            k_pre,
+            v,
+            past_key,
+            past_value,
+            self._seqlens_k,
+            self._total_seq_len,
+            self._cos_cache,
+            self._sin_cache,
+            _domain="com.microsoft",
+            _outputs=3,
+            **gqa_attrs,
         )
 
         return outputs[0], outputs[1], outputs[2]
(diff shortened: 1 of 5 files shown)


titaiwangms:
Thanks for the PR! The `op_multi_out` migration itself is **approve-ready** — zero graph changes, 2718 tests passing, and it directly unblocks `onnxscript` 0.7.1 compatibility. The `hy_v3` model is structurally solid but has a few items needed before merge:

**Required before merge:**

1. **`create_attention_bias()` missing `dtype=self._dtype`** in `Hy3TextModel.forward` — without this, fp16/bf16 runs use a float32 mask value which can produce incorrect masking behavior (potential NaN in attention softmax on some hardware). Easy fix, high impact.

2. **Test for `moe_router_use_sigmoid=True` code path** — this logic in `_configs.py` silently overrides an explicit `scoring_func` value. It's untested, and exactly the kind of silent-priority rule that most needs a test.

3. **Test for text-only Hy3 (`num_local_experts=None`) fallback** — this code path is not covered. Both this and item 2 are flagged by codecov.

4. **Comment on config fallback priority ordering** — the `scoring_func`/`moe_router_use_sigmoid` priority rule in `_configs.py` needs a comment documenting intent (e.g. `# moe_router_use_sigmoid takes priority over scoring_func field`). Same for the `router_scaling_factor` and `num_shared_experts` fallback chains — these are silent model-specific couplings with no explanation.

**Recommended (this PR or follow-up):**

5. **Move `_DeepSeekMoEFFN` + `DeepSeekMoEGate` to `mobius/components/_moe.py`** — `hy_v3.py` imports `_DeepSeekMoEFFN` from `mobius.models.deepseek`, which is a cross-model private symbol import (models → models dependency, not components → models). Since `mobius/components/_moe.py` already houses `MLP` and is the established home for MoE primitives, moving these there (and renaming `_DeepSeekMoEFFN` → `DeepSeekMoEFFN`) makes both `deepseek.py` and `hy_v3.py` consumers of shared components rather than one importing from the other.

6. **`Hy3TextModel.forward` signature divergence** — `inputs_embeds` and `dtype` params present in `DeepSeekV3TextModel.forward` are absent here. Please confirm this is an intentional simplification for Hy3 (if so, a short comment would prevent future devs from 'fixing' it); if it's an oversight, it may be a correctness gap.

**Cosmetic:**

7. **Update PR description** — the current description only mentions the `op_multi_out` migration ("5 files, 9 calls"). The `hy_v3` model addition (217 lines) and the `_configs.py` fixes are not reflected. Worth updating so reviewers understand the full scope.

**Positive note:** The `hy_v3.py` module docstring and the `preprocess_weights` weight-key mapping table are genuinely excellent — clear, complete, and exactly the documentation standard we want for all new model files.

src/mobius/models/hy_v3.py
justinchuby, line 1: this shouldn’t be here

justinchuby:
@titaiwangms hy_v3  removed and isolated to another PR. Thanks

PR #295, opened 2026-05-08
fix: provide lm_head.weight for tied VLM decoders in preprocess_weights

justinchuby:
## Summary

Fix VLM decoder weight loading for models with `tie_word_embeddings=True` by using the shared `tie_word_embeddings()` utility.

## Problem

When a VLM model (e.g. Qwen-VL, InternVL) has `tie_word_embeddings=True`, HuggingFace stores only the embedding weight in the checkpoint — `lm_head.weight` is absent. The previous VLM decoder `preprocess_weights` code discarded any weight not under the `language_model.` prefix, so `lm_head.weight` was never populated. The ONNX model then had an uninitialized LM head, producing garbage logits.

## Fix

Use `tie_word_embeddings()` from `_weight_utils.py` after stripping the `language_model.` prefix. This copies the embedding tensor reference to the `lm_head.weight` key when it is missing.

## How ONNX Initializer Sharing Works End-to-End

The weight-sharing mechanism spans three layers:

1. **`tie_word_embeddings(state_dict)`** — When `lm_head.weight` is missing, assigns `state_dict["lm_head.weight"] = state_dict["model.embed_tokens.weight"]`. Both keys now point to **the same Python tensor object** (same `data_ptr()`).

2. **`apply_weights(model, state_dict)`** in `_weight_loading.py` — Iterates over the state dict and assigns tensors to ONNX initializers. It tracks `tensor.data_ptr()` to detect shared storage. When it encounters `lm_head.weight` and sees the same `data_ptr()` as the already-assigned `model.embed_tokens.weight`, it:
   - Calls `initializer.replace_all_uses_with(canonical)` to redirect all graph uses
   - Deletes the duplicate initializer from `model.graph.initializers`

3. **Result** — The saved ONNX file contains a **single copy** of the embedding table, used by both the Gather (embedding lookup) and MatMul (LM head projection) nodes. No duplication.

## Changes

- `src/mobius/models/qwen_vl.py`: Use `tie_word_embeddings()` in `preprocess_weights`
- `src/mobius/_weight_utils.py`: Enhanced docstring explaining the full sharing mechanism

### DIFF
--- a/src/mobius/_weight_utils.py
+++ b/src/mobius/_weight_utils.py
@@ -368,11 +368,21 @@ def tie_word_embeddings(
     embed_key: str = "model.embed_tokens.weight",
     head_key: str = "lm_head.weight",
 ) -> None:
-    """Ensure both embedding and LM head weights are present (tied).
+    """Ensure both embedding and LM head weights are present and tied.
 
-    If one of the two keys is missing, copies the other. This handles
-    the common HuggingFace ``tie_word_embeddings=True`` pattern where
-    only one of the two weights is stored in the checkpoint.
+    When HuggingFace saves a model with ``tie_word_embeddings=True``, only
+    one of the two weights (embedding or LM head) is stored in the
+    checkpoint.  This function fills in the missing key by assigning it
+    to **the same Python tensor object** as the existing key.
+
+    This identity relationship is what enables true ONNX weight sharing
+    downstream: :func:`~mobius._weight_loading.apply_weights` detects
+    that two state-dict entries share the same underlying storage (via
+    ``data_ptr()``), creates a **single** ONNX initializer for the first
+    occurrence, and redirects all graph uses of the second initializer to
+    point at the first one.  The result is one copy of the embedding
+    table in the ONNX file, used by both the Gather (embedding lookup)
+    and MatMul (LM head projection) nodes.
 
     Mutates *state_dict* in place.
 
--- a/src/mobius/_weight_utils.py
+++ b/src/mobius/_weight_utils.py
@@ -396,11 +396,10 @@ def tie_word_embeddings(
     elif embed_key not in state_dict and head_key in state_dict:
         state_dict[embed_key] = state_dict[head_key]
     elif embed_key not in state_dict and head_key not in state_dict:
-        logger.warning(
-            "tie_word_embeddings: neither '%s' nor '%s' found in state_dict; "
-            "tying has no effect.",
-            embed_key,
-            head_key,
+        raise ValueError(
+            f"tie_word_embeddings: neither '{embed_key}' nor '{head_key}' "
+            f"found in state_dict. Check that the key names match the "
+            f"weight dictionary after any prefix stripping."
         )
 
 
--- a/src/mobius/_weight_utils.py
+++ b/src/mobius/_weight_utils.py
@@ -395,6 +395,13 @@ def tie_word_embeddings(
         state_dict[head_key] = state_dict[embed_key]
     elif embed_key not in state_dict and head_key in state_dict:
         state_dict[embed_key] = state_dict[head_key]
+    elif embed_key not in state_dict and head_key not in state_dict:
+        logger.warning(
+            "tie_word_embeddings: neither '%s' nor '%s' found in state_dict; "
+            "tying has no effect.",
+            embed_key,
+            head_key,
+        )
 
 
 def vlm_decoder_weights(
--- a/src/mobius/_weight_utils_test.py
+++ b/src/mobius/_weight_utils_test.py
@@ -227,11 +227,13 @@ class TestTieWordEmbeddings:
         assert sd["model.embed_tokens.weight"].data_ptr() == t1.data_ptr()
         assert sd["lm_head.weight"].data_ptr() == t2.data_ptr()
 
-    def test_neither_present_no_change(self):
-        """If neither present, no change."""
+    def test_neither_present_raises(self):
+        """Raise ValueError if neither key is found (catches key mismatches)."""
+        import pytest
+
         sd = {"other.weight": torch.randn(4)}
-        tie_word_embeddings(sd)
-        assert list(sd.keys()) == ["other.weight"]
+        with pytest.raises(ValueError, match=r"neither.*found"):
+            tie_word_embeddings(sd)
 
     def test_custom_keys(self):
         """Custom embed/head keys."""
--- a/src/mobius/models/qwen_vl.py
+++ b/src/mobius/models/qwen_vl.py
@@ -120,7 +120,8 @@ class Qwen25VLCausalLMModel(nn.Module):
                 renamed[f"decoder.{key}"] = value
                 stripped = key[len("model.") :]
                 renamed[f"embedding.{stripped}"] = value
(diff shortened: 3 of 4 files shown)


justinchuby:
I don't think we should copy them. The weights should be shared

titaiwangms (commented):
## Review summary

Direction is right and existing tests pass cleanly (2736 unit tests in 32s, 115 qwen tests), but adversarial trace turned up one path that's still silently broken, plus a coverage gap that lets it slip through the suite.

### Major

1. **`Qwen3VLDecoderModel.preprocess_weights()` is silently still broken.** The rename loop (`qwen_vl.py:843-852`) strips both `model.` and `language_model.` unconditionally, so for an HF key like `model.language_model.embed_tokens.weight` the dict ends up with `embed_tokens.weight` (no `model.` prefix). The new call `tie_word_embeddings(renamed)` uses defaults `embed_key="model.embed_tokens.weight"`, `head_key="lm_head.weight"` — neither is present, so the helper silently no-ops and `lm_head.weight` is never created. The standalone Qwen3-VL decoder build with tied weights will hit the same `ValueError: Component 'decoder' has 1 initializer(s) without weights: 'decoder.lm_head.weight'` this PR is trying to fix.
   Either:
   ```python
   tie_word_embeddings(renamed, embed_key="embed_tokens.weight", head_key="lm_head.weight")
   ```
   or rework the rename so `model.embed_tokens.weight` lands in `renamed` (so it matches the standalone ONNX initializer name and keeps default-args semantics consistent with `Qwen25VLDecoderModel`).

2. **No regression test for any of the four fixed paths.** The full suite passes, but:
   - No test invokes `preprocess_weights` on `Qwen25VLCausalLMModel`, `Qwen25VLDecoderModel`, `Qwen3VL3ModelCausalLMModel`, or `Qwen3VLDecoderModel`.
   - All qwen-VL graph-build tests use `tie_word_embeddings=False`.
   - Graph-build tests don't call `preprocess_weights` at all — the failure surfaces at `apply_weights` time.
   - The only qwen-VL `preprocess_weights` test (`test_qwen35_vl_preprocess_weights_model_prefix`) is for a different class in `qwen35.py` and doesn't set tying.
   Add ~4 cheap co-located tests in `src/mobius/models/qwen_vl_test.py` with a tiny fake state dict and `tie_word_embeddings=True`, asserting `decoder.lm_head.weight in renamed` and `data_ptr()` identity with the embed. These would have caught both the original bug and bug #1 above.

3. **Missing "why" comment at the composite tie sites.** The deleted comment was wrong; nothing replaced it. The non-obvious reason — `onnxscript` qualifies parameter names by module path, so the `__init__`-time alias `decoder.lm_head.weight = decoder.model.embed_tokens.weight` doesn't cross composite module boundaries — is the *entire* reason this PR exists. Without a comment, the next person will re-delete the block. Suggested:
   ```python
   # onnxscript qualifies params by module path, so the in-tree alias set in
   # Qwen25VLDecoderModel.__init__ does not cross composite boundaries. Establish
   # the identity here so apply_weights sees a single data_ptr() across both
   # initializers.
   if self.config.tie_word_embeddings:
       tie_word_embeddings(renamed, embed_key=..., head_key=...)
   ```

### Minor

4. **`_weight_utils.tie_word_embeddings` silent no-op masks routing bugs.** When `tie=True` and *both* configured keys are absent, the helper does nothing — which is exactly what surfaces as bug #1. Recommend: raise (or at least log a clear warning naming both keys) when neither key is present and the caller said `tie=True`. Tightly coupled to this PR; worth rolling in.

5. **`_weight_utils.tie_word_embeddings` docstring says "copies"; the implementation is identity assignment.** `state_dict[head_key] = state_dict[embed_key]` is a Python reference assignment, same object, same `data_ptr()`. The PR's correctness depends on identity, not copy. The next maintainer reading "copies" will reasonably "fix" it to `clone()` and silently break the tying. Update the docstring to say something like *"assigning it to the same Python tensor object so downstream consumers see one `data_ptr()`."* (Note: the diff/branch does not currently change `_weight_utils.py` — please ship the docstring fix with this PR.)

6. **Composite preprocess drops `lm_head.*` before the helper runs.** In `Qwen25VLCausalLMModel:126-128` and `Qwen3VL3ModelCausalLMModel:788-790`, when `tie=True` the loop skips `lm_head.*` keys entirely. If a (non-standard but legal) checkpoint stored only `lm_head.weight` and not the embed, that value is dropped before `tie_word_embeddings` could backfill the embed. Defense-in-depth: keep `lm_head.*` in `renamed` regardless and let the helper normalize.

7. **`Qwen2VLCausalLMModel.preprocess_weights` (line 414) not migrated.** After this PR the file has three different idioms for the same concept:
   - inline assignment (Qwen2VL)
   - post-loop helper with explicit keys (composite Qwen2.5/3-VL)
   - post-loop helper with defaults (standalone Qwen2.5/3-VL decoders)
   Plus the legitimately-different `.pop()` in single-model `Qwen3VLCausalLMModel`. Migrate Qwen2VL to the helper now, or add a `# TODO: align with Qwen2.5/Qwen3 pattern` comment so the asymmetry is flagged rather than invisible.

8. **Stale comment at `Qwen3VLCausalLMModel:718`.** The wording `# lm_head.weight is tied at graph level; discard any separate checkpoint entry.` is word-for-word the comment this PR deletes as wrong elsewhere. For this single-model class it's accurate (in-tree alias survives onnxscript), but the wording reads ambiguously after this PR. Tighten to make the single-model-vs-composite distinction explicit, e.g. *"Single-model build: onnxscript preserves the in-tree alias, so the `lm_head` checkpoint entry is redundant — pop it."*

### Nit

9. `embed_key`/`head_key` strings (`"decoder.model.embed_tokens.weight"` / `"decoder.lm_head.weight"`) duplicated in two composites — module-level constants if a third composite lands.
10. `Qwen3VLDecoderModel.preprocess_weights` docstring is one line; sister `Qwen25VLDecoderModel` has a four-line docstring covering HF key format and standalone-vs-composite distinction. Match the level.
11. `# Handle weight tying` at `qwen_vl.py:191` replaces a wrong-but-explicit comment with a content-free one. One line: *"HF checkpoint omits lm_head.weight when tied; fill it in for the initializer."*

### Praise

- Right call using the existing helper instead of growing a new pattern.
- Both composite paths and both standalone-decoder paths get the same treatment in one PR — good symmetry (modulo bug #1).
- Commit messages name the wrong assumption, the failure mode, and the L1/L5 verification — exactly the right context for a non-obvious weight-loading fix.

### QA / test run

- `python -m pytest tests/build_graph_test.py tests/cli_test.py src/ -q -k "not phi4mm and not apply_weights_unknown" --tb=short -n auto` → **2736 passed, 41 skipped, 0 failed** in 32.4s.
- Qwen-only subset → **115 passed, 24 skipped, 0 failed** in 12.6s.
- Lint cannot run in the test env: `ruff 0.12.12` installed, `pyproject.toml:133` references `RUF067` requiring `ruff 0.15.11`. **Pre-existing environment mismatch, not caused by this PR.**

---

Net: the fix's intent is correct and the existing 3-model composite paths trace cleanly, but bug #1 (`Qwen3VLDecoderModel`) and the absent regression coverage mean this still slips the same class of failure. Worth one more pass.

PR #298, opened 2026-05-09
Fix Qwen3-VL example script and add CLI options

justinchuby:
## Summary

Fix `examples/qwen3_vl_ort_genai.py` crash and add CLI options.

## Problem

Running `python examples/qwen3_vl_ort_genai.py` crashes with:
```
RuntimeError: Invalid input name: input_ids
```

**Root cause**: Qwen3-VL 3-model split decoder expects `inputs_embeds` (not `input_ids`). ORT GenAI's `append_tokens` and `set_inputs` with `input_ids` don't route through the embedding model for text-only generation. This is an ORT GenAI runtime limitation for 3-model VLM pipelines.

## Fix

- Replace broken text-only `generate()` with clear error directing users to `--image`
- Add `--device`, `--ep`, `--dtype` CLI options (matching other examples)
- Pass dtype/ep through to `build_and_export`
- Update docstring documenting the limitation

## Usage (after fix)

```bash
# Multimodal generation (works):
python examples/qwen3_vl_ort_genai.py --image testdata/pipeline-cat-chonk.jpeg

# Build with specific dtype/EP:
python examples/qwen3_vl_ort_genai.py --dtype bf16 --ep cuda --image path/to/image.jpg

# Build and save only:
python examples/qwen3_vl_ort_genai.py --save-to output/qwen3vl/ --dtype f16 --ep cuda
```

### DIFF
--- a/examples/qwen3_vl_ort_genai.py
+++ b/examples/qwen3_vl_ort_genai.py
@@ -6,10 +6,15 @@
 
 Builds the 3-model ONNX package (decoder, vision encoder, embedding),
 saves it in the flat layout expected by onnxruntime-genai, and runs
-text generation — with or without an image.
+multimodal generation with an image.
 
 Qwen3-VL uses the same 3-model I/O contract as Qwen2.5-VL, so it can
 be loaded by onnxruntime-genai with ``model.type = "qwen2_5_vl"``.
+
+NOTE: Text-only generation is not supported with the 3-model VLM split
+because the decoder expects ``inputs_embeds`` and ORT GenAI does not
+currently route ``input_ids`` through the embedding model for text-only.
+Use ``--image`` for all generation.
 
 Requirements::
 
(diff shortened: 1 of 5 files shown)


examples/qwen3_vl_ort_genai.py
```diff
@@ -2,344 +2,377 @@
...
-Usage::
+Supported dtype/EP combinations::
+
+    - CPU:  f32
+    - CUDA: f32, f16, bf16
 
-    # Text-only generation:
-    python examples/qwen3_vl_ort_genai.py
+    For f16 on CUDA, the decoder runs on CPU to avoid ORT CUDA EP
+    Memcpy-induced NaN with half-precision RMSNorm at opset 24.
```
justinchuby: Use nightly

examples/qwen3_vl_ort_genai.py
```diff
@@ -2,344 +2,377 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
-"""Qwen3-VL generation with onnxruntime-genai.
+r"""Qwen3-VL multimodal generation with ONNX Runtime.
```
justinchuby, line 5: Use genai

PR #300, opened 2026-05-14
Switch ORT GenAI export to Model Package layout

xiaoyu-work:
Mobius's `--runtime ort-genai` export now produces an EP-agnostic ORT-GenAI Model Package (v4 spec) directly loadable by the `onnxruntime-genai/src/models/model_package.cpp` loader, instead of the legacy flat directory with an EP-specific genai_config.json.

Variant decisions (session_options, provider_options, EP-specific tuning, multi-variant assembly) are deferred to downstream packagers (Olive workflows, EP-specific compilers).

Layout produced:

    <pkg>/
      manifest.json
      configs/{genai_config.json, tokenizer*, *_processor.json}
      <component>/metadata.json
      <component>/base/{variant.json, model.onnx, model.onnx.data*}

Mobius emits exactly one variant named `base`, declared compatible with the major ORT EPs (CPU + CUDA + DML + WebGPU + NvTensorRTRTX).

Schema details (per v4 spec, verified against the C++ loader):

* manifest.json: schema_version "1.0" + components string array.
* metadata.json: variants is a *map* keyed by variant name; each entry has ep_compatibility[] of {ep, device?, compatibility?}.
* variant.json: files[] entries carry filename + optional session_options/provider_options/shared_files (objects, not arrays) + optional consumer_metadata.genai_config_overlay.

genai_config.json schema changes (mobius output):

* removed: model.<role>.filename (replaced by .component)
* removed: model.<role>.session_options / .provider_options
* removed: search.past_present_share_buffer (was EP-derived)
* removed: search.max_length cap (was EP-derived)
* added:   model.<role>.component (package component name)

New helpers:

* mobius/integrations/ort_genai/_package_writer.py: writers for manifest/metadata/variant.json. Validates that optional object fields (provider_options, shared_files, session_options) are dicts at write time; omits them when empty so the loader sees absent (not empty array).
* ModelPackage.save_package_layout(): writes <dir>/<component>/<variant>/model.onnx for any number of components.

Verified end-to-end:

* `python -m mobius build --model microsoft/Phi-4-mini-instruct --dtype f16 --runtime ort-genai --no-weights <out>` produces a package that round-trips through the v4 schema verifier.
* 2732 passed, 41 skipped (full mobius test suite).
* 16 _package_writer_test.py tests including round-trip checks and rejection of list-form provider_options/shared_files.

Docs updated: ort-genai-config, onnx-export-quantization, foundry-local skills now describe the new package layout.

### DIFF
--- a/.agents/skills/foundry-local/SKILL.md
+++ b/.agents/skills/foundry-local/SKILL.md
@@ -100,8 +100,13 @@ cp output/configs/* "${DST}/"
 # <component>/ subdirectories (multi-component).
 for comp in decoder embedding vision_encoder audio_encoder; do
   if [ -d "output/${comp}/base" ]; then
-    if [ "$(ls -A output)" = "decoder" ] || [ ! -d "output/embedding" ]; then
-      # Single-component LLM: flatten model files into the model root
+    if [ "${comp}" = "decoder" ] \
+      && [ ! -d "output/embedding" ] \
+      && [ ! -d "output/vision_encoder" ] \
+      && [ ! -d "output/audio_encoder" ]; then
+      # Single-component LLM: flatten decoder files into the model root.
+      # Check sibling component dirs explicitly because a real Model
+      # Package also contains configs/ and manifest.json at the top level.
       cp output/${comp}/base/* "${DST}/"
     else
       # Multi-component: keep one subdir per component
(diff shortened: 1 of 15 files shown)


xiaoyu-work:
@copilot fix lint issue

PR #321, opened 2026-05-28
Enhance embedding introspection for GenAI config

apsonawane:
## Summary

Two fixes for Gemma4 multimodal export and genai_config generation.

### Introspect embedding model outputs in genai_config.json

The Gemma4 embedding model produces `per_layer_inputs` alongside `inputs_embeds`, but `genai_config.json` only listed `inputs_embeds` in the embedding outputs section. This caused onnxruntime-genai to not bind the `per_layer_inputs` output, so ORT allocated a separate buffer and the decoder received uninitialized data.

Added `_introspect_outputs()` to auto-discover embedding model outputs from the ONNX graph, and `embedding_output_names` parameter to `GenaiConfigGenerator.with_vision()`.

**Files:** `integrations/ort_genai/auto_export.py`, `integrations/ort_genai/genai_config.py`

### Pass `--ep` to mobius CLI in export script

`build_and_export()` in `gemma4_genai.py` did not pass the execution provider to `mobius build`. This caused all exports to use the default EP (which has empty `gqa_dtypes`), so KV-shared layers fell back to standard Attention instead of GQA — leading to `past_present_share_buffer` failures at runtime.

Now `--device cuda` correctly maps to `--ep cuda`, producing all-GQA models.

**Files:** `examples/gemma4_genai.py`

### Testing

- CPU fp32 export: GQA=35, Attention=0 ✅
- CUDA fp16 export: GQA=35, Attention=0 ✅
- Text-only generation (CPU): clean output ✅
- Multimodal image+text generation: no NaN, pixel_position_ids extracted correctly ✅
- per_layer_inputs forwarded from embedding to decoder ✅
- Backward compatible: existing non-Gemma4 models unaffected ✅

### DIFF
--- a/examples/gemma4_genai.py
+++ b/examples/gemma4_genai.py
@@ -77,8 +77,10 @@ def build_and_export(
     output_dir: str,
     *,
     dtype: str = "f32",
+    device: str = "cpu",
 ) -> None:
     """Build ONNX model with mobius CLI and write ORT GenAI config."""
+    ep = "cuda" if device == "cuda" else "cpu"
     cmd = [
         sys.executable,
         "-m",
@@ -88,6 +90,8 @@ def build_and_export(
         model_id,
         "--dtype",
         dtype,
+        "--ep",
+        ep,
         "--optimize",
         "--runtime",
         "ort-genai",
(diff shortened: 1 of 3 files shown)


justinchuby:
Wasn't the decision to cast inputs in genai? I just merged a PR for that behavior. Or is there a technical difficulty with vision (overflow etc. )?

justinchuby (approved):
Approving to unblock. LMK the state of the input contract

apsonawane:
yes, sorry for the confusion. Casting is happening in genai. I reverted it here

justinchuby (approved):

PR #323, opened 2026-05-29
introspection: add list_components for lightweight component discovery

xiaoyu-work:
Adds mobius.list_components(model_id, *, task=None, trust_remote_code=False) -> list[str] which returns the ONNX component names that mobius.build() would produce, without downloading weights or constructing any graphs.

For transformer models it downloads only the HuggingFace config.json (via transformers.AutoConfig) and dispatches via the same registry + task lookup as build(), including the qwen3_5_moe -> qwen3_5_moe_vl override. For diffusers pipelines it downloads only model_index.json and mirrors the iteration, filtering, and sub-package flattening rules of build_diffusers_pipeline so the returned names exactly match the keys of the resulting ModelPackage.

Component names are read statically from each task's ComponentSpec (or default to ['model'] for single-component tasks).

Tests cover: per-task component lookup, composite-config unwrapping, transformer + diffusers paths, explicit task override, and graceful errors when neither config.json nor model_index.json are usable.

### DIFF
--- a/src/mobius/__init__.py
+++ b/src/mobius/__init__.py
@@ -39,6 +39,7 @@ __all__ = [
     "ep_registry",
     "get_build_dtype",
     "get_ep",
+    "list_components",
     "models",
     "optimize_model",
     "register_ep",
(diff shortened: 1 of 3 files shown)


justinchuby (commented):
LGTM thanks. Could you address the copilot reviews? They seem relevant

PR #331, opened 2026-06-04
Emit `component` field per role in genai_config.json

xiaoyu-work:
ORT-GenAI's model-package loader uses the `model.<role>.component` field to find which package-component sub-directory the role's ONNX file lives in. When omitted, downstream packaging tools have to derive it heuristically from the role's filename path.

Add a small `_component_field()` helper that returns `{"component": <leading dir>}` for any role whose filename has a leading directory (e.g. `vision_encoder/model.onnx` -> `vision_encoder`), and `{}` otherwise so callers can spread it into a role dict unconditionally. Apply it to all four ORT-GenAI roles emitted by `GenaiConfigGenerator`:

- `model.vision`
- `model.embedding`     (VLM token-embed bridge)
- `model.speech`        (audio encoder)
- `model.decoder`

Text-only models with a flat `model.onnx` get no component field, matching today's single-component layout.

### DIFF
--- a/src/mobius/integrations/ort_genai/genai_config.py
+++ b/src/mobius/integrations/ort_genai/genai_config.py
@@ -13,7 +13,23 @@ from __future__ import annotations
 
 import json
 import os
+from pathlib import PurePosixPath
 from typing import Any
+
+
+def _component_field(filename: str | None) -> dict[str, str]:
+    """Return ``{"component": <leading dir>}`` for a sub-dirred filename.
+
+    Each model role's ONNX file lives under its own sub-directory (e.g.
+    ``vision_encoder/model.onnx``); the leading directory doubles as the role's
+    component label. Returns ``{}`` when ``filename`` has no leading directory
+    so callers can spread the result into a role dict unconditionally.
+    """
+    if filename:
+        parts = PurePosixPath(filename).parts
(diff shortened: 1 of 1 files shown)


src/mobius/integrations/ort_genai/genai_config.py
```diff
@@ -13,9 +13,25 @@
 
 import json
 import os
+from pathlib import PurePosixPath
 from typing import Any
 
 
+def _component_field(filename: str | None) -> dict[str, str]:
+    """Return ``{"component": <leading dir>}`` for a sub-dirred filename.
```
justinchuby, line 21: Are the field names predefined, or arbitrary? Is it possible to get it directly from the model package or some new input to with_vision, with_audio etc.? What if the file name does not follow this path construction convention?

xiaoyu-work: it is from mobius path, i guess mobius hard coded its component name?

justinchuby: The component path can still be user specified, even though it is currently following the convention. Is it possible to directly get the component name? Or, how is this field used?

jambayk:
closing this since it's not required anymore

PR #340, opened 2026-06-05
fp16/GQA export fixes + capture-safe static-cache attention (supersedes #328)

### LINKED ISSUE #341: Static-cache: add opt-in strict mode for partial present-shape params in _register_kv_cache_outputs
## Summary

Introduced in #340. `_register_kv_cache_outputs` (`src/mobius/tasks/_cache_utils.py`, the `~_register_kv_cache_outputs` helper, +63/-2 from a former 3-line stub) stamps explicit `present.*` output shapes/dtypes when **all** present-shape params are supplied. When only a **partial** set of present-shape params is provided, it currently emits `logger.warning(...)` and **silently falls back** to the known-wrong GQA shape-inference path (fail-open). A wiring bug therefore ships a structurally-wrong model with only a log line — no hard failure.

## Risk

Fail-open behavior: a partial/misconfigured present-shape set produces an incorrect GQA static-cache model that may run but produce wrong results, with nothing but a warning to signal it. This is easy to miss in automated export pipelines that don't scrape logs.

## Proposed fix

Add an opt-in `strict: bool = False` parameter (or a module-level config flag) that **raises `ValueError`** on a partial present-shape set instead of warn-and-proceed, listing the provided vs. missing params. The default stays `warn` for backward compatibility; production export paths opt into `strict`. Add a unit test asserting the raise.

**Scope:** small, isolated to `_cache_utils.py` + its test. Non-blocking.

## Provenance

- **Origin:** introduced by #340 (the function was a 3-line stub at the PR merge-base `5423421`). Deferred from #340's review as a tracked follow-up.

## Note on Major-3b (separate / out of scope)

The related `apply_weights()` INFO-only logging of unmapped weights (`src/mobius/_model_package.py:306`) is **pre-existing on `main`** — `_model_package.py` is untouched by #340 (`git diff --stat 5423421..c2bedd6` is empty for it). It is explicitly **out of #340 scope** and is **not** part of this follow-up. It is lower priority (a save-time `_check_weights` already raises on missing initializers); track separately if desired.

titaiwangms:
# fp16/GQA export fixes + capture-safe static-cache attention (Option-Y)

## Purpose

This PR delivers two related improvements to ONNX export:

1. **fp16 / GQA export correctness** — fixes constant-folding and grouped-query
   attention (GQA) handling so fp16/bf16 models export with correct present-KV
   shapes and dtypes.

2. **Capture-safe static-cache attention ("Option-Y")** — a static (pre-allocated,
   right-padded) KV-cache attention configuration that is **CUDA-graph-capture-safe**
   and actually runnable on ORT today.

   The static-cache path uses `is_causal=0` + an **explicit causal mask** +
   `nonpad_kv_seqlen` (opset-24 `Attention` input #6). With an explicit `attn_mask`
   present, ORT routes the op to the **Memory-Efficient Attention (MEA)** kernel,
   which applies both the per-batch `nonpad_kv_seqlen` key bound and the causal mask.
   This is currently the **only CUDA-runnable static-cache configuration**: the
   opset-24 kernel rejects `is_causal=1` together with `nonpad_kv_seqlen` when
   `S_q != total_kv` (returns `NOT_IMPLEMENTED`), until the `is_causal`+`nonpad` Flash
   path lands upstream (onnx/onnx#8054).

> ## ⚠️ Upstream coordination & temporary-workaround status
>
> **Option-Y is a deliberate, temporary workaround — not the intended end-state.**
> This section documents the cross-repo plan so reviewers understand what is permanent
> and what will be reverted.
>
> - **Upstream spec fix (the contract): [onnx/onnx#8068](https://github.com/onnx/onnx/pull/8068).**
>   #8068 fixes the opset-24 (and opset-23 errata) `Attention` `is_causal` semantics to
>   use **bottom-right (offset-aware)** causal alignment on the `nonpad_kv_seqlen` /
>   no-`past_key` path — i.e. it makes the ONNX spec + reference agree with what production
>   kernels already do for cached decode. This is the upstream contract Option-Y is written
>   against. _(Implements the RFC in onnx/onnx#8054.)_
>
> - **Why Option-Y exists today.** Even with #8068 defining the correct semantics, **today's
>   ORT kernel hard-rejects** `is_causal=1` + `nonpad_kv_seqlen` (no `past_key`) when
>   `S_q != total_kv` (`NOT_IMPLEMENTED`). So we cannot emit the simple maskless
>   `is_causal=1` form and run on CUDA yet. Option-Y (`is_causal=0` + explicit causal mask +
>   `nonpad_kv_seqlen`, routed to MEA) is the **only currently-runnable** static-cache
>   configuration. mobius therefore does **not hard-block** on #8068 — it ships a
>   runnable-today formulation so it can merge independently.
>
> - **The node-count delta is intentional and maintainer-accepted, pending revert.** The
>   explicit Option-Y mask adds nodes vs the maskless `is_causal=1` baseline, which is what
>   trips the benchmark **"Compare results"** node-count gate (see *Known items* below). This
>   is a deliberate correctness/runnability trade-off, **not a regression**, and we are
>   requesting maintainer-accept of the delta **for the duration of the workaround only**.
>
> - **Revert plan.** Once **both** [onnx/onnx#8068](https://github.com/onnx/onnx/pull/8068)
>   (spec/reference) **and** the **ORT runtime kernel update**
>   ([microsoft/onnxruntime#28904](https://github.com/microsoft/onnxruntime/issues/28904)) land, mobius will revert Option-Y to **`is_causal=1` + `nonpad_kv_seqlen`
>   with no explicit mask**. That removes the extra nodes, clears the node-count gate, and
>   **re-enables Flash-attention eligibility** for this configuration. The mobius-side revert
>   is tracked in **[onnxruntime/mobius#345](https://github.com/onnxruntime/mobius/issues/345)**.
>
> **Dependency chain:** onnx#8068 (spec) → ORT kernel update ([microsoft/onnxruntime#28904](https://github.com/microsoft/onnxruntime/issues/28904)) →
> mobius revert ([onnxruntime/mobius#345](https://github.com/onnxruntime/mobius/issues/345), drops Option-Y mask, unlocks Flash).

### Major-2 — hoist the static-cache causal mask (build once, share across layers)

The static-cache causal mask depends only on shapes + `write_indices`, so it is
**layer-invariant**. It is now built **once** at the task level and threaded as a
single `ir.Value` into every decoder layer's attention, instead of being rebuilt
per layer.

- Output is **bit-identical by construction** — same mask tensor, fewer graph nodes.
- New tests assert the mask is built once and that all layers share the **same**
  `ir.Value` (identity), plus a `None`-fallback parity test for the non-static path.

### Major-1 — intentionally NOT adding a graph-level `key_positions < nonpad` term

A reviewer suggested adding an explicit `And(causal, key_positions < nonpad)` term to
the static-cache mask to defend the padding bound. After investigation we are
**intentionally not adding it** — it is redundant:

- `nonpad_kv_seqlen` is **standard `ai.onnx` opset-24 `Attention` input #6** (absent in
  opset 23). Any conformant opset-24 runtime **must** mask key positions
  `j >= nonpad_kv_seqlen[b]` as the per-batch key bound, **independent of** the explicit
  `attn_mask`.
- Verified three independent ways: **empirically** (poisoning KV slots `[nonpad, S_kv)`
  yields `max|Δ| = 0` on both CUDA and CPU; non-vacuous positive control changes output),
  against **ORT source** (MEA applies both `seqlens_k` right-padding and `attn_bias`; the
  cutlass key loop hard-stops at `k_end = nonpad[b]`), and against the **ONNX spec**.
- The **causal half** of the mask (`GreaterOrEqual` on positions) **is** kept — it is
  load-bearing because `is_causal=0` means the op does not apply causality itself.
- A scalar `nonpad_kv_seqlen` can only express a **compact valid prefix** `[0, nonpad)`,
  never ragged/interior holes — so static caches rely on the contiguous-fill invariant
  (padding only on the right), documented as a precondition.

The mask-builder docstring and CHANGELOG were reframed to state this honestly (the
padding bound is enforced by the kernel via input #6, not by a redundant mask term).

### Major-3a — `_cache_utils` present-shape stamping is now fail-closed

`_register_kv_cache_outputs` (`src/mobius/tasks/_cache_utils.py`) takes six
"present-shape" parameters that are **all-or-nothing**: pass all six to stamp explicit
`present.*` types (required so `GroupQueryAttention` exports declare the correct
`head_dim`), or none to opt out and infer. A **partial** set (1–5 of 6) is always a
wiring slip — and previously it merely **logged a warning and proceeded**, shipping a
structurally-wrong model (mis-derived `present.*` `head_dim`) with only a log line.

This is now **fail-closed**: a partial present-shape set **raises `ValueError`** naming
the provided and missing parameters, instead of warn-and-proceed. All current call sites
pass either **0** parameters (intentional infer opt-out) or **all 6** (stamp), so there
is **no production regression** — the raise only fires on a genuine future wiring bug.
A unit test that previously asserted the silent fallback now asserts the raise (the
regression proof).

## Known items

- **Benchmark "Compare results" node-count increase** is the **intended structural
  cost** of the explicit Option-Y causal mask on the tiny benchmark model (the explicit
  mask + static-cache wiring adds nodes vs the `is_causal=1` baseline that does not run
  on CUDA for this config). This is a deliberate correctness/runnability trade-off, not a
  regression — requesting **maintainer-accept of the node-count delta** at merge.

  This increase is **temporary**. The explicit Option-Y mask exists only because ORT
  currently hard-rejects `is_causal=1` + `nonpad_kv_seqlen` (no `past_key`) for this
  static-cache shape, which forces the `is_causal=0` + explicit-mask formulation; the
  extra `num_nodes` over the maskless `is_causal=1` baseline is therefore **known and
  intended**, not a regression. Once onnx/onnx#8068 (the spec change defining bottom-right
  causal alignment for this path; implements onnx/onnx#8054) and the corresponding ORT
  follow-up land, mobius will **revert to `is_causal=1` + `nonpad_kv_seqlen` with no
  explicit mask** — which removes these nodes, clears the benchmark node-count gate, and
  unlocks Flash-attention eligibility for this configuration. The mobius-side revert is
  tracked as a follow-up ([onnxruntime/mobius#345](https://github.com/onnxruntime/mobius/issues/345); see the
  *Upstream coordination & temporary-workaround status* section above).

## Testing

- **Full fast test suite:** zero new failures vs the base-branch baseline.
- **New / updated passing tests:** mask-built-once, shared-`ir.Value` identity across
  layers, `None`-fallback parity for the non-static path, the static-cache causal-mask
  behavior tests, and the `_register_kv_cache_outputs` partial-set **fail-closed** test
  (previously asserted warn-and-proceed; now asserts `ValueError`).
- **CUDA end-to-end:** static-cache attention runs green on CUDA (MEA path).
- **Lint:** clean.

---

Closes #341 (Major-3a `_cache_utils` fail-open → fail-closed, fixed in-PR rather than
deferred). Major-3b (`apply_weights` INFO-only unmapped weights) is pre-existing on
`main` and remains out of scope.

### DIFF
--- a/.agents/skills/mobius-onnx-export-gotchas/SKILL.md
+++ b/.agents/skills/mobius-onnx-export-gotchas/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: mobius-onnx-export-gotchas
-description: Use when building/exporting ONNX models with the `mobius build` CLI (especially Phi-3 / Phi-3.5 or any model with `--execution-provider cuda` GQA fusion and/or `--static-cache`). Covers the current CLI syntax, the dtype flag values, the GQA-vs-static-cache interaction, how to verify fp16 GQA exports load in onnxruntime (the historical packed-QKV FLOAT32 load bug is fixed as of df203cc), and why fp16 GQA exports need VALUE-based weight checks (corr≈1.0 / norm), not just initializer count/dtype, to catch silently-zeroed packed-QKV weights.
+description: Use when building/exporting ONNX models with the `mobius build` CLI (especially Phi-3 / Phi-3.5 or any model with `--execution-provider cuda` GQA fusion and/or `--static-cache`). Covers the current CLI syntax, the dtype flag values, the GQA-vs-static-cache interaction, how to verify fp16 GQA exports load in onnxruntime (the historical packed-QKV FLOAT32 load bug is fixed by the fp16 GQA fold-fix), and why fp16 GQA exports need VALUE-based weight checks (corr≈1.0 / norm), not just initializer count/dtype, to catch silently-zeroed packed-QKV weights.
 ---
 
 # mobius ONNX export gotchas
@@ -32,7 +32,7 @@ op). That breaks the pattern the GQA rewrite matches, so combining
 - **ONNX-Attention + in-place cache:** `--execution-provider default --static-cache --max-seq-len N`.
 
 ## 3. FIXED: fp16 GQA export previously left packed-QKV weights as FLOAT32 → model wouldn't load
-**Status: fixed as of commit `df203cc`.** Native fp16 Phi-3.5 GQA export now loads directly in the ORT
+**Status: fixed (the fp16 GQA fold-fix).** Native fp16 Phi-3.5 GQA export now loads directly in the ORT
 CUDA EP with **no manual post-cast** (32 GroupQueryAttention nodes, all-fp16 initializers). If you are on
 that commit or later, you should not hit this — skip to the verification snippet below. The history is
 kept here because old artifacts exported before the fix still carry fp32 packed weights.
@@ -122,7 +122,7 @@ The packed-QKV initializer is `Transpose(Concat(q, k, v, axis=0))`. If the fold 
 (`FoldConcatInitializersPass` / `FoldTransposedInitializerPass`) leave the packed-Concat output dtype
 UNKNOWN / defaulted-to-fp32 while the data is fp16, the serializer **skips** it and it loads as
 **near-zero** — the weights are silently dead. (This is the §3 failure mode; the upstream fix in
-`df203cc` stamps the fp16 dtype at the fold-pass source. A post-hoc cast is NOT a fix — it re-corrupts.)
+The fp16 GQA fold-fix stamps the fp16 dtype at the fold-pass source. A post-hoc cast is NOT a fix — it re-corrupts.)
 
 ### Why count/dtype checks fail (the trap)
 The BROKEN export and the FIXED export can have the **same initializer count and the same fp16/fp32 dtype
@@ -154,7 +154,7 @@ QA's `gqa_weight_integrity_gate.py` (`--self-check --strip-audit --scan-all`, pe
 implements exactly this gate.
 
 ## 6. FIXED: GQA `present.*` KV-cache outputs declared the wrong `head_dim`
-**Status: fixed as of commit `cf6c5c4`.** A native fp16 GQA export now declares
+**Status: fixed (the GQA present-KV shape fix).** A native fp16 GQA export now declares
 `present.{i}.{key,value}` with the correct `head_dim`, symmetric to its `past_key_values.{i}.*` inputs.
 
 ### Symptom (pre-fix)
--- a/.agents/skills/mobius-onnx-export-gotchas/SKILL.md
+++ b/.agents/skills/mobius-onnx-export-gotchas/SKILL.md
@@ -1,0 +1,200 @@
+---
+name: mobius-onnx-export-gotchas
+description: Use when building/exporting ONNX models with the `mobius build` CLI (especially Phi-3 / Phi-3.5 or any model with `--execution-provider cuda` GQA fusion and/or `--static-cache`). Covers the current CLI syntax, the dtype flag values, the GQA-vs-static-cache interaction, how to verify fp16 GQA exports load in onnxruntime (the historical packed-QKV FLOAT32 load bug is fixed as of df203cc), and why fp16 GQA exports need VALUE-based weight checks (corr≈1.0 / norm), not just initializer count/dtype, to catch silently-zeroed packed-QKV weights.
+---
+
+# mobius ONNX export gotchas
+
+## 1. CLI syntax (editable repo differs from older docs)
+`mobius build` requires `--model <hf_id>` and takes the **output dir as a POSITIONAL** arg.
+There is **no `-o` flag for `build`** (`-o` exists only on `build-gguf`).
+
+```bash
+mobius build --model microsoft/Phi-3.5-mini-instruct \
+  --dtype f16 --execution-provider cuda \
+  --external-data onnx --trust-remote-code \
+  /path/to/output_dir
+```
+
+- `--dtype` choices: `f16`/`float16`, `bf16`/`bfloat16`, `f32`/`float32`. **`fp16` is INVALID.**
+- `--execution-provider` is an alias of `--ep`. `cuda` + fp16/bf16 triggers GQA fusion;
+  `default` keeps plain ONNX `Attention`.
+
+## 2. `--static-cache` is incompatible with GQA fusion
+`--static-cache` wraps each attention with `TensorScatter` (in-place KV cache for the **ONNX Attention**
+op). That breaks the pattern the GQA rewrite matches, so combining
+`--execution-provider cuda --static-cache` yields **0 GroupQueryAttention + N Attention + 2N TensorScatter**
+(mobius prints: "GQA fusion expected … but found 0 GroupQueryAttention and N Attention nodes").
+
+- **GQA model:** `--execution-provider cuda` **alone**. GQA's shared KV buffer
+  (`past_present_share_buffer`) is enabled at **runtime** via IO-binding past & present to the same
+  OrtValue — NOT via `--static-cache`.
+- **ONNX-Attention + in-place cache:** `--execution-provider default --static-cache --max-seq-len N`.
+
+## 3. FIXED: fp16 GQA export previously left packed-QKV weights as FLOAT32 → model wouldn't load
+**Status: fixed as of commit `df203cc`.** Native fp16 Phi-3.5 GQA export now loads directly in the ORT
+CUDA EP with **no manual post-cast** (32 GroupQueryAttention nodes, all-fp16 initializers). If you are on
+that commit or later, you should not hit this — skip to the verification snippet below. The history is
+kept here because old artifacts exported before the fix still carry fp32 packed weights.
+
+### Symptom (pre-fix)
+For an fp16 GQA export, a folded per-layer packed QKV weight
+(`..q_proj.weight__k_proj.weight__v_proj.weight__axis_0__concat`) was emitted as **FLOAT32**, while its
+MatMul's other input was fp16. onnxruntime then rejected the model at load on both CPU and CUDA EPs:
+
+```
+Type Error: Type parameter (T) of Optype (MatMul) bound to different types
+(tensor(float16) and tensor(float)) in node (node_MatMul_*)
+```
+
+You'd also see at save time: `The value type for shape [H, 3H] is not known. Skipping serialization`.
+
+### Root cause
+`_cast_module_dtype` casts module params to fp16, but the resulting initializer `Value`s lose their
+declared `.dtype` (it becomes `None`) while their `const_value` stays fp16. The fold passes
+`FoldConcatInitializersPass` (`src/mobius/_passes/_fold_concat.py`) and `FoldTransposedInitializerPass`
+(`src/mobius/_passes/_fold_transpose.py`) then defaulted the folded initializer's dtype to `FLOAT`,
+serializing the packed QKV / transposed weights as fp32.
+
+### The fix
+A shared helper `initializer_dtype()` (`src/mobius/_passes/_dtype_utils.py`) resolves the effective dtype
+from the declared type, **falling back to `const_value` when the type annotation was dropped** (preferring
+the data dtype and warning on stale-metadata disagreement). Both fold passes use it to stamp the correct
+dtype on the new initializer's `TensorType` and `LazyTensor`, and `FoldConcatInitializersPass` now also
+skips folding before weights are loaded (mirroring `FoldTransposedInitializerPass`). A regression test
+loads the fp16 GQA export in the ORT CPU EP to lock this in.
+
+### Verify (still worth running on any fp16 build)
+```python
+import onnx
+m = onnx.load("model.onnx", load_external_data=False)
+fp32 = [i.name for i in m.graph.initializer if i.data_type == onnx.TensorProto.FLOAT]
+print(len(fp32), "FLOAT32 initializers (should be 0 for fp16)")
+```
+
+### Convention (prevents the whole class from reappearing)
+Any pass that **materializes a new initializer** must resolve its dtype via
+`initializer_dtype()` (`src/mobius/_passes/_dtype_utils.py`), **never** `value.dtype or ir.DataType.FLOAT`.
+The bug class originates in `_cast_module_dtype` dropping a `Value`'s declared `.dtype` (→ `None`) while its
+`const_value` stays fp16; a bare `.dtype or FLOAT` fallback then silently mis-types the result as fp32. Fold
+passes (`_fold_concat.py`, `_fold_transpose.py`) already follow this; mirror it in any future
+initializer-producing pass. Siblings still reading `.dtype` directly remain exposed — a follow-up should
+grep `_passes/` for `.dtype or ir.DataType` and consider re-stamping the type in `_cast_module_dtype` to kill
+the class at source.
+
+### Salvaging a stale pre-fix artifact (only if re-exporting is not an option)
+Prefer re-exporting on the fixed code. If you must repair an old model, cast its FLOAT32 initializers to
+fp16 and re-save. **Gotcha when re-saving with external data:** if you save with `location="X.data"` and
+then rename the file, the references inside `model.onnx` still point to `X.data`. Either save directly
+with `location="model.onnx.data"`, or rewrite each initializer's `external_data` `location` entry.
+
+```python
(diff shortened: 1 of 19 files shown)


titaiwangms:
## Review synthesis (4-reviewer team: readability, code, critical, deep)

**Verdict: Strong PR, no Critical issues.** The causal-mask math and opset-24 `Attention` spec adherence were independently verified correct. The items below are about contract clarity, perf, stale docs, and test coverage — not a proven math bug.

### Major
1. **Static-cache mask assumes a compact / right-trimmed cache** (`components/_common.py`, `components/_attention.py`). The mask is purely `j <= write_indices[b] + t` and never consumes `nonpad_kv_seqlen`. It is correct **only** under the invariant `nonpad == write_index + S_q` per batch (the deep reviewer confirmed padding is subsumed under that invariant). Left/interior padding — e.g. a ragged batched prefill scattering pad tokens into low cache slots — would be treated as attendable and corrupt batched-generation logits. Suggest documenting + enforcing the compact-cache I/O contract at the serving boundary, or adding a real validity mask.
2. **Per-layer mask rebuild** (`components/_attention.py`). The mask depends only on `query` shape, `max_seq`, and shared `write_indices`, yet is reconstructed for every attention layer (O(layers × S_q × max_seq) duplicated graph/bias construction — a CUDA-graph memory/perf tax for long prompts). Suggest building it once at the task/model level or caching it in `StaticCacheState`.
3. **Stale docstrings contradict the `is_causal=0` switch** (`tasks/_causal_lm.py:64` and `:364`; the latter sits in a load-bearing "Known Limitations" block). Both still say `is_causal=1`. Update to reference the explicit offset-aware mask.
4. **MLA present/past wiring untested at the task level** (`tasks/_causal_lm.py` + `tasks/_task_test.py`). `test_present_outputs_match_past_inputs` only runs default config where `key_head_dim == value_head_dim == head_dim`, so the distinct-head-dim MLA path is unverified at the call site. Add an MLA-like config test asserting past/present KV symmetry.

### Minor
- **Misleading test name** `test_static_cache_has_no_tensorscatter_left_unmasked` actually asserts `Range`/`GreaterOrEqual` presence — nothing about TensorScatter. Rename (e.g. `test_static_cache_graph_contains_causal_mask_ops`).
- **Opaque "Option Y" label** (`tests/static_cache_decode_test.py:2340`) — internal design-discussion term with no definition in the codebase. Drop it; the surrounding prose already describes the behavior.
- **Partial-params fails open** (`tasks/_cache_utils.py`) — a partial present-shape param set warns and falls back to the known-wrong GQA inference path rather than raising. Reasonable as a documented choice, but consider raising for an API wiring bug.
- **No batch>1 value-level mask test** despite per-batch `write_indices` being the core motivation. Add one with differing `write_indices` per row.
- **`_outputs=3` on the static path** (`components/_attention.py`) materializes present outputs that are discarded; the opset-24 spec note says `present_key`/`present_value` outputs should not be combined with `nonpad_kv_seqlen` (pre-existing, not introduced here). Try `_outputs=1`.
- **`np.array_equal` on CUDA logits** (`tests/static_cache_decode_test.py`) is brittle across kernels/drivers; prefer `np.testing.assert_allclose(..., rtol, atol)` for the "unchanged" checks.

### Nit
- `stamp` → `should_stamp_shapes` (`tasks/_cache_utils.py`); replace commit hash `df203cc` in assertion messages with a human-readable phrase; `q_offsets`/`key_positions` naming asymmetry in `create_static_cache_causal_mask`.

### Praise
`initializer_dtype` helper, the `create_static_cache_causal_mask` docstring + the 16-line `is_causal` rationale comment (names the ORT guard and the Flash-vs-MEA trade-off), and the negative-control style in `static_cache_decode_test.py` (poisoned in-range vs out-of-range slots) were all called out as exemplary.

<sub>Synthesized from a 4-model review team (readability, code, critical, deep reviewers).</sub>

justinchuby:
LMK when this is ready, thanks

titaiwangms:
## Round 2 review (delta `c87d7761..c2bedd69`, code + critical reviewers)

Reviewed only the changes since round 1. **The targeted round-1 fixes are correctly and non-vacuously applied** — verified by both reviewers and by running the affected fast/build tests locally (all green, including the new MLA test and the static-cache graph-build suite under `_outputs=1`).

### ✅ Round-1 findings resolved
- **`is_causal=1` stale docstrings** (`tasks/_causal_lm.py`) — updated to `is_causal=0` + explicit-mask, including the Falcon/ALiBi "Known Limitations" entry.
- **MLA present/past wiring untested** — new `test_present_outputs_match_past_inputs_mla_distinct_head_dims` yields genuinely distinct key(16)/value(8) head dims, has a precondition assert, and exercises the stamping path. Verified passing.
- **`_outputs=3` → `_outputs=1`** (`components/_attention.py`) — correct; matches the codebase idiom and the cache still comes from the TensorScatter `updated_k/updated_v`. Static-cache graph-build tests pass.
- **Misleading test name** → `test_static_cache_graph_contains_causal_mask_ops` (now matches what it asserts).
- **"Option Y" internal label** removed from `static_cache_decode_test.py`.
- **`np.array_equal` brittleness** → "unchanged" controls now use `assert_allclose` (1e-5); negative controls correctly keep exact inequality.
- **batch>1 value-level coverage** → new `test_static_cache_decode_mask_is_per_batch_on_cuda` added.
- **Compact-cache contract** (round-1 Major) → now thoroughly documented in `create_static_cache_causal_mask` (see Major-1 below — documented, not enforced).

### Still open

**Major**
1. **Compact-cache contract is documented but not enforced** (`components/_common.py`). The mask remains purely `j <= write_indices[b] + t` with no `key_positions < nonpad_kv_seqlen[b]` term and no structural validation. Static-cache mode has no `attention_mask` input, so a batched prefill with left/ragged/interior padding scatters pad K/V into low slots the mask treats as attendable → silently corrupted logits. The new docstring narrows the contract, but the graph interface still accepts invalid input. Either fail closed / make misuse structurally impossible, add the defensive `nonpad` bound to the mask, or reject left/ragged padding at the serving boundary. *(Acceptable to defer if the serving integration provably only produces compact caches — but that guarantee should live in code, not just prose.)*
2. **Per-layer static mask rebuild** (`components/_attention.py:178`) — `create_static_cache_causal_mask(...)` is still called inside `_apply_attention`, i.e. once per decoder layer, though it depends only on batch/query/cache shape and `write_indices`. Mask construction cost and graph size scale with layer count. Build once at the task/model level and thread it through.
3. **Partial fail-open paths remain** — `_register_kv_cache_outputs` partial-param set still warns-and-proceeds, and `apply_weights()` partial-weights still only logs. Both can ship a structurally-wrong/partially-initialized model unless callers opt into strict checks. (Carried from round 1; raised again here.)

**Minor (new)**
- **The new per-batch CUDA test can pass for the wrong reason** (`static_cache_decode_test.py`). The probe slot is `3` while row 0 has `nonpad_kv_seqlen=3`, so ORT's per-batch `nonpad` bound alone masks the slot for row 0 — the "row 0 unchanged" assertion can hold even if the 4D mask incorrectly broadcast another row's frontier. `nonpad == write + 1` (decode) cannot isolate per-batch *mask* behavior. Use a multi-token probe where the tested slot is `< nonpad_kv_seqlen` for **both** rows but causally future for one and in-frontier for the other.

**Nit**
- The tolerance comment in `static_cache_decode_test.py` claims a real leak "cannot" hide under the 1e-5 band given the 50.0 poison. Not strictly guaranteed (a leaked slot's attention weight can be tiny). Soften to "very unlikely in this setup."

### Verdict
Round 2 cleanly resolves the actionable round-1 doc/test/`_outputs` items. The remaining Majors are pre-existing design trade-offs (contract enforcement, per-layer rebuild, fail-open) rather than regressions — fine to land if tracked as follow-ups, with the compact-cache enforcement being the one I'd most want addressed before broad batched-generation use.

<sub>Round-2 synthesis from code + critical reviewers, with fixes verified against a local test run.</sub>

titaiwangms:
## Review synthesis (5-reviewer team: readability, code, critical, deep, integration)

**Verdict: mergeable.** No Critical or Major blockers. The core attention-mask math was independently verified against the **ONNX opset-24 Attention reference implementation** and is derived-correct for prefill, decode, and chunked phases.

### Correctness — verified (grounded in the opset-24 reference)
- Mask math `j ≤ write_indices[b] + t` correct for prefill (triangular), decode, and chunked offsets.
- Bool polarity (`True`=attend), the 4D `[B,1,S_q,max_seq]` shape (per-batch `write_indices`), and present/past KV shape symmetry all correct.
- **Bonus:** `_outputs=3 → 1` fixes a *latent spec violation* — opset-24 forbids `nonpad_kv_seqlen` together with `present_key`/`present_value` outputs. And `is_causal=1 → 0 + explicit mask` is the semantically correct construction (the spec's upper-left causal alignment would mis-align offset decode writes), not merely an ORT workaround.

### The one "Major" — reclassified as follow-up (not a blocker)
**Incomplete rollout of present-shape stamping.** Only `_causal_lm.py` passes the 6 stamping params; `_vision_language`, `_speech_to_text`, `_multimodal`, `_tts`, `_gemma4`, `_hunyuan_vl_mot`, etc. still use the infer path. But stamping didn't exist before this PR, so those tasks are no worse than `main` — an incomplete fix, not a regression. Fine for a CausalLMTask-scoped PR; suggest a **follow-up** to extend stamping to GQA-using multimodal decoders.

### Adjudicated: unenforced compact-cache invariant
Two reviewers flagged that correctness relies on `nonpad == write_indices + S_q` with no in-graph guard. **Spec-grounded resolution:** the opset-24 *reference implementation* (`op_attention.py:154-161`) adds the `j ≥ nonpad` padding bias as an **independent additive term**, so the bound holds for any conformant kernel — not just ORT's MEA path. No code change required; documentation/contract clarity only.

### Worth addressing in-PR (Minor)
1. **Doc precision** (`_common.py`): attribute the nonpad bound to the **opset-24 reference**, not just `attention.cc`; note the poison-test verification can't isolate kernel-enforcement from the causal mask (under the compact invariant the causal `GreaterOrEqual` already masks those slots).
2. **Missing test:** FoldConcat `const_value=None` skip branch is untested.
3. **Missing test:** pin `_outputs=1` for static-cache Attention (a regression to 3 wouldn't fail current behavior tests).
4. **Export** `create_static_cache_causal_mask` in `components/__init__.py` `__all__` (now imported cross-package).

### Nits (optional)
- Rename `stamp` → `should_stamp` (`_cache_utils.py`); de-duplicate the identical `safe=True` comment in `_fold_concat.py`; trim the ~32-line inline comment / "Cache contract" docstring block that restates the function docstrings; `query_seq_source` → `seq_len_ref`.
- `attention_mask` symbolic-dim string drift (`"past_seq_len + seq_len"` vs canonical `"past_sequence_len + sequence_len"`) — pre-existing, dynamic path only.

### Follow-ups to file
- Extend present-shape stamping to other GQA decoder tasks.
- Revert to `is_causal=1` + `nonpad_kv_seqlen` (drop explicit mask) once onnx/onnx#8054 + ORT land — already acknowledged in the PR.

<sub>🤖 Synthesized from a 5-model review team (Claude, GPT, Gemini).</sub>

titaiwangms:
@justinchuby This is ready. But if we think ORT fix can get in pretty quick and static cache attention is not used at the moment, maybe we can wait my ort pr.

justinchuby:
Thanks - could you isolate bug fixes into potentially another PR so it is easier to review? I think we can probably fix the bugs first, and then assume the ort patch you referred to to produce simpler graphs here.

titaiwangms:
Closing in favor of #351, which splits out the **bug fixes** (fp16→fp32 fold widening; GQA `present.*` head_dim stamping; fail-closed `_register_kv_cache_outputs`, closes #341) into their own reviewable PR, per @justinchuby's review request.

The **Option-Y** static-cache graph workaround from this PR (`is_causal=0` + explicit causal mask + `nonpad_kv_seqlen`, forcing MEA) is **dropped, not landed** — we'll emit the simpler maskless `is_causal=1` + `nonpad_kv_seqlen` end-state (Flash-eligible) directly once onnx/onnx#8068 + microsoft/onnxruntime#28958 ship in a pinnable ORT release and mobius bumps its ORT pin. That end-state is tracked by #345.

PR #350, opened 2026-06-09
Fix GPU integration/L4/L5 test failures: TF32, transformers 5.x vision renames, gemma3 multimodal

justinchuby:
## Summary

Ran the full integration + L4 (golden) + L5 (generation) suite on GPU
(`MOBIUS_TEST_DEVICE=cuda`, 8× H200) to find regressions and long-standing
bugs. Triaged 88 failures by root cause. **PR #338 (gemma4 bidirectional
overlay + GQA-cap removal) introduced zero regressions** — every spot-checked
failure reproduces identically on the parent commit `f972184`.

This PR lands the verified, high-impact fixes. Remaining buckets (deep
numeric/decode bugs, ref-API drift, env/external) are tracked separately and
will follow.

### 1. Disable TF32 on GPU (`tests/conftest.py`)
On Ampere+/Hopper the ORT CUDA EP uses TF32 for fp32 matmuls by default, while
the PyTorch reference computes in true fp32. The ~1e-2 logit discrepancy
spuriously failed **~35** fp32 numeric-parity tests (rtol/atol 1e-3). Set
`NVIDIA_TF32_OVERRIDE=0` in conftest before any CUDA library initializes (uses
`setdefault` so users can opt back in). Verified: with the env var unset,
`gpt2` and `qwen2.5-0.5b` L4 now pass.

### 2. transformers ≥5.x flattened ViT/CLIP weight names (`src/mobius/models/vit.py`)
transformers 5.x flattened the ViT state dict to `layers.N.*` with consolidated
`attention.{q,k,v,o}_proj` / `mlp.fc1/fc2` names. The legacy rename map no
longer matched, leaving graph initializers unfilled (ORT load failure). Added
an **additive** new-naming branch (legacy 5.0–5.9 path preserved) and aligned
the in-test torch reference modules with the mobius graph param names.

### 3. gemma3 multimodal vision encoder (`src/mobius/models/gemma3.py`)
Two bugs broke the gemma-3 image-text-to-text pipeline:
- The full-VLM `preprocess_weights` prefixed vision weights but didn't rename
  the vision MLP `fc1/fc2` → `up_proj/down_proj`, so FCMLP initializers were
  never filled (ORT load failure).
- The vision encoder returned rank-3 `(batch, tokens, hidden)`, but the
  embedding sub-model declares `image_features` as rank-2 and gathers along
  axis 0. Squeeze the batch dim to honor the 2-D contract (matching the
  `PixtralVLTask` precedent and the ort-genai runtime).

With both, **gemma-3-4b-it L4 golden passes** on CUDA.

## Verification
- TF32 fix: `gpt2`, `qwen2.5-0.5b` L4 pass with env unset.
- ViT/CLIP parity tests pass.
- gemma-3-4b-it L4 golden passes; 25 gemma3 build/task unit tests pass.
- `lintrunner` clean on all changed files.

---

## Architecture-diff notes (review follow-up)

Two entries in the `arch-diff-bot` comment (`e005489 -> 2f13c7d`) deserve explanation:

### falcon: `Gelu` -> `Sigmoid + Mul`
The bloom-GELU fix changed falcon's MLP from a hardcoded `activation="gelu"` to `activation=config.hidden_act or "gelu"` so `FCMLP` can be shared with Bloom (which needs `gelu_pytorch_tanh`). For a **real** falcon config this is a no-op: `ArchitectureConfig.hidden_act` resolves to `"gelu"` via `config.activation` (HF `FalconConfig` has no `hidden_act` and defaults `activation="gelu"`). The arch-diff only showed SiLU (`Sigmoid+Mul`) because the **synthetic test config** used the generic `_base_config` default `hidden_act="silu"`. Fixed by setting `hidden_act="gelu"` in falcon's synthetic config so it matches real Falcon.

### `RotaryEmbedding: num_heads: 2 -> 4`
This is a **false positive** from the diff tool's positional node matching, *caused by* the SiLU regression above -- not a real change. Each falcon layer emits two `RotaryEmbedding` nodes: Q-rotary (`num_heads=4`) and K-rotary (`num_heads=2`, GQA `kv_heads=2`). Inserting `Sigmoid+Mul` per layer shifted all subsequent node indices by +2, so the tool aligned base's K-rotary (`num_heads=2`) against head's Q-rotary (`num_heads=4`). Per-head counts are unchanged at both base and head. Restoring `Gelu` removes the extra nodes, realigns indices, and makes this spurious entry disappear.


titaiwangms:
## Review synthesis — PR #350

Reviewed by a 5-model team (readability, correctness, adversarial, spec-adherence, cross-module integration) plus independent spot-checks. The PR's core graph math (Phi4MM NaViT position-IDs, HD mask crop, LoRA gate *selection*, Bloom/Falcon activations) was independently verified as faithful to HuggingFace. Findings are deduped and prioritized below.

### 🔴 Critical / Major

**1. Parity `_AMBIGUOUS_JACCARD = 0.9` is mathematically inconsistent with its intent** — `src/mobius/_testing/parity.py:264,273`
For two size-10 sets sharing `k` elements, `jaccard = k/(20−k)`. `≥0.9` requires `k=10` (identical sets); 9/10 overlap = `9/11 ≈ 0.818` → still **FAIL**. So the branch added to rescue "CUDA fp32 near-ties with 9/10 overlap" never fires for that case — it only fires when the top-10 sets are *identical*. The comment "at least 9 of the 10 highest tokens agree" is therefore wrong, and the `onnx_top1 in golden_top10` clause becomes redundant. Separately: as written, an identical-top-10 case where a low-ranked golden token is promoted to argmax with a *large* logit gap would be silently downgraded to AMBIGUOUS (non-FAIL), since the guard requires neither `argmax == golden_top2` nor a gap bound.
**Fix:** align the comment with the code AND tighten the guard (require `argmax == golden_top2` and/or a logit-gap bound); add a boundary test with exactly 9/10 overlap.

**2. Example scripts don't wire `vision_gate` / `speech_gate`** — `examples/phi4mm_ort_genai.py:133`, `examples/phi4mm_multimodal.py:419`
The decoder now declares both gates as **required** inputs and the embedding emits them, but the genai config maps only `inputs_embeds`, and the manual generate loop does `inputs_embeds = embed_out["inputs_embeds"]` and drops the gates → runtime "missing input" crash. The golden test infra routes them dynamically (so tests pass), but the example scripts are broken.
**Fix:** map the gates in the genai config embedding outputs + decoder inputs, and extract/feed them in `prepare_decoder_feeds`.

**3. `batch > 1` is structurally unsupported but the graphs declare symbolic batch** — `src/mobius/models/gemma3.py:~100` (`Squeeze([0])`) and Phi4MM scalar gates in `tasks/_phi4mm_multimodal.py:1215`
`Squeeze([0])` and the scalar `vision_gate`/`speech_gate` (shape `[]`) assume batch=1, but the task inputs use a symbolic batch dim. This matches HF's one-image-at-a-time runtime, so it is a real but contained latent constraint.
**Fix:** either pin batch=1 in the input contract, or handle batch explicitly (reshape to `[-1, hidden]` instead of squeezing; per-row gates `[batch, 1, 1]`).

**4. New xfails may defer real mobius bugs rather than fix them** — `tests/e2e_golden_test.py:208+` (judgment call)
`phi4mm-multi-image-audio` ("decoder-side gap, cos~0.983") and `nemotron-h` decode divergence are CPU-reproducible *mobius* behavior, not external flakes. The Gemma4 CUDA-only xfails are attributed to an ORT CUDA EP bug but carry no linked issue or minimized reproducer. Per the repo's "root-cause before remedy" convention this is worth confirming; the reasoning text is unusually thorough and the PR explicitly tracks these separately, so this is a maintainer judgment call, not a clear defect. Confirming the CUDA-only attribution needs a run (CPU pass + CUDA fail on the same exported graph).

### 🟡 Minor
- **Div-by-zero** if a crop is fully padded (`nb_h`/`nb_w` == 0) — `src/mobius/models/phi.py:~707`. Latent (Phi4MM never emits it); clamp `≥ 1` before the `Div`.
- **`self._lora_gates` not cleared** before repopulate in `_Phi4MMDecoderModel.forward()` — stale-handle risk if the module is reused; `clear()` first.
- **No unit tests** for the new `vit.py` transformers≥5.x rename branch (`layers.N.*`).
- **bf16 Gemma4 test** dropped the elementwise tolerance for cosine>0.999 + argmax only — could hide non-last-position corruption; keep a finite max/mean-diff ceiling too.
- **Readability:** stale module docstring + dead `else` branch + misleading `inputs_embeds` binding name in `_phi4mm_multimodal.py`; the "populated in forward()" comment on `self._lora_gates` should clarify this is build-time (not inference-time) mutation; `col0`/`nb_h` could use clearer names.

### ✅ Verified correct (non-findings)
NaViT bucketize math (`floor(r·P/nb_h)·P + floor(c·P/nb_w)`, with no-padding collapse to the standard raster); HD stride-2 mask crop matches HF corner-sampling; LoRA gate *selection* matches HF `InputMode`/`set_lora_adapter` (VISION_SPEECH → vision); Bloom `gelu_pytorch_tanh` / Falcon `hidden_act` fallback; `grid_thw` → `image_grid_thw` rename consistent with the 3-model graph; new optional `attention_mask` / `gate_holder` params are backward-compatible; YolosConfig rectangular sizing; global TF32-disable is safe.

**Highest priority:** #1 (Jaccard) and #2 (example gate wiring) are concrete, in-scope bugs. #3 and #4 are design/scope judgment calls.

<sub>🤖 Generated with a multi-model review team (Claude + GPT + Gemini).</sub>

justinchuby:
Added gemma4 KV-shared attention fix (commit f93778e).

**Root cause:** the non-GQA Attention fallback for Gemma4 KV-shared layers fed the full shared K/V (no past) with `is_causal=1` *and* a float causal bias from `create_attention_bias`. Since the bias already encodes full bottom-right causality, `is_causal=1` double-applied the op's built-in causal mask — and CPU (bottom-right) vs CUDA (upper-left) disagree on its alignment for `q_len < kv_len`, so decode diverged on CUDA after the first token.

**Fix:** pass `is_causal=0` in this fallback (bias-only masking, EP-agnostic). gemma-4-e2b/e4b L4+L5 (text/image/speech) now pass on **both CPU and CUDA**; removed the temporary CUDA-only L5 xfails.

Filed the underlying ORT CPU/CUDA inconsistency as microsoft/onnxruntime#29020.

justinchuby:
Addressed review feedback (commit da340dc):

**`vit.py` — transformers 5.x prefix stripping (@Copilot):** Confirmed real. transformers 5.10 `ViTForImageClassification` emits `vit.layers.N.*`, and `_rename_vit_weight` returned `None` for those (silently dropping layer weights) because the prefix-strip allowlist omitted `layers.`. Fixed by adding `layers.` to the allowlist; verified `vit.layers.N`, `vision_model.layers.N`, `dinov2.layers.N` and bare `layers.N` all map correctly now.

**`phi4mm_integration_test.py:64` — '_PHI4MM_COMPAT_APPLIED is not used' (@github-code-quality):** False positive. The module global is a one-shot guard: written at line 64 and read at line 53 (`if _PHI4MM_COMPAT_APPLIED: return`) via the `global` declaration at line 52. Leaving as-is.

Also added a gemma4 graph unit test (`test_gemma4_kv_shared_fallback_attention_is_causal_zero`) locking in the KV-shared `is_causal=0` invariant from the prior commit.

justinchuby:
@titaiwangms thanks for the thorough multi-model review — addressed the concrete findings:

**Fixed (commits `fe33780`, `ebb5354`, `c21b052`):**
- 🔴 **# 1 Jaccard math** — `compare_golden`'s AMBIGUOUS gate used a `jaccard >= 0.9` ratio, which for two size-10 sets requires *identical* sets (9/10 overlap → jaccard 9/11 = 0.818), so the "9 of 10 agree" intent never fired. Switched to count-based `overlap >= 9` **and** additionally require the golden argmax to stay in the ONNX top-2 — so an identical top-10 with a low token promoted to #1 (large gap) is no longer masked. Added boundary tests (9/10 swap → AMBIGUOUS; buried golden top1 → FAIL).
- 🔴 **# 2 example gate wiring** — `examples/phi4mm_ort_genai.py` and `examples/phi4mm_multimodal.py` now wire `vision_gate`/`speech_gate` (genai_config embedding outputs + decoder inputs; extracted/fed in the manual decode loop). The CLI auto-export path already introspects these, so only the hand-written examples were affected.
- 🟡 `_Phi4MMDecoderModel.forward` now `self._lora_gates.clear()`s before repopulating (no stale-gate leak across calls).
- 🟡 NaViT position-id: clamp `nb_h`/`nb_w` divisors to ≥1 (fully-padded crop div-by-zero; masked to id 0 anyway).
- 🟡 Added `vit_test.py` unit tests for the transformers ≥5.x flattened `layers.N.*` rename branch.
- 🟡 bf16 gemma4 prefill test: kept cosine+argmax as the primary gate but added a loose finite max/mean-abs ceiling to catch gross NaN-free divergence.

**Already resolved earlier in this PR:**
- 🔴 **# 4 gemma4 xfail** — root-caused + fixed (`is_causal=0` in the shared-KV fallback, bias-only causality), CPU+CUDA L4/L5 pass, ORT inconsistency filed as microsoft/onnxruntime#29020. The nemotron-h and phi4mm-multi-image-audio xfails were each independently root-caused as genuine (hybrid Mamba2 decode drift; decoder-side near-tie flip) — not deferred GPU regressions.

**Deferred / judgment calls (want your input):**
- 🔴 **# 3 batch>1** — symbolic batch dim is declared but several paths assume batch=1 (gemma3 `Squeeze([0])`, phi4mm scalar gates). This is a real scope gap but a larger design change; I'd prefer to track it separately rather than expand this PR. Open to either (a) documenting batch=1 as a precondition, or (b) a follow-up issue.
- Readability nits (stale docstrings, dead `else`, naming) — happy to do a cleanup pass if you want it in this PR.

justinchuby:
**Re: finding #3 (batch>1) — addressed, and it surfaced a real bug.**

Investigating this revealed that text decoders did **not** actually support batch>1 (contrary to the declared symbolic batch dim): `create_padding_mask`/`create_sliding_window_mask` returned a 3-D `(batch, q, total)` bool mask, which the ONNX Attention op right-aligns onto `(batch, q_num_heads, q_seq, kv_seq)` — so the batch axis was read as `q_num_heads`. Harmless for batch=1 (broadcasts as heads=1), but ORT rejects batch>1: *"attn_mask ... not compatible with q_num_heads"*.

**Fix (commit `2f13c7d`):** both maskers now emit a 4-D `(batch, 1, q, total)` mask with an explicit singleton head dim (`create_attention_bias` already did this). batch=1 output is numerically identical (extra unit dim only). Verified `batch=2` prefill with ragged per-row padding runs and yields independent rows for qwen2/llama/mistral/gemma2 (plain, GQA, sliding-window). Added rank + per-row-independence unit tests and an ORT `batch=2` execution test.

**Multimodal:** the VLM *decoder* now genuinely supports batch>1, but the multimodal *splits* remain batch=1 by design — gemma3 vision `Squeeze([0])`, phi4mm scalar `vision_gate`/`speech_gate`, and the single flattened feature stream. True multimodal batch>1 (per-row modalities, ragged features, per-row LoRA gates) is a larger effort that also has to reconcile with the onnxruntime-genai MultiModal pipeline (one sequence at a time). I documented these preconditions explicitly (`Phi4MMMultiModalTask` docstring, gemma3 vision comment) and filed #354 to track it.

Issue #345, opened 2026-06-08
Emit maskless is_causal=1 + nonpad_kv_seqlen static-cache graph once upstream lands


titaiwangms:
## Goal

Once the upstream bottom-right `is_causal` support lands and is pinnable, emit the **maskless** static-cache attention graph directly:

```
Attention(..., is_causal=1, nonpad_kv_seqlen=<seqlens_k>)   # no explicit attn_mask, no past_key
```

This is the **end-state** for the static-cache (`--static-cache`, ONNX `Attention` + `TensorScatter` in-place KV-cache) path. It is strictly better than the explicit-mask form on two axes: **fewer nodes** *and* **Flash-eligible** (an explicit `attn_mask` forces the Memory-Efficient Attention kernel; the maskless form routes to Flash for fp16/bf16).

## Background

mobius #340 originally proposed **"Option-Y"** as a bridge: `is_causal=0` + an explicit offset-aware causal mask + `nonpad_kv_seqlen`, which forces ORT to the MEA kernel — the only currently-runnable static-cache config, because today's opset-24 `Attention` CUDA kernel rejects `is_causal=1` + `nonpad_kv_seqlen` (no `past_key`, `S_q != total_kv`) with `NOT_IMPLEMENTED`.

**Option-Y was dropped, not landed** (the bug-fix half of #340 shipped separately as #351). So there is **nothing to revert** — this issue now tracks emitting the maskless end-state directly.

## Dependency chain

The blocker is not "PR merged" but "a *released* ORT containing the kernel fix, pinned by mobius":

1. onnx/onnx#8068 — spec/errata: bottom-right `is_causal` + `nonpad_kv_seqlen` (external KV cache)
2. microsoft/onnxruntime#28904 / microsoft/onnxruntime#28958 — CPU/CUDA `Attention` kernel implementing it (maskless `is_causal=1` + `nonpad_kv_seqlen` becomes Flash-eligible)
3. an ORT **release** cut that contains the above
4. mobius bumps its ORT pin past that release (the node-count gate baselines on the pinned ORT)

## Action (when the pin can bump)

- Switch the static-cache export from the explicit-mask form to maskless `is_causal=1` + `nonpad_kv_seqlen`.
- Confirm CUDA routes to Flash (fp16/bf16) and the node count drops.
- Revisit the related independent gaps: numerical-parity CI coverage (#329) and the bias-decoder external-cache path (#349).

PR #351, opened 2026-06-10
Fix fp16/GQA static-cache export correctness (split from #340)

### LINKED ISSUE #341: Static-cache: add opt-in strict mode for partial present-shape params in _register_kv_cache_outputs

titaiwangms:
Splits the **bug-fix** half of #340 into its own reviewable, mergeable PR, per @justinchuby's review request ("could you isolate bug fixes into potentially another PR").

## Fixes

1. **fp16 GQA export emitting fp32 packed weights.** When building fp16 models, `_cast_module_dtype` casts params to fp16 but the folded initializer `Value`s lost their declared `.dtype` (None) while `const_value` stayed fp16. `FoldConcatInitializersPass` / `FoldTransposedInitializerPass` then defaulted the packed/transposed initializer to `FLOAT`, serializing fp32 weights and making ORT reject the model with a fp16/fp32 `MatMul` type-parameter error on both CPU and CUDA EPs. New shared helper `mobius._passes._dtype_utils.initializer_dtype()` resolves the effective dtype from `const_value` when the type annotation was dropped. Also strips dead pre-pack weights via `graph.remove(node, safe=True)` so DCE drops the orphaned q/k/v_proj source initializers.

2. **GQA `present.*` head_dim mis-declaration.** `_register_kv_cache_outputs` now stamps explicit `present.*` KV-cache output shapes/dtypes for GQA instead of relying on the known-wrong shape-inference path.

3. **Fail-closed `_register_kv_cache_outputs`** (closes #341). A partial set of present-shape parameters now **raises `ValueError`** (naming provided + missing params) instead of logging a warning and shipping a structurally-wrong model. All-six (stamp) or none (infer) are unaffected.

Includes regression tests for all three (155 tests; an e2e ORT CPU-EP load test reproduces the original fp16/fp32 `MatMul` failure without the fix).

## Scope note

The **Option-Y** static-cache graph workaround from #340 (`is_causal=0` + explicit causal mask + `nonpad_kv_seqlen`, forcing MEA) is **dropped, not landed**. The maskless `is_causal=1` + `nonpad_kv_seqlen` end-state (Flash-eligible, fewer nodes) will be emitted directly once onnx/onnx#8068 + microsoft/onnxruntime#28958 ship in a pinnable ORT release and mobius bumps its ORT pin — tracked by #345.

Supersedes the bug-fix portion of #340.


src/mobius/_passes/_dtype_utils_test.py
```diff
@@ -0,0 +1,49 @@
...
+
+import logging
+
+import numpy as np
+import onnx_ir as ir
+
+from mobius._passes._dtype_utils import initializer_dtype
+
+
+def _value(dtype: ir.DataType | None, const: np.ndarray | None) -> ir.Value:
```
justinchuby, line 15: Use ir.value()?

src/mobius/_passes/_fold_transpose.py
```diff
@@ -107,6 +109,11 @@ def call(self, model: ir.Model) -> ir.passes.PassResult:
 
                 new_name = f"{inp.name}_t"
 
+                # Resolve the dtype from the declared type or, when it was
+                # dropped during graph building, from the loaded const_value, so
```
justinchuby, line 113: Which process will drop it? Can we fix that process too?

src/mobius/_passes/_dtype_utils.py
```diff
@@ -0,0 +1,55 @@
...
+    folded initializers consistent with the weights they are derived from.
+
+    Returns ``None`` only when neither the declared type nor ``const_value`` is
+    available; callers decide on a final fallback.
+    """
+    declared = value.dtype
+    const_dtype = value.const_value.dtype if value.const_value is not None else None
+
+    if declared is not None and const_dtype is not None and declared != const_dtype:
+        logger.warning(
```
justinchuby: Can we raise in this case? This way we can catch and fix the root cause.

justinchuby (approved):

titaiwangms:
Thanks @justinchuby for the review and approval — the suggestions all push toward the root
cause, which is the right instinct. Follow-up on each below (one turned out to be larger than
this PR's scope, so we're deferring its source fix).

**P1 — raise on a declared-vs-`const_value` dtype contradiction (`_dtype_utils.py`).** Agreed,
and this fits the fail-closed posture of the rest of the PR. We now **raise** when a declared
dtype genuinely contradicts the serialized `const_value` dtype — that's a should-never-happen
state we'd rather surface than paper over. We've deliberately kept the **`declared is None`**
branch non-raising: that fallback (resolve from `const_value`) is the core fp16-recovery path
this PR exists for — it's the legitimate "graph building dropped the declared type" case, not
a contradiction. So: raise on the contradiction, fall back on the missing-declaration case.

**P2 — which process drops the dtype (`_fold_transpose.py`).** Good question — we traced it,
and the answer turned out to be more involved than a single cast site. The cast site itself
(`_cast_module_dtype`) already stamps the declared dtype, so it isn't the drop point. The
declared dtype is actually lost in the **GQA PackQKV rewrite**
(`_group_query_attention.py`): the `Concat`/`Transpose` intermediates it introduces carry no
declared dtype, which is what the fold step then has to recover.

That rewrite is outside the scope of this bug-fix PR, spans multiple call sites, and we
can't fully validate a change to it in our current environment, so we'd prefer not to land a
blind cross-scope edit here. We've opened a follow-up (mobius#355) to stamp the declared
dtype at the source in that rewrite. In the meantime the symptom is fully covered by this
PR's two-layer defense — the fold-pass dtype stamping plus the `initializer_dtype()`
`const_value` fallback — which keeps the fold path robust regardless of which upstream pass
drops the type, as exercised by the `_fold_dtype_e2e` tests. See mobius#355 for tracking.

**P3 — test fixtures (`_dtype_utils_test.py`).** Applied — switched the fixtures to build via
the `ir.Value` constructor instead of the `ir.val()` factory, since the factory validates and
refuses these intentionally-degenerate fixtures (a dropped declared type alongside a
`const_value`, and a declared type that contradicts the `const_value` dtype). The P1
contradiction path is now exercised by `test_raises_on_dtype_contradiction`.

**On the automated "PR description describes Olive vision metrics" comment.** Thanks — we
double-checked, and the current PR description is accurate: it covers the fp16/GQA static-cache
export fixes (the fp16→fp32 fold-widening fix, the GQA `present.*` `head_dim` stamping, and the
fail-closed KV-cache registration), notes that the Option-Y graph workaround was dropped, and
points to mobius#345 for the maskless end-state. There's no Olive content in it — the bot looks
to have read a transient/earlier state. No change needed there.

Thanks again for the careful review.

PR #358, opened 2026-06-16
Fix Qwen2.5/3-VL vision PackedMultiHeadAttention crash on CUDA

justinchuby:
## Problem

Building **Qwen2.5-VL-3B** with Mobius and evaluating on **CUDA** crashes at the first windowed vision block:

```
[E ... PackedMultiHeadAttention node 'vision_encoder/visual/blocks.0/attn/PackedMultiHeadAttention_node_214']
Status Message: Input 'cumulative_sequence_length' should have 1 dimension with size equal to batch_size + 1
```

## Root cause

The EP-gated packed path (`_emit_packed_mha`, used when the EP advertises `supports_packed_multi_head_attention`, i.e. CUDA / trt-rtx) built `token_offset` as shape **`(1, N)`**, declaring `batch_size = 1` to `com.microsoft::PackedMultiHeadAttention`.

ORT derives `batch_size` from `token_offset.shape[0]` and requires `cumulative_sequence_length` (our `cu_seqlens`) to have length `batch_size + 1` (`contrib_ops/cuda/bert/packed_multihead_attention.cc`). Whenever `cu_seqlens` enumerates more than one sub-sequence:

- **windowed** vision blocks (`cu_window_seqlens`, length `num_windows + 1`), or
- multi-frame **video** / multi-image **full-attention** blocks (length `T + 1`),

`num_sub_seqs + 1 > 2`, so the kernel rejects the input and the run crashes. The Copilot diagnosis in the issue was accurate.

Even in the non-crashing single-sub-sequence case, `batch_size = 1` made the kernel compute **full** attention over all `N` tokens instead of the intended **block-diagonal** attention — so the path was doubly broken.

`_qwen3_vl_vision.py` had the identical copy-pasted bug (latent crash for multi-image / video).

## Fix

Build `token_offset` with shape `(num_sub_seqs, max_seq_len)` following ORT's `GetPaddingOffset` convention — valid padded-layout token indices (`b * max_seq_len + s`) in packed order, then padding-slot indices. This matches the already-correct `BlockDiagonalToPackedMHA` rewrite rule, extracted into a shared `build_packed_token_offset` helper in `components/_common.py` and reused by both vision encoders.

Also stop unconditionally down-casting q/k/v to float16: the PackedMHA CUDA kernel supports **both float32 and float16**, so only **bfloat16** builds (unsupported) are cast to f16; f32/f16 stay native to preserve precision.

## Verification

- **CUDA parity:** with the fix, the packed path matches the standard block-diagonal attention path (CPU) on a 4-window grid (`grid_thw=[1,12,12]`) that crashed before — **cosine 1.0, max abs diff ~7e-8**.
- New exact-value unit tests for `build_packed_token_offset` against ORT's own reference data (e.g. `cu=[0,1,3] -> [[0,2],[3,1]]`, including padding indices `>= token_count`, single-subsequence identity, uniform windows).
- Full non-integration suite green (1318 passed); existing `qwen2_5_vl` / `qwen3_vl` graph tests and packed-attention rewrite tests unaffected.

### DIFF
--- a/src/mobius/components/_common.py
+++ b/src/mobius/components/_common.py
@@ -292,6 +292,15 @@ def build_packed_token_offset(op: OpBuilder, cu_seqlens) -> ir.Value:
     Returns:
         ``(batch_size, max_seq_len)`` INT32 ``token_offset`` tensor.
     """
+    # Index/axis tensors are materialised as explicit Constant nodes (rather
+    # than relying on Python-list auto-conversion) so this helper works under
+    # both the component OpBuilder and the onnxscript rewriter op context.
+    axes_0 = op.Constant(value_ints=[0])
+    axes_1 = op.Constant(value_ints=[1])
+    neg_one = op.Constant(value_ints=[-1])
+    start_1 = op.Constant(value_ints=[1])
+    int_max = op.Constant(value_ints=[INT64_MAX])
+
     cu_seqlens_i32 = op.Cast(cu_seqlens, to=ir.DataType.INT32)
 
     # batch_size = len(cu_seqlens) - 1  (number of packed sub-sequences).
@@ -301,10 +310,10 @@ def build_packed_token_offset(op: OpBuilder, cu_seqlens) -> ir.Value:
     )
 
     # Per-sub-sequence lengths and the padded sequence dimension.
-    starts = op.Slice(cu_seqlens_i32, [0], [-1], [0])  # cu[:-1]
-    ends = op.Slice(cu_seqlens_i32, [1], [INT64_MAX], [0])  # cu[1:]
+    starts = op.Slice(cu_seqlens_i32, axes_0, neg_one, axes_0)  # cu[:-1]
+    ends = op.Slice(cu_seqlens_i32, start_1, int_max, axes_0)  # cu[1:]
     lengths = op.Sub(ends, starts)  # (batch_size,)
-    max_len = op.Squeeze(op.ReduceMax(lengths), [0])  # scalar INT32
+    max_len = op.Squeeze(op.ReduceMax(lengths), axes_0)  # scalar INT32
 
     # Padded position grid: pos[b, s] = b * max_len + s.
     zero_i32 = op.Cast(op.Constant(value_int=0), to=ir.DataType.INT32)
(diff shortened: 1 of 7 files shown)


titaiwangms:
## Review team synthesis 🛬

Five reviewers (readability, code, critical, deep, integration) reviewed this PR in parallel. **Verdict: correct, ready to merge.** No Critical or Major correctness blockers. The findings below are in-scope polish + out-of-scope follow-ups.

### ✅ Correctness — verified, grounded in ORT source
- **`build_packed_token_offset` reproduces ORT's `getTokenOffset` exactly.** Grounded in `contrib_ops/cuda/bert/bert_padding.cu` (valid offsets `b*seq_len + s` in sub-seq-major order, then padding slots). The row-major dual-`Compress` (valid then padding) matches ORT's `for i … for j` loop nesting, with `sequence_length = max(lengths)`. Verified executably: **20,000 random `cu_seqlens` → 0 mismatches**; ORT's own example and all four PR test vectors reproduced byte-for-byte (incl. the ragged `cu=[0,2,5] → [[0,1,3],[4,5,2]]`).
- **Shape invariant correct.** `packed_multihead_attention.cc` derives `batch_size = token_offset.shape[0]` and rejects `cu_seqlens` length ≠ `batch_size+1`. The new `(num_sub_seqs, max_seq_len)` shape is exactly what the kernel requires — fixing the old `(1, N)` crash.
- **Parity with the existing reference.** The helper is line-for-line the same algorithm as `BlockDiagonalToPackedMHA` in `rewrite_rules/_packed_attention.py` (the already-correct path) — genuine centralization, no divergence.
- **dtype change is correct.** The kernel registers only `float` and `MLFloat16` (no BFloat16). So f32/f16 native passthrough is valid and bf16 *must* be cast — both halves grounded in the kernel registration table.
- **Qwen3 docstring fix is real, not masking a bug.** `v` traces to `Split(qkv, num_outputs=3)` → `(total_seq, hidden_size)`; the old `3*hidden_size` comment was wrong. Good catch.

### Minor (in-scope, optional)
1. **dtype enum consistency** — the conditional cast block uses magic ints `to=10  # FLOAT16` (and `key`/`value` casts in `_qwen3_vl_vision.py` carry no comment), while the rest of the codebase — including the new helper three lines up — uses `ir.DataType.FLOAT16`. Prefer the named enum.
2. **Test gaps** — all 4 new tests feed `int64`, but the helper documents `INT32 or INT64`; add one INT32-input case. No graph test asserts the dtype-gating wiring (bf16 inserts Cast-to-f16; f32/f16 pass through).
3. **Naming nits** — helper's `cu_i32` vs callers' `cu_seqlens_int32`; `valid`/`padding` read as booleans but hold indices (`valid_indices`/`padding_indices` per the docstring's own vocabulary); `pos_shape` → `pos_matrix_shape`.

### Follow-up suggestions (out of scope for this PR)
- **Unify the duplication**: `rewrite_rules/_packed_attention.py:172-202` still inlines the identical token_offset logic. Now that the shared helper exists, the rewrite rule could call it (would need exporting from `components/__init__.py`) to kill the drift risk. Both reviewers who flagged this agree it's not blocking.
- **Trust-boundary hardening** (open question, not introduced by this PR): the native packed path assumes `cu_seqlens` is monotonic and `cu_seqlens[-1] == token_count`. ORT validates only token_offset rank and cu length, so a malformed externally-derived `grid_thw` could fault on CUDA rather than fail-closed. Pre-existing property of the packed path; worth a defensive check + negative test someday.
- **bf16 guard is build-dtype-keyed** (deep reviewer Question): gating on `get_build_dtype()` rather than the tensor's actual dtype is correct for all current callers, but is an implicit invariant — keying off the tensor dtype would make it self-contained.

Nice work — the math is clearly derived rather than copy-pasted, and the docstring + worked example are excellent.

<sub>🤖 Synthesized from a 5-agent review fan-out (readability / code / critical / deep / integration).</sub>

PR #359, opened 2026-06-17
Add NeMo .nemo integration + FastConformer-RNNT streaming ASR model

justinchuby:
## Summary

Adds support for **`nvidia/nemotron-speech-streaming-en-0.6b`** (a FastConformer-CacheAware-RNNT streaming ASR model distributed as a NeMo `.nemo` archive) in two layers:

1. **New `.nemo` integration** (`src/mobius/integrations/nemo/`) — analogous to the existing GGUF integration. An archive reader (`NeMoArchive`), a config mapping (`model_config.yaml` → `ArchitectureConfig`), and a `build_from_nemo` pipeline. Resolves local paths and HuggingFace Hub references.

2. **FastConformer-RNNT model** (`models/nemo_rnnt.py` + `tasks/_rnnt.py`) — emitted as three ONNX sub-models (encoder / prediction decoder / joint) via a new `RNNTTask`.

## Architecture

- **Encoder:** 8× causal `dw_striding` conv subsampling → 24 Conformer layers (relative-position MHA with `rel_shift`, FF/attn/conv/FF macaron structure, layer-norm conv module). Positional encodings built dynamically in-graph.
- **Prediction:** embedding + 2-layer LSTM with PyTorch→ONNX gate reordering; zero start-of-sequence embedding (NeMo `add_sos=True`).
- **Joint:** enc/pred projections → ReLU → vocab logits → log-softmax.

## Validation

| Level | Coverage | Result |
|-------|----------|--------|
| L1 | Graph build + I/O shapes (random weights) | 7 pass |
| L4 | Parity vs committed NeMo golden | encoder ~5e-7, decoder ~5e-7, joint ~2.7e-5 |
| L5 | Greedy decode + incremental-vs-one-shot decoder state | pass |

Parity validated against `nemo_toolkit` 2.7.3 reference.

## Contract / limitations

- **Offline full-context only** — the encoder consumes the full feature sequence (causal convs, no cache-aware streaming state); chunked streaming is out of scope for this export.
- **batch=1 / equal-length** — no length input/output yet.
- **fp32 only** — the builder rejects non-fp32 (graph emits fp32 constants); unsupported NeMo encoder variants (non-`dw_striding`/`rel_pos`/`layer_norm`, `xscaling`) are rejected with a clear error.


justinchuby:
### Follow-up: review-gap fixes (commit e7c35d4)

- **Public API + CLI parity with GGUF**: `build_from_nemo` is now exported from top-level `mobius`, and a `mobius build-nemo <path-or-hf-repo>` subcommand was added (dtype/EP/output options).
- **Reproducible & pinned golden**: integration test pins the HF commit SHA; the committed `.npz` now carries a `meta` blob (model id, revision, NeMo version, dtype, seed, token/SOS/blank ids); added `scripts/generate_nemo_rnnt_golden.py` to regenerate it.
- **Cleanups**: import shared components from the public `mobius.components` API; removed a dead `TYPE_CHECKING` block.

### Quality-checklist scope (explicit N/A waivers)

These standard items don't apply to this model and are intentionally skipped:

- **L2 YAML case / L3 synthetic parity**: the source artifact is a NeMo `.nemo` archive, not a Transformers HF config/model, so the HF-config-driven YAML/synthetic-parity harness can't load it. Coverage is provided by specialized **L1** unit tests + custom **L4** (NeMo golden parity) and **L5** (RNN-T greedy decode + incremental-vs-one-shot decoder state) tests.
- **ORT GenAI / Foundry Local**: RNN-T isn't part of the ORT GenAI decoder pipeline (it needs a custom transducer decode loop), so genai_config / Foundry smoke tests are N/A. The greedy decode contract is documented in `tasks/_rnnt.py`.
- **Multi-dtype (fp16/bf16)**: the graph emits fp32 constants (positional encodings, masks, scales); the builder explicitly rejects non-fp32 export rather than silently mismatching. fp16/bf16 support is a documented follow-up.

justinchuby:
### Cache-aware streaming encoder added (commit c07dda4)

The third follow-up is done. The RNNT package now also exports an
`encoder_streaming` model for chunk-by-chunk inference, alongside the offline
full-context `encoder`.

**Contract:** inputs `audio_signal` (B,feat,T_chunk), `length`, and NeMo's
per-layer caches `cache_last_channel` (L,B,70,d), `cache_last_time` (L,B,d,8),
`cache_last_channel_len` (B,); outputs the chunk encoding + the three updated
caches. Targets a single stream of equal-length chunks (batch=1 / homogeneous),
matching NeMo's cache-aware streaming design; ragged batches use the offline
`encoder`.

**Implementation:** streaming threaded through the components via optional cache
kwargs on `forward()` (non-square rel-pos attention with K/V = concat(cache,
chunk); causal conv left-context from `cache_last_time`; `drop_extra_pre_encoded`
handling; chunk-aligned offset mask; per-layer cache stacking). The streaming
graph is built from a deep copy of the module so its parameters realize as
independent initializers with identical names.

**Validation:** new reproducible streaming golden generator + compact golden.
L1 unit test (I/O + cache growth) and an L4 chunk-chained parity test vs NeMo:
`enc_out` matches to ~1e-6 over two chained chunks and cache lengths are exact.
Full non-integration suite (2908) + the 4 nemo integration tests + lint all pass.

All three documented follow-ups (fp16/bf16, ragged-batch length masking,
streaming) are now complete.

titaiwangms:
## Review — multi-model team review (readability, correctness, adversarial, spec-adherence, integration)

Overall this is a high-quality PR. The FastConformer-RNNT math was independently verified against the NeMo reference source (rel-pos decomposition via fused `op.Attention`, `rel_shift` incl. the rectangular streaming case, positional encoding, Macaron/conv ordering, subsampling length rule, LSTM gate reorder, joint + blank bookkeeping all check out). The clever `op.Attention` decomposition (folding `pos_bias_u` into Q, pre-scaling `matrix_bd` into the post-scale bias domain) is mathematically equivalent to NeMo. The `.nemo` reader is safe (`tar.extractfile` avoids path traversal, `yaml.safe_load`, `torch.load(weights_only=True)`). **No Critical issues.**

### Major

**1. Streaming cache length grows by physical chunk frames (`tq`), not valid frames (`length_sub`)** — `src/mobius/models/nemo_rnnt.py:830-833` *(flagged independently by two reviewers)*
```python
new_len = op.Min(op.Add(cache_last_channel_len, op.Squeeze(tq)), ...)
```
`length_sub` (the per-sample valid post-drop length) is computed but unused for the cache-length update. For a final/short chunk or `length=0`, this over-increments `cache_last_channel_len`, so the next step's `_att_bias_streaming` treats invalid cached frames as valid keys → cross-chunk state corruption. Since `length` is an exposed input, callers reasonably expect it honored. Suggest `op.Add(cache_last_channel_len, length_sub)`, and add a streaming test with `length < chunk_T` (the current test uses only full-length chunks and won't catch this).

**2. `_nemo_model_type` is monkey-patched onto a dataclass instead of using the native `model_type` field** — `src/mobius/integrations/nemo/_config_mapping.py:141`, `_builder.py:70-79` *(flagged independently by two reviewers)*
`ArchitectureConfig` already declares `model_type: str | None = None` (`_configs/_base.py:497`). Stashing `config._nemo_model_type` creates a dynamic attribute invisible to type checkers / `asdict` / serialization; worse, `dataclasses.replace(config, dtype=...)` in `_builder.py` silently drops it, which is exactly why it must be re-stashed on the next line — a latent bug for the next `replace()` caller. Suggest passing `model_type=...` natively into `ArchitectureConfig(...)` and reading `config.model_type`.

**3. `_validate_encoder` does not guard `use_bias`; a NeMo-default config builds silently-wrong** — `src/mobius/integrations/nemo/_config_mapping.py` (`_validate_encoder`)
The conformer stack hardcodes `bias=False` everywhere (FF, q/k/v/out, conv modules), but NeMo defaults `use_bias=True` for all of these (`conformer_modules.py:79,255,277-317`, `multi_head_attention.py:72,96-99`; `linear_pos` is correctly always bias-free). Because `apply_weights` is non-strict (unmatched keys → warning only, `_weight_loading.py:116-121`), a `.nemo` trained with the default `use_bias=True` builds successfully while silently dropping every bias → wrong output. This is the exact "loads weights but silently wrong" failure the validator's own docstring exists to prevent. The supported `nemotron-speech-streaming-en-0.6b` happens to use `use_bias=False` (hence the clean ~5e-7 parity), so this is latent. Suggest adding `"use_bias": (False, bool(enc.get("use_bias", True)))` to the `checks` dict.

**4. No `att_context_style` guard; `att_context_size[1] == -1` divides by zero** — `src/mobius/models/nemo_rnnt.py` (`FastConformerEncoder.__init__`, `_att_bias`)
The masks always implement NeMo's `chunked_limited` rule (verified correct for that style), but `att_context_style == "regular"`/`chunked_limited_with_rc` models pass validation yet get the wrong mask. Also `self._chunk_size = right + 1` → `right == -1` (NeMo's "unlimited right" sentinel, special-cased at `conformer_encoder.py:826-828`) yields `chunk_size = 0` → `op.Div(..., 0)` in `_att_bias`. (`left == -1` *is* handled correctly.) Suggest validating `att_context_style == "chunked_limited"` and `att_context[1] >= 0`. Related: `cache_size == 0` (from `att_context_size[0] == 0`) makes `op.Slice(kv, starts=[-cache_len], ...)` with `-0` return the whole sequence instead of an empty cache — another reason to guard these config edges.

### Minor

- **genai_config log-eps inconsistency** — `_genai_config.py:37,370`: `_LOG_EPS = 2**-24` (NeMo-compatible) vs `log_zero_guard_value = 1e-10` in the written `audio_processor_config`. For low-energy mel bins these differ materially; use a single source of truth and add a test asserting the written value equals `_LOG_EPS`.
- **Unpinned VAD download** — `_genai_config.py` (`_download_silero_vad`): `hf_hub_download(_SILERO_REPO, _SILERO_FILE)` has no `revision`/hash pin, so `include_vad=True` bundles ship a mutable external model. Pin a revision (and/or default `include_vad=False`).
- **CLI `--task` not wired** — `build_from_nemo` exposes `task: str | None`, but `nemo_parser` in `__main__.py` has no `--task` arg, unlike the standard build parser. Either wire it or drop the unused kwarg.
- **Brittle implicit weight sharing** — no `weight_prefix_map` on the model, so `apply_weights(prefix_map=None)` broadcasts every key against every sub-graph; this is what lets the deepcopied `encoder`/`encoder_streaming` share `encoder.*` initializers. Works, but document it on `RNNTTask.build` or declare an explicit map.
- **Untrusted-archive resource limits** — `_reader.py`: no caps on tar member count / config size / checkpoint size. Code-exec is mitigated (`weights_only=True`), but document that `.nemo` inputs must be trusted for resource consumption, or add sane limits.
- **Dead config fields** — `fastconformer_pos_emb_max_len`, `fastconformer_subsampling_factor`, `fastconformer_xscaling` are mapped/stored/asserted but never read (PE is dynamic; subsampling hardcodes 3 stages; xscaling forced False). Add a "stored for round-tripping, not applied" comment or drop them.
- **Streaming output-contract note** — `_forward_streaming` returns all `T_out` frames; NeMo trims to `valid_out_len` (`conformer_encoder.py:567-569`). Presumably delegated to the GenAI runtime — worth a one-line docstring note so a non-GenAI consumer doesn't double-count overlapped frames.
- **`batch_norm` attribute holds a `LayerNorm`** — confusing without a comment; add one noting it matches NeMo's checkpoint key `...conv.batch_norm.*` and is always a LayerNorm under `conv_norm_type="layer_norm"`.

### Question

- **Fully-masked query row**: NeMo does `masked_fill(-INF)` then post-softmax `masked_fill(0)`; this PR adds `-10000` and does **no** post-softmax zeroing, so a fully-masked row yields a value-average (note: with `-10000`, not `-inf`, this does **not** NaN). Benign today — offline padded rows are re-zeroed via `op.Where(valid, …)` and the streaming contract forbids intra-chunk padding — but please confirm no batched-streaming path can produce an all-masked query row that escapes the output zeroing.

### Nits
- Spurious parens making a non-tuple look like a tuple — `_config_mapping.py:1151` (`("fastconformer_rnnt")`).
- Two-char names hurt the dense kernels: `ll` reused with two meanings, `de`, `ch`/`ct`/`nch`/`nct` in `_forward_streaming`.
- `model_roles["decoder"] = "encoder"` (`tasks/_rnnt.py`) looks like a copy-paste error; add a comment that none of the four sub-models use decoder/GQA fusion.

### Praise
- The rectangular streaming `rel_shift` (query window offset by `cache_size` against a full `Tk` key window) is correctly derived, not copy-pasted from the square case — the easiest place to get Transformer-XL streaming wrong; verified exact via numpy simulation.
- Clean `NeMoArchive` reader; loud, actionable `_validate_encoder` failures; excellent shape-contract comments and ORT-GenAI layout docstrings.
- The looser joint parity (2.7e-5 vs 5e-7) is expected, not a red flag: an extra `LogSoftmax` over `vocab+1` atop a deeper accumulation chain, and it's argmax-invariant for greedy decode.

<sub>🤖 Synthesized from a 5-model review team (Claude readability + Opus deep/spec + GPT correctness & adversarial + Gemini integration). Spec claims grounded against NeMo source; the two top Majors verified directly against the diff.</sub>

justinchuby:
Thanks for the exceptionally thorough multi-model review — the spec-grounding against the NeMo source was especially valuable. Addressed in `2be9ce1` (+ earlier `f17b9db`, `dfdf05f`):

### Fixed

**Major 1 — streaming cache length** ✅ Now grows by the chunk's *valid* frame count (`length_sub`) instead of the physical post-drop `T_out`. Identical for full chunks (the validated path), correct for short/final chunks where padding previously inflated `cache_last_channel_len`.

**Major 2 — `_nemo_model_type` monkey-patch** ✅ Switched to the native `ArchitectureConfig.model_type` field; removed the `dataclasses.replace()` re-stash latent bug.

**Major 3 — `use_bias` guard** ✅ `_validate_encoder` now rejects `use_bias=True` (the stack hardcodes `bias=False`; non-strict loading would silently drop every bias).

**Major 4 — `att_context_style` / unlimited-right guards** ✅ Reject `att_context_style != "chunked_limited"` and `att_context_size[1] == -1` (the `chunk_size = right + 1 = 0` → div-by-zero case). `left == -1` stays handled (clamps to cache 70).

**Minor — list-valued `pre_encode_cache_size`** ✅ Already inconsistent between paths; `nemo_to_config` now mirrors the genai-config path (and the olive recipe) by taking the last entry.

**Minor — unpinned VAD download** ✅ Pinned `_SILERO_REVISION` to a fixed commit SHA.

**Nits** ✅ Fixed the spurious `("fastconformer_rnnt")` parens; documented the `"decoder": "encoder"` role mapping. Also inlined a now-trivial `_swish` wrapper into native `op.Swish`.

Added 4 unit tests (use_bias / att_context_style / unlimited-right / list configs). All 39 nemo unit + 5 integration tests (real model) pass, lint clean.

### Respectfully pushing back

**Minor — genai_config log-eps "inconsistency".** `_LOG_EPS = 2⁻²⁴` (genai_config `log_eps`) and `log_zero_guard_value = 1e-10` (audio_processor_config) are *different fields* consumed by different stages, and both match the upstream reference recipe ([`nvidia-nemotron-speech-streaming-en-0.6b/cpu/optimize.py`](https://github.com/microsoft/olive-recipes/blob/main/nvidia-nemotron-speech-streaming-en-0.6b/cpu/optimize.py#L163) lines 163 & 259) byte-for-byte. Collapsing them to one value would *diverge* from the reference, so I've left them as-is.

**Question — fully-masked query row.** Confirmed benign and no batched-streaming path can escape it: offline padded rows are re-zeroed via `op.Where(valid, …)`, and the streaming contract forbids intra-chunk padding (a chunk is either fully valid or `length=0`, and the cache-length fix above keeps masked keys out). With `-10000` (not `-inf`) there's also no NaN risk.

### Deferred (doc-only, low-risk)

Dead-field round-trip comments, untrusted-archive resource caps, the unused `--task` kwarg on `build_from_nemo`, and the explicit `weight_prefix_map` documentation — all noted; happy to follow up if you'd like them in this PR rather than a doc pass.

PR #363, opened 2026-06-19
Speculative-decoding drafters: Gemma4-Assistant, DFlash, Qwen3.6 MTP, EAGLE-3

jambayk:
## Summary

Adds speculative-decoding **drafter** support to mobius for four families, sharing a common pattern (drafter model + task IO contract + config + registry entry + golden tests): **Gemma4-Assistant**, **DFlash** (Qwen3), the **Qwen3.6 MTP** self-speculative head, and **EAGLE-3** chain drafters.

### Gemma4-Assistant (`gemma4-assistant`)
- `Gemma4AssistantCausalLMModel`: a small Gemma4-style head that consumes the target's `inputs_embeds` + per-layer-type **shared KV** (full / sliding) and emits `(logits, projected_state)`; supports the `use_ordered_embeddings` centroid-routed sparse LM head.
- Emits `com.microsoft::GroupQueryAttention` directly when the EP supports it; falls back to the generic Attention path otherwise.
- Config (`Gemma4AssistantConfig`) lifts the nested `text_config` + assistant fields; registry routes both `model_type` and the architecture name.
- L4/L5 golden tests (replayed target-derived tensors).

### DFlash (`dflash-draft`)
- `DFlashDraftModel`: block-diffusion drafter that conditions on multi-layer target hidden states; no embed/lm_head of its own (borrowed from the target). Component `DFlashDecoderLayer` with separate K/Q rotary position streams.
- `build(..., output_layer_indices=[...])` and `ArchitectureConfig.output_layer_indices` expose selected decoder layers' post-residual hidden states as `hidden_states.{k}` outputs for the drafter to consume.
- L4 golden test (hidden-state parity on `draft_hidden`).

### Qwen3.6 MTP (`qwen35-mtp`)
- `Qwen35MtpModel`: the Qwen3.6 multi-token-prediction head, a single `full_attention` Qwen3.5 decoder layer with an input projection `fc(concat[pre_fc_norm_embedding(inputs_embeds), pre_fc_norm_hidden(hidden_states)])`, reusing `Qwen35Attention` (doubled-Q gate, per-head Q/K offset-RMSNorm, partial mRoPE). Borrows the target's shared embed/lm_head, so it takes `inputs_embeds` + target `hidden_states` and emits `mtp_hidden`. `preprocess_weights` selects the checkpoint's `mtp.*` weights (HF discards them on load).
- `Qwen35MtpConfig.from_transformers` reads `text_config` and forces a single full-attention layer.
- L4 golden test, including a case-level `architecture` override (resolved via the registry to module class + task) so a shared base checkpoint can be routed to an auxiliary head; the golden is captured from a PyTorch reference since HF has no MTP module.

### EAGLE-3 (`eagle3-draft`)
- `Eagle3DraftModel`: a single-layer EAGLE-3 chain drafter. It borrows the target's `embed_tokens` and keeps its own compressed draft-vocab `lm_head` (the `d2t` / `t2d` draft-to-target remap buffers are dropped from the graph and consumed orchestrator-side). One graph serves both prefill and chain decode via `combined = fc(fused_hidden) + recycled_hidden`, where the caller zeros the unused input each step (fused at step 0, recycled at steps >= 1; `fc` has no bias so `fc(0) = 0`).
- `Eagle3Attention` subclasses the base attention with 2H-wide q/k/v projections (the concat of `input_layernorm(embed)` and `hidden_norm(combined)`), plain Llama attention with GQA and no qk-norm.
- Graph IO: inputs `inputs_embeds`, `fused_hidden` (3H aux), `recycled_hidden` (H), `attention_mask`, `position_ids`, single-layer GQA KV; outputs `draft_logits` (draft vocab), `next_hidden` (H, recycled), present KV.
- Two checkpoint formats are supported. **AngelSlim** (`Eagle3LlamaForCausalLM` for Qwen3-4B, `LlamaForCausalLMEagle3` for Qwen3-8B): flat config, `midlayer.*` weights. **speculators / RedHat** (`Eagle3Speculator` for Qwen3, `Eagle3DraftModel` for Gemma4): the arch config is nested under `transformer_layer_config`, weights are `layers.0.*`, a copy of the target embedding is shipped (dropped; the drafter borrows the target's), and `norm_before_residual` is set.
- `Eagle3Config.from_transformers` parses both layouts and normalizes rope (`rope_theta` or nested `rope_parameters`). The forward branches on `norm_before_residual` (residual seeded from the normalized hidden vs the raw fused feature), and the optional pre-fc norms `norm_before_fc` (one RMSNorm over the fused 3H) and `fc_norm` (per-aux RMSNorm on each H chunk) are applied before `fc`. `target_hidden_size != hidden_size` is rejected.
- Unit tests cover model params, the 2H attention shapes, both weight layouts, the nested config parse, `norm_before_residual` / `norm_before_fc` / `fc_norm` builds, and registry routing for all four architecture strings.

## Shared infrastructure
- `output_layer_indices` build kwarg + config field (intermediate hidden-state outputs), reused by both DFlash and EAGLE-3 to expose target aux hidden states.
- Golden harness: optional `architecture` field on test cases (force module class + task), `qwen35-mtp` task type, and `mtp_hidden` handling in the L4 extractor.

## Testing
- Unit tests for each drafter (model params, task graph IO, config parsing, registry routing).
- L4/L5 golden tests. The Qwen3.6 (27B) golden is `ci_skip` (too large for CI hardware) and runnable locally on GPU.


src/mobius/models/base.py
```diff
@@ -169,8 +177,11 @@ def forward(
                 attention_bias = None
 
         present_key_values = []
+        captured_hidden_states: list[ir.Value] = []
+        output_layer_indices = getattr(self, "output_layer_indices", None)
+        capture_set = set(output_layer_indices or ())
         past_kvs = past_key_values or [None] * len(self.layers)
-        for layer, past_kv in zip(self.layers, past_kvs):
+        for layer_idx, (layer, past_kv) in enumerate(zip(self.layers, past_kvs)):
             hidden_states, present_kv = layer(
```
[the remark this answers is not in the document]
jambayk: Good catch — fixed in 867893a. `TextModel.forward` now validates `output_layer_indices` up front (rejects duplicates and out-of-range indices) and builds the intermediate hidden-state list from an index→tensor dict, so every requested layer is guaranteed present and an invalid config fails fast with a clear `ValueError` instead of emitting a `None` that would crash graph building.

scripts/generate_golden.py
```diff
@@ -1343,6 +1343,264 @@ def _generate_phi4mm_multimodal(case: TestCase, json_path: Path, device: str) ->
...
+        "float16": torch.float16,
+        "bfloat16": torch.bfloat16,
+    }
+    torch_dtype = dtype_map.get(case.dtype, torch.float32)
+
+    tok = AutoTokenizer.from_pretrained(target_id)
+    target = AutoModelForCausalLM.from_pretrained(target_id, dtype=torch_dtype).to(device).eval()
+    assistant = (
+        AutoModelForCausalLM.from_pretrained(assistant_id, dtype=torch_dtype).to(device).eval()
+    )
```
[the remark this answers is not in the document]
jambayk: This is a false positive for the transformers version this script targets. `dtype=` is the current `from_pretrained` kwarg; `torch_dtype=` is the now-deprecated alias. The gemma4-assistant golden artifacts were generated successfully through this path with `dtype=`, so it does not raise. Leaving as-is.

tests/build_graph_test.py
```diff
@@ -4654,6 +4654,18 @@ def test_jamba_preprocess_weights_moe_renames(self):
...
     "bamba",
     "jamba",
+    # Speculative-decoding draft models with bespoke IO contracts
+    # (DFlash drafter takes noise_embedding + target_hidden instead of
+    # input_ids; the generic ALL_CAUSAL_LM_CONFIGS matrix can't drive it).
+    # Covered by src/mobius/models/_dflash_test.py.
+    "DFlashDraftModel",
+    # Gemma4-Assistant: scaffolding only — forward stub raises
+    # NotImplementedError; covered by
+    # src/mobius/models/_gemma4_assistant_test.py.
```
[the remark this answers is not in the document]
jambayk: Fixed in 867893a — the comment now describes the bespoke-IO reason (the assistant consumes `inputs_embeds` + the target's shared KV instead of `input_ids`, so the generic ALL_CAUSAL_LM_CONFIGS matrix can't drive it) rather than the outdated 'scaffolding / NotImplementedError' note.

src/mobius/models/_gemma4_assistant_test.py
```diff
@@ -0,0 +1,477 @@
...
+- Registry routes both ``model_type="gemma4_assistant"`` and
+  ``architectures=["Gemma4AssistantForCausalLM"]`` to the same class +
+  task.
+- Task name resolves to a ``Gemma4AssistantTask``.
+
+The end-to-end build path (``Gemma4AssistantTask.build`` →
+``Gemma4AssistantCausalLMModel.forward`` → ONNX) is intentionally **not**
+covered here — both raise ``NotImplementedError`` until the model
+forward is implemented; see the checklist in
+``mobius/models/gemma4_assistant.py``.
```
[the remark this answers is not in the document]
jambayk: Fixed in 867893a — the module docstring now reflects the implemented forward and the `TestGemma4AssistantBuildGraph` end-to-end build coverage (including the ordered-embeddings sparse LM head).

src/mobius/_configs/_base.py
```diff
@@ -1360,6 +1461,125 @@ def from_transformers(cls, config, parent_config=None) -> Gemma4Config:
...
+
+    Fields (assistant-specific):
+        backbone_hidden_size: Hidden size of the target model the
+            assistant was trained against (so the assistant's
+            ``pre_projection`` and ``post_projection`` know the right
+            input/output dims for the shared hidden state).
+        use_ordered_embeddings: When True, the assistant routes its
+            output through a centroid-based ordered-embedding LM head
+            (``Gemma4AssistantMaskedEmbedder``).  Currently unsupported
+            in mobius — see ``Gemma4AssistantCausalLMModel``.
```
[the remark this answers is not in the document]
jambayk: Fixed in 867893a — the `use_ordered_embeddings` docstring no longer says 'currently unsupported in mobius'; it now notes the centroid-routed ordered-embedding head is built by `Gemma4AssistantCausalLMModel`.

justinchuby (commented):
## Community review — PR #363 (multi-drafter: Gemma4-Assistant, DFlash, Qwen3.6 MTP)

This is primarily **static analysis** — I did not run the heavy golden/generation suite locally (see CI-parity note below), so the parity claims are reasoned from the graph construction, not reproduced._

**Verdict:** Clean, consistent, well-documented pattern overall — but there is **one functional MUST-FIX** (Gemma4-Assistant sliding-window masking is silently never applied) plus a couple of correctness contracts worth tightening before relying on parity. Inline comments mark the exact spots; this summary covers the cross-cutting items.

### 🔴 MUST-FIX — Gemma4-Assistant sliding-window masking is never applied
Detailed inline on `components/_gemma4_assistant.py` (`__init__` + non-GQA fallback) and `models/gemma4_assistant.py` (`local_window_size`). In short:
- `Gemma4AssistantAttention.__init__` never assigns `self.sliding_window`.
- `models/gemma4_assistant.py:311` reads `getattr(self.layers[0].self_attn, "sliding_window", -1) or -1` → **always `-1`**, so the GQA sliding guard (`components/_gemma4_assistant.py:149-150`) never fires, and the non-GQA fallback (`:191-204`) passes **no mask at all**.
- Net effect: sliding layers attend the **entire shared-KV buffer** instead of the last `sliding_window` tokens → divergence from HF whenever `kv_len > sliding_window` (512 in the test config). The golden uses a ~6-token prompt so it never triggers, and the unit tests only assert output shapes — so this is invisible to current coverage.
- This contradicts the production Gemma4 path (`models/gemma4.py:1497,1744,1773`) and the drafter's own documented design (`tasks/_gemma4_assistant.py:70-74`).
- **Fix:** store `self.sliding_window = config.sliding_window` in `__init__`, use it only for sliding layers, add `sliding_window` to the **fallback** attention bias too, and stop reading `self.layers[0]` (layer 0 is not guaranteed to be a sliding layer — read each layer's own type).

### 🟡 Should-address
1. **GQA symbolic-dim contract** (inline at `models/gemma4_assistant.py:293`): `seqlens_k`/`total_seq_len` are computed once and reused for both sliding and full layers, but the two layer types use **distinct symbolic KV dims** (`tasks/_gemma4_assistant.py:75-79` vs `:97-107`). This is only correct if the orchestrator feeds equal-length sliding/full buffers. Recommend sharing one `kv_len` symbolic dim or asserting the contract explicitly.
2. **CI parity is effectively unverified.** `gh pr checks` shows **L4 Golden / L5 Generation / Integration(fast)** failing because `libcudart.so.13` is missing at collection of `tests/e2e_golden_test.py` — an **environmental** failure, *not* introduced by this PR (the diff touches no deps / no onnxruntime pin). The consequence is that **none of the three drafter goldens actually executed**. On top of that, the Qwen3.6 MTP golden is `ci_skip`, so **MTP has zero executed parity coverage**. Suggest getting at least one drafter golden green (or a smaller MTP proxy) before treating parity as established.

### 🟢 Questions / nits (inline where the diff allowed; rest here)
- **DFlash mixed K normalization** (inline `components/_dflash.py:113`): context K/V come from the globally `hidden_norm`'d `target_hidden` (`models/dflash.py:97-98`) while noise K/V uses per-layer `input_layernorm` → `K = concat(differently-normalized ctx, noise)`. Please confirm this matches the `z-lab/dflash` reference.
- **DFlash past-KV re-append** (inline `components/_dflash.py:139`): `K = [past_K ‖ k_ctx ‖ k_noise]` re-appends `k_ctx` every step — confirm the orchestrator doesn't double-count context keys across speculative steps.
- **DFlash fp16 fc overflow:** the yaml note forces float32 in the fc projection due to fp16 overflow, but there's no **build-time guard** against an fp16 DFlash build. Consider adding one so a misconfigured build fails loudly instead of silently overflowing.
- **MTP golden asserts a vector, not a token** (inline `tests/e2e_golden_test.py:495`): the L4 golden takes argmax/cosine on `mtp_hidden` (the pre-`lm_head` vector) against a custom vLLM-ported PT reference (`draft/mtp/mtp_pt_reference.py`), not HF. Reasonable given no HF ref, but it validates a vector fingerprint, **not** the actual decoded draft token through the shared `lm_head`.
- **Gemma4-Assistant `layer_scalar`** (inline `components/_gemma4_assistant.py:264`): per-layer `nn.Parameter([1])`, but the comment calls it a "tied buffer." Confirm HF stores a per-layer tensor; otherwise loading silently leaves it at the `1.0` default.
- **Cosmetic** (inline `_registry.py:466`): the comment references a nonexistent `_ARCHITECTURE_OVERRIDES` constant in `_builder.py` (the task description attributed this to `_builder.py:391`; it actually lives in `_registry.py`).

### Per-drafter confidence
| Drafter | Static analysis | Executed parity | Confidence |
|---|---|---|---|
| Gemma4-Assistant | sliding-window mask bug (🔴) | golden not run (env) | **Low until fixed** |
| DFlash | sound; 2 contracts to confirm (🟢) | golden not run (env) | Medium |
| Qwen3.6 MTP | sound; vector-only golden | `ci_skip` + env fail → none | **Low (unverified)** |

### What's good (verified sound)
`output_layer_indices` plumbing is backward-compatible (default preserves the legacy 2-tuple; `len(result)==3` guards the new path). The DFlash post-residual hidden-state capture point is correct (`base.py:201-216`). The MTP `fc`/concat ordering, the single forced `full_attention` layer, the `mtp.*` weight selection, and the registry architecture-override routing all check out. The centroid-routed sparse LM head has no off-by-one. This is a clean, consistent pattern — the MUST-FIX is a single missing assignment with an outsized blast radius, not a structural problem.

src/mobius/components/_gemma4_assistant.py
```diff
@@ -0,0 +1,437 @@
...
+        self.num_attention_heads = config.num_attention_heads
+        # Full-attention layers may use a distinct kv-head count; sliding always
+        # uses the standard num_key_value_heads.  Mirrors Gemma4TextAttention.
+        if is_full and config.num_global_key_value_heads is not None:
+            self.num_key_value_heads = config.num_global_key_value_heads
+        else:
+            self.num_key_value_heads = config.num_key_value_heads
+        # Gemma4 hardcodes scale=1.0 (not 1/sqrt(d)); attn_logit_softcapping
+        # is wired through to the ONNX Attention op's native ``softcap``.
+        self.scaling = 1.0
```
justinchuby, line 96: 🔴 **MUST-FIX (1/2): `self.sliding_window` is never set in `__init__`.**

This constructor sets `head_dim`, kv-head counts, `scaling`, `softcap`, RoPE config… but never assigns `self.sliding_window`. As a result, the model-level lookup at `models/gemma4_assistant.py:311` (`getattr(self.layers[0].self_attn, "sliding_window", -1) or -1`) always resolves to `-1`, so:
- the GQA sliding guard at `:149-150` (`if gqa_ctx.local_window_size > 0`) never fires, and
- the non-GQA fallback at `:191-204` passes **no mask at all** (see my other inline).

Net: sliding layers attend the **full shared-KV buffer**, diverging from HF whenever `kv_len > sliding_window` (512 in the test config). The ~6-token golden never trips it and the unit tests assert shapes only, so it's silent today. Production Gemma4 wires this correctly (`models/gemma4.py:1497,1744,1773`) and the drafter's own task doc promises it (`tasks/_gemma4_assistant.py:70-74`).

**Fix:** add `self.sliding_window = config.sliding_window` here, and apply it only on sliding layers.

src/mobius/components/_gemma4_assistant.py
```diff
@@ -0,0 +1,437 @@
...
+        # Full attention: is_causal=0, no attention_bias.  See module
+        # docstring for why this is correct.  For q_len < kv_len the two EPs
+        # disagree on ``is_causal`` alignment; with is_causal=0 + no bias the
+        # graph is correct on every EP.
+        attn_output, _present_k, _present_v = _apply_attention(
+            op,
+            q,
+            k,
+            v,
+            attn_mask=None,
```
justinchuby: 🔴 **MUST-FIX (2/2): non-GQA fallback passes `attn_mask=None` → no sliding mask.**

Even once `self.sliding_window` exists, this generic-attention fallback hands `attn_mask=None` / `is_causal=0` with no windowed bias, so a sliding layer routed through here attends the entire shared K/V. The full-attention case is fine, but a **sliding** layer here must receive a window-limited attention bias.

**Fix:** when this layer is a sliding layer, build and pass a sliding-window attention bias (last `self.sliding_window` keys) instead of `attn_mask=None`.

src/mobius/models/gemma4_assistant.py
```diff
@@ -0,0 +1,350 @@
...
+                to=ir_mod.DataType.INT32,
+            )
+            gqa_ctx_by_type: dict[str, GQAContext] = {
+                "sliding_attention": GQAContext(
+                    seqlens_k=seqlens_k,
+                    total_seq_len=total_seq_len,
+                    cos_cache=self.rotary_emb_local.cos_cache,
+                    sin_cache=self.rotary_emb_local.sin_cache,
+                    local_window_size=(
+                        getattr(self.layers[0].self_attn, "sliding_window", -1) or -1
```
justinchuby: 🔴 **MUST-FIX root cause: `local_window_size` is always `-1`, and `self.layers[0]` is the wrong source.**

Two problems on this line:
1. `getattr(self.layers[0].self_attn, "sliding_window", -1)` — `sliding_window` is never set on `Gemma4AssistantAttention` (see `components/_gemma4_assistant.py.__init__`), so this is **always `-1`** and the sliding GQA guard at `components/_gemma4_assistant.py:149-150` never activates.
2. Reading `self.layers[0]` assumes layer 0 is a sliding layer — that isn't guaranteed by the Gemma4 sliding/full interleave pattern. Each sliding `GQAContext` should derive its window from the **sliding** config, not from whatever layer happens to be index 0.

**Fix:** set `self.sliding_window = config.sliding_window` in the attention `__init__`, and source `local_window_size` from `config.sliding_window` for the `sliding_attention` context (drop the `self.layers[0]` lookup).

src/mobius/models/gemma4_assistant.py
```diff
@@ -0,0 +1,350 @@
...
+        if use_gqa:
+            # Materialize RoPE caches as graph initializers (GQA references
+            # them directly).  Calling .forward() on the RoPE module is what
+            # registers the cos/sin Parameter — without this they would be
+            # absent from the graph.
+            _ = self.rotary_emb_local(op, position_ids)
+            _ = self.rotary_emb_global(op, position_ids)
+
+            one_i32 = op.Constant(value_int=1)
+            seqlens_k = op.Cast(
```
justinchuby, line 299: 🟡 **`seqlens_k` / `total_seq_len` are computed once here and reused for both layer types.**

Both the `sliding_attention` and `full_attention` `GQAContext`s below share this single `seqlens_k`/`total_seq_len`. That's only correct if the sliding and full **KV buffers are equal length** — but the task defines them with **distinct symbolic dims** (`tasks/_gemma4_assistant.py:75-79` sliding vs `:97-107` full). If the orchestrator ever feeds unequal-length sliding/full buffers, the shared `total_seq_len` is wrong for one of them.

Recommend either sharing a single `kv_len` symbolic dim across both layer types, or asserting the equal-length contract explicitly so a future divergence fails loudly.

src/mobius/components/_gemma4_assistant.py
```diff
@@ -0,0 +1,437 @@
...
+        self.input_layernorm = RMSNorm(config.hidden_size, eps=config.rms_norm_eps)
+        self.post_attention_layernorm = RMSNorm(config.hidden_size, eps=config.rms_norm_eps)
+        self.pre_feedforward_layernorm = RMSNorm(config.hidden_size, eps=config.rms_norm_eps)
+        self.post_feedforward_layernorm = RMSNorm(config.hidden_size, eps=config.rms_norm_eps)
+        # ``layer_scalar`` is a tied (persistent) buffer in HF
+        # (``register_buffer("layer_scalar", torch.ones(1))``).  In mobius
+        # we declare it as a Parameter so the weight loader populates it
+        # from the HF state dict; at runtime it's applied as the last op
+        # in this layer.
+        self.layer_scalar = nn.Parameter([1])
```
justinchuby, line 290: 🟢 **`layer_scalar` is a per-layer `nn.Parameter([1])`, but the comment above calls it a "tied buffer."**

If HF actually stores `layer_scalar` as a per-layer tensor in the state dict, fine — but please confirm. If the name/shape doesn't match the HF key, weight loading will **silently** skip it and leave the scalar at its `1.0` default, which would be a quiet parity bug (no error, just wrong scaling).

src/mobius/components/_dflash.py
```diff
@@ -0,0 +1,190 @@
...
+            position_embeddings=q_position_embeddings,
+            num_heads=self.num_attention_heads,
+            rotary_embedding_dim=self.rotary_embedding_dim,
+            interleaved=self._rope_interleave,
+        )
+
+        # --- K, V from [target_hidden ‖ noise] ---
+        k_ctx = self.k_proj(op, target_hidden)
+        k_noise = self.k_proj(op, hidden_states)
+        k = op.Concat(k_ctx, k_noise, axis=1)
```
justinchuby, line 113: 🟢 **Mixed normalization in the K concat — please confirm against the reference.**

`k_ctx` is projected from `target_hidden`, which was **globally** `hidden_norm`'d (`models/dflash.py:97-98`), while `k_noise` is projected from `hidden_states` that went through the **per-layer** `input_layernorm`. So `K = concat(differently-normalized ctx, noise)`. Same applies to V on the next lines.

This may well be intentional (it mirrors a cross-attention design), but it's subtle enough to verify against `z-lab/dflash` `dflash/model.py` before trusting parity.

src/mobius/components/_dflash.py
```diff
@@ -0,0 +1,190 @@
...
+        # --- Non-causal attention.  op.Attention concatenates past_key /
+        # past_value with the supplied K / V internally, giving the final
+        # K = [past_K ‖ k_ctx ‖ k_noise] — exactly what DFlash needs. ---
+        attn_output, present_key, present_value = _apply_attention(
+            op,
+            q,
+            k,
+            v,
+            attn_mask=None,
+            past_key=past_key_value[0] if past_key_value is not None else None,
```
justinchuby, line 139: 🟢 **`K = [past_K ‖ k_ctx ‖ k_noise]` re-appends `k_ctx` every step — confirm no double-counting.**

With `past_key_value` set, `op.Attention` concatenates `past_key` with the freshly-built `k` (which already includes `k_ctx`). Across multiple speculative steps that means the context keys get appended again each step unless the orchestrator deliberately excludes prior context from `past_key`. Please confirm the orchestrator's KV-cache contract so context keys aren't double-counted across draft steps.

tests/e2e_golden_test.py
```diff
@@ -471,6 +485,14 @@ def _extract_logits(
...
     if "logits" in outputs:
         return outputs["logits"]
+    if task_type == "dflash-draft" and "draft_hidden" in outputs:
+        # DFlash outputs draft_hidden (pre-lm_head). Treat it as the logit
+        # vector for the argmax gate (mirrors the hidden-state task handling).
+        return outputs["draft_hidden"]
+    if task_type == "qwen35-mtp" and "mtp_hidden" in outputs:
+        # The Qwen3.6 MTP head outputs mtp_hidden (pre-lm_head, the head borrows
+        # the target's lm_head). Treat it as the logit vector for the argmax gate.
+        return outputs["mtp_hidden"]
```
justinchuby, line 522: 🟢 **MTP golden validates a pre-`lm_head` vector, not the decoded draft token.**

`_extract_logits` returns `mtp_hidden` (the head's pre-`lm_head` hidden vector), and the L4 golden takes argmax/cosine on it against a custom vLLM-ported PT reference (`draft/mtp/mtp_pt_reference.py`), not HF. That's reasonable given there's no HF reference, but worth flagging: it asserts a **vector fingerprint**, not the actual draft **token** produced after the shared `lm_head`. Combined with this golden being `ci_skip`, MTP token-level correctness is currently unasserted.

src/mobius/_registry.py
```diff
@@ -424,6 +461,20 @@ def _detect_fallback_registration(hf_config) -> ModelRegistration | None:
     "qwen": ModelRegistration(QwenCausalLMModel),
     "qwen3": ModelRegistration(Qwen3CausalLMModel),
     "qwen3_5_text": ModelRegistration(Qwen35CausalLMModel),
+    # DFlash drafters share ``model_type=qwen3`` with the base Qwen3 family;
+    # build() routes to DFlashDraftModel via the ``architectures`` field
+    # (see _ARCHITECTURE_OVERRIDES in _builder.py).  The registry entry
```
justinchuby: 🟢 **Cosmetic:** this comment references `_ARCHITECTURE_OVERRIDES` in `_builder.py`, but no such constant exists. The actual re-routing lives in `build()`'s architecture-field logic. Either drop the name or point at the real symbol to avoid sending readers on a dead-end lookup.

titaiwangms:
## Review team synthesis — PR #363 (speculative-decoding drafters)

Ran a 5-model review fan-out (readability / code / critical / deep / integration) over the ~4.4k-line diff. The deep reviewer cross-checked the math against the public references (vLLM `qwen3_next_mtp.py`, `z-lab/dflash`) and the in-tree Gemma4 path. Findings deduplicated and prioritized; load-bearing claims verified against the PR branch.

### 🔴 Major

**1. Gemma4-Assistant `sliding_attention` layers don't apply the sliding-window mask (both emission paths).** *(triple-converged: code + critical + deep reviewers)*
- GQA path: `local_window_size` is read as `getattr(self.layers[0].self_attn, "sliding_window", -1) or -1` (`models/gemma4_assistant.py:311`), but `Gemma4AssistantAttention` is constructed with `sliding_window=None` (`:243`) and never sets the attribute — so it resolves to `-1` (window disabled) for **every** layer.
- Fallback/generic path: sliding layers run with `attn_mask=None, is_causal=0` and no window at all.
Both diverge from the production Gemma4 path, which builds `create_attention_bias(..., sliding_window=...)` for `sliding_attention` layers (`models/gemma4.py:1769`), and from HF. The golden case is structurally blind to it (14-token context ≪ sliding window, so window never truncates). Impact is bounded — wrong drafts only lower acceptance rate, the target still verifies, so final output stays correct — but it's a real reference deviation and an inter-path inconsistency (the GQA path *tries* to pass a window, which is itself evidence the shared KV is expected to exceed it).
**Fix:** thread `config.sliding_window` into `Gemma4AssistantAttention` for sliding layers and build a windowed bias in the fallback path (mirroring `gemma4.py:1769`); or assert the shared sliding-KV is pre-windowed and drop `local_window_size` for symmetry.

**2. Gemma4-Assistant skips Gemma4 final logit soft-capping.** *(critical; confirmed static)*
`models/gemma4_assistant.py:189-211` returns `lm_head`/`masked_embedding` logits directly, with no `cap * tanh(logits/cap)`. The production Gemma4 model applies it (`models/gemma4.py:1886,1958`) and `Gemma4AssistantConfig` inherits `final_logit_softcapping` from `Gemma4Config`. Argmax-only goldens pass (tanh is monotonic), but logit magnitudes / sampling probabilities / HF parity are wrong **if the checkpoint sets the cap**.
**Action:** confirm whether the assistant checkpoint carries a nonzero `final_logit_softcapping`; if so, apply it after both head branches and add a fixture with a nonzero cap.

**3. `output_layer_indices` is a silent contract failure on non-supporting modules.** *(integration; also raised by code reviewer)*
`tasks/_causal_lm.py` (both `CausalLMTask.build` and `HybridCausalLMTask.build`). If `config.output_layer_indices` is set but the module returns a 2-tuple, `intermediate_hidden_states` is `None` and `_register_intermediate_hidden_states` is silently skipped — the build succeeds but the requested `hidden_states.{k}` graph outputs are dropped, surfacing as a downstream crash at deploy time rather than export time. Custom backbones (`Qwen35TextModel`, `Qwen35MoETextModel`, `_JambaTextModel`, …) don't implement the capture, so a future DFlash-on-Qwen3.5 pairing hits exactly this.
**Fix:** in both `build()` methods, raise `ValueError` when `output_layer_indices` is requested but the module returned no intermediate hidden states.

### 🟡 Minor

**4. DFlash/MTP goldens treat hidden states as logits.** The L4 extractor returns `draft_hidden`/`mtp_hidden`, so `top1_id` is a hidden-dim index, not a vocab token id. The cosine/tolerance gate is the real check (deep reviewer confirms it's a genuine reference-derived parity gate, not vacuous), but it can't catch borrowed-`lm_head` wiring, tied-head dtype/layout, or orchestrator decode bugs. Rename to make "hidden-state parity only" explicit, or add a token-logit integration test.

**5. Golden generator missing a `qwen35-mtp` handler.** `scripts/generate_golden.py` registers `gemma4-assistant` and `dflash-draft` but not `qwen35-mtp`, so that golden isn't reproducible through the standard toolchain and will drift.

**6. Replay feeds hard-cast to `float32`** for the gemma-assistant / dflash paths (`tests/e2e_golden_test.py`), unlike the dtype-aware MTP path — will diverge if these cases gain fp16/bf16 variants. Thread `case.dtype` through.

**7. Missing negative tests** for duplicate / out-of-range `output_layer_indices`, despite the new validation logic.

### ❔ Open questions (deferred — need confirmation, not blocking)

- **DFlash target-context caching.** The critical reviewer flagged that `[target_hidden ‖ noise]` K/V flows into `present_key_values`, implying per-block re-caching / cache growth. The deep reviewer verified DFlash's K/V construction is **byte-for-byte faithful** to `z-lab/dflash` (`concat → k_norm → RoPE`, `is_causal=0`) and the golden is captured from the reference `spec_generate()`. Per the spec-grounded verdict this is likely **not a bug**, but please confirm the cache semantics match how the orchestrator re-feeds `target_hidden` per block.
- **`MaskedEmbedder` assumes `token_ordering` is a true vocab permutation.** `ScatterElements`/torch `scatter_` parity holds only if selected centroids map to disjoint token sets; add a one-line assert/comment that `token_ordering` is a permutation.
- **`layer_scalar` per-layer multiplier** (`_gemma4_assistant.py:1266,1299`) couldn't be checked against an upstream reference (model is private). Defaults to identity, so harmless if absent, but confirm placement against the upstream `Gemma4TextDecoderLayer`.

### 🔵 Nits / readability
- `if len(result) == 3:` tuple-length dispatch in `base.py` / `_causal_lm.py` is an invisible contract — name the variable + comment, or return a small dataclass.
- `Qwen35MtpModel.forward` uses `hs` instead of the codebase-wide `hidden_states`; and `target_hidden` as a *local* collides with its cross-file meaning ("tensor from the target model").
- `qwen3_6-mtp.yaml` filename uses `_` where siblings use `-`; `:class:` Sphinx role used on a dataclass field (should be `:attr:`).

### ✅ Strengths
Verified-correct against references: Qwen3.6 MTP concat/projection ordering (`fc(concat[norm(embed), norm(hidden)])`) matches vLLM exactly; DFlash split Q/K rotary streams (`q_position_ids = position_ids[:, -q_len:]`) reproduce the reference slice precisely; `output_layer_indices ↔ DFlash offset=1` capture convention matches end-to-end; goldens are reference-derived parity gates. Excellent module docstrings (topology rationale + per-tensor shapes), clean separation of drafter IO from the generic causal-LM path, and good dup/range validation on `output_layer_indices`.

*Generated by a multi-model review team (readability/code/critical/deep/integration); spec claims grounded against vLLM + z-lab/dflash references and the PR branch.*

src/mobius/components/_gemma4_assistant.py
jambayk: Fixed in b03e1b1. `Gemma4AssistantAttention.__init__` now sets `self.layer_type` and `self.sliding_window = config.sliding_window` (only on sliding layers), so the GQA `local_window_size` guard activates and the fallback can build a windowed bias.

src/mobius/components/_gemma4_assistant.py
jambayk: Fixed in b03e1b1. The generic-attention fallback now builds a causal+sliding bias via the shared `create_attention_bias(..., sliding_window=self.sliding_window)` for sliding layers (causal is a no-op for the q_len==1 draft step), instead of passing `attn_mask=None`. `attention_mask` is threaded through the decoder layer to the attention.

src/mobius/models/gemma4_assistant.py
jambayk: Fixed in b03e1b1. The sliding GQA context now sources `local_window_size` from `config.sliding_window` (stored on the text model as `self.sliding_window`), dropping the `self.layers[0]` lookup, and `self.sliding_window` is now set in the attention `__init__`.

src/mobius/models/gemma4_assistant.py
jambayk: Documented in b03e1b1 with an explicit NOTE: a single `seqlens_k`/`total_seq_len` (from one `attention_mask`) drives both GQA contexts, which assumes the orchestrator feeds equal-length sliding/full shared-KV buffers — it shares one max-cache-len buffer. Kept the distinct symbolic dims for the graph signature but flagged the contract.

src/mobius/components/_gemma4_assistant.py
jambayk: Confirmed + comment clarified in b03e1b1. HF registers it per layer (`register_buffer("layer_scalar", torch.ones(1))`), so the state-dict key is `model.layers.{i}.layer_scalar` and the mobius `nn.Parameter([1])` per layer matches it. Reworded the misleading 'tied buffer' wording.

src/mobius/components/_dflash.py
jambayk: Confirmed against the reference — intentional. In z-lab/dflash `dflash/model.py`: `target_hidden` is globally `hidden_norm(fc(...))`'d once before the layer loop (line ~334), while `hidden_states` goes through the per-layer `input_layernorm` (line ~281); then `k_ctx = k_proj(target_hidden)`, `k_noise = k_proj(hidden_states)`, `k = cat([k_ctx, k_noise])` (lines ~226-230). So the mixed normalization in the concat mirrors the reference exactly (same for V).

src/mobius/components/_dflash.py
jambayk: Confirmed — no double-counting, by the orchestrator's crop contract. The reference loop crops the draft KV back each block (`past_key_values_draft.crop(start)`, dflash/model.py ~line 120) before appending the fresh `[k_ctx ‖ k_noise]`, so context keys are not accumulated across steps. Our orchestrator follows the same contract (`draft_kv.crop(start)` per block), so `past_key` never carries a prior step's `k_ctx`.

tests/e2e_golden_test.py
jambayk: Acknowledged — by design the MTP head has no `lm_head` of its own (it borrows the target's, `mtp_use_dedicated_embeddings=false`), so the ONNX output is `mtp_hidden`; the L4 golden fingerprints that vector against the vLLM-ported PT reference (HF has no MTP module). Token-level + end-to-end correctness is covered outside this harness: in the archive workspace the mobius MTP ONNX reproduces the PT reference at cosine 1.0, matches vLLM's ~0.966 single-token acceptance, and the self-speculative ORT loop is verified token-identical (lossless) to greedy across prompts. Happy to add a post-lm_head token assertion if we later wire the shared lm_head into this golden case.

src/mobius/_registry.py
jambayk: Fixed in b03e1b1 — dropped the dead `_ARCHITECTURE_OVERRIDES` reference; the comment now just says build() routes by the checkpoint config's `architectures` field.

justinchuby:
@jambayk LMK when this is ready!

justinchuby (approved):

Issue #349, opened 2026-06-09
Emit opset-24 external KV cache (nonpad_kv_seqlen) for bias decoders to recover near-GQA decode (Gemma-4, sliding-window)


justinchuby:
### Summary

For models that require a **float attention bias** on the decoder (e.g. Gemma‑4 multimodal's bidirectional vision‑block overlay, or any sliding‑window model), mobius currently emits the decoder with an **internal `past`/`present` KV cache**. With a bias present this routes ORT's ONNX‑domain `Attention` op (`core/providers/cuda/llm/attention.cc`) to the **internal‑cache MEA path**, which the op's own in‑source perf note (L205‑231) calls ~15‑30% slower than contrib GQA's in‑place decode.

ORT already supports a faster alternative on the same standard `Attention` op: the **opset‑24 external KV cache** (`nonpad_kv_seqlen` + `TensorScatter`). On the MEA path it combines `attn_mask`/additive bias **+ GQA + external cache** and "achieves near‑parity with contrib GQA performance" (same note, plus ~L778‑781, L834). This would let a **single** standard‑`Attention` graph carry the bidirectional‑block bias during prefill *and* hit the near‑GQA decode path — no contrib GQA, no prefill/decode model split, and correctly (Flash is precluded by any bias and by `head_dim=512` global layers, so MEA‑external ≈ GQA is the realistic ceiling here).

### Task

Add an exporter path that emits the decoder using the **opset‑24 external KV cache** form (`nonpad_kv_seqlen` + `TensorScatter`) for models that need a decoder attention bias, instead of internal `past/present`.

Applies to Gemma‑4 (`gemma4_unified`, `use_bidirectional_attention == "vision"`) and is generally relevant to sliding‑window / float‑bias decoders.

### Blocked on / coordinate with runtime

This is only end‑to‑end useful once **onnxruntime‑genai** drives the external‑cache path. Today genai's KV‑cache manager (`src/models/kv_cache.cpp`) only knows `past_present_share_buffer` (contrib GQA in‑place) vs the default dynamic `past`/`present`; it has no `nonpad_kv_seqlen` / `TensorScatter` support. Tracked upstream in **microsoft/onnxruntime-genai#2204**.

### Acceptance / verification

- Decoder graph emits `nonpad_kv_seqlen` (opset 24) external cache; `genai_config.json` wired accordingly.
[...]

PR #364, opened 2026-06-19
Enable static-cache + Flash-attention path (runtime-gated, ready for ORT #28958)

### LINKED ISSUE #329: CI: add numerical parity coverage for the static-cache export path
## Gap

PR #328 adds the static-cache ONNX `Attention` export path (`is_causal=0` + a phase-split `If`). Its correctness is covered in CI on two axes, but a third is missing:

- **Structural correctness** — CI-gated by `tests/build_graph_test.py` (CPU, runs per-PR).
- **Runtime-CUDA kernel routing** — now gated by `tests/static_cache_decode_test.py` in the integration-fast GPU job (added in #328).
- **Numerical parity** — **not covered.** The committed `tests/synthetic_parity_test.py` only exercises the **dynamic** path (`is_causal=1`). The 20/20 static-path parity validated during #328 development was a **manual QA run** (`parity_llama.json`), not a committed test.

## Proposed follow-up

Author a synthetic static-cache parity test:

- Build a tiny in-memory synthetic model (same approach `static_cache_decode_test.py` uses).
- Export the **static-cache** variant and a **dynamic reference**.
- Assert token-id and/or logit parity within tolerance over N decode steps.
- Wire it into a **GPU CI tier** (static-cache decode requires CUDA).

This complements the runtime-kernel routing test added in #328 — that test checks the *right kernel runs*; this one would check the *numbers are right*.

Refs: #328

titaiwangms:
## What

mobius `main` already emits the correct maskless `is_causal=1` + `nonpad_kv_seqlen` + `TensorScatter` static-cache decoder graph. This PR adds the runtime **enablement, verification, and CI** for that path on the CUDA Flash-attention kernel — it is **not** graph surgery. Nothing in the emitted graph is reverted or rewritten.

The path becomes runnable once an ONNX Runtime build containing [microsoft/onnxruntime#28958](https://github.com/microsoft/onnxruntime/pull/28958) is installed (Flash eligibility widening for the bottom-right-causal errata [onnx/onnx#8068](https://github.com/onnx/onnx/pull/8068)).

## Issues

- Closes #329 — this PR fully delivers the static-cache numerical-parity CI coverage that #329 asks for (`tests/static_cache_parity_test.py`: static vs dynamic vs HuggingFace, chunked-prefill zero-guard, V convex-hull valid-row invariant, exact-triangle `nonpad == q_seq`).
- Part of #345 — this PR delivers the code/probe/CI half of #345 ("emit maskless `is_causal=1` + `nonpad_kv_seqlen` static-cache graph once upstream lands"). The ORT pin bump + node-count rebaseline + ONNX 1.22 pin + skill docs remain deferred until an official PyPI ORT release with microsoft/onnxruntime#28958, so #345 stays open (intentionally `Part of`, not `Closes`).

## Changes

- **(a) Conditional opset 24→23 lowering** in `_builder.py` via `_graph_requires_opset24` (a **recursive** subgraph scan) + `_apply_opset_lowering`, so graphs carrying `TensorScatter` or the `Attention` `nonpad_kv_seqlen` input correctly **stay at opset 24**. Flag-gated, default off. (`_builder_test.py` exercises the real branch.)
- **(b) Canonical capability probe** `src/mobius/_testing/ort_capabilities.py` — `supports_static_cache_flash()` is a **functional, fail-closed-but-loud runtime probe** (not a version-string check). It builds a minimal `TensorScatter` + maskless `Attention` graph and runs it on CUDA. A **known-answer value check** closes a latent CPU-fallback fail-open: because ORT implicitly appends the CPU EP, a CUDA build that declines the node would silently run on CPU with wrong (top-left) values; the probe's deterministic reference (`2.0`) rejects that → `NEEDS_FIX` → `False`. A structured `_ProbeOutcome` enum distinguishes the expected pre-#28958 reject from unexpected probe errors (logged with `exc_info`).
- **(c) Static-cache parity test (#329)** `tests/static_cache_parity_test.py` — static vs dynamic vs HuggingFace, chunked-prefill zero-guard, V convex-hull valid-row invariant, exact-triangle `nonpad == q_seq`.
- **(d) e2e CUDA Flash-dispatch test** `tests/static_cache_flash_e2e_test.py` — asserts the ONNX-domain `Attention` actually routes to **Flash** (via VERBOSE dispatch capture), gated on SM ≥ 8.0 (`_flash_capable_gpu`) and `onnxruntime_QUICK_BUILD`.
- **(e) Probe consolidation** — deleted `tests/_static_cache_support.py`; one canonical probe module, no shim.

## Gating

All new GPU tests **skip automatically** unless the installed ORT can actually run the path (probe-gated). CI stays green today and **flips green automatically** once an official ORT release containing microsoft/onnxruntime#28958 is installed — **zero code change needed** to enable.

## Verified

- **5/5** static-cache tests pass on an A100 (SM 8.0) with a post-#28958 ORT (full targeted suite: 14 passed including builder tests).
- **Fail-closed on pre-#28958 confirmed two ways**: (1) source — the CUDA kernel guard `causal_cross_no_past = is_causal && (q_seq != total_seq) && (past == 0)` in `Attention<T>::ComputeInternal` (`llm/attention.cc`) raises `NOT_IMPLEMENTED` for the `S_q=1` decode shape (no fast-path bypass); (2) empirically — a real `onnxruntime-gpu==1.27.0` isolated venv raises `NotImplemented` → `supports_static_cache_flash() == False`.

## Explicitly out of scope (intentionally held, separate follow-ups)

- ORT dependency pin bump + node-count rebaseline + ONNX 1.22 pin + skill docs — parked until an official PyPI ORT release with #28958 exists.
- `examples/static_cache_generation.py` nonpad-before-scatter check (`verify-example-nonpad`).
- Pre-existing `RUF067` ruff version-skew in `pyproject.toml`.
- onnxruntime-genai#2204-blocked bias-decoder external-KV work (#349).

## References

- [onnx/onnx#8068](https://github.com/onnx/onnx/pull/8068) — causal top-left → bottom-right errata
- [microsoft/onnxruntime#28958](https://github.com/microsoft/onnxruntime/pull/28958) — Flash eligibility widening

---


titaiwangms:
## Review summary

**Verdict: sound, mergeable.** No Critical issues across a five-reviewer pass (readability / correctness / adversarial / spec-deep / cross-module). The math and spec are independently verified: opset-24 `Attention` input #6 `= nonpad_kv_seqlen` ✓, the `MOBIUS_ORT_LOWER_OPSET_FOR_EP` default is `False` ✓, and the CPU `_builder_test.py` passes (9/9) ✓. **2 Major, several Minor/Nit, 2 open questions** below.

### 🟠 Major

**M1 — The probe's `NEEDS_FIX` classification is too broad, which contradicts its own "fail-closed but not fail-silent" thesis.** (`src/mobius/_testing/ort_capabilities.py`)
- Generic `onnxruntime.Fail` is in `_EXPECTED_REJECT_ERRORS` (~line 350), so a real runtime failure on a *fixed* build (CUDA OOM, illegal memory access, `TensorScatter` crash) is mislabeled "needs #28958" and silently skips the suite.
- Any value mismatch maps to `NEEDS_FIX` logged at **DEBUG** (~line 614), not just the known top-left `1.0` signature — so a genuine numerical regression is masked as "needs fix."
- *Suggestion:* only treat `Fail` as expected when the message carries the known reject signature; classify other mismatches/failures as `PROBE_ERROR` at WARNING.

**M2 — Duplicated opset-lowering logic has drifted.** `src/mobius/_testing/ort_inference.py` `_should_lower_opset` (lines 124-127) only scans **top-level** nodes for `TensorScatter`; it lacks both the recursion *and* the `Attention` input-#6 check that the new canonical `_graph_requires_opset24` (`src/mobius/_builder.py`) has. Not corrupting today (the static-cache path co-emits a top-level `TensorScatter`, so it is caught), but it is a latent hazard: a maskless `Attention` without `TensorScatter`, or either op nested inside `If`/`Loop`/`Scan`, would be silently lowered to opset 23 → invalid model. *Suggestion:* have `_should_lower_opset` reuse `_graph_requires_opset24` (extract to a shared util if import cycles bite).

### 🟡 Minor
- **Collection-time CUDA probe:** `tests/static_cache_flash_e2e_test.py` evaluates `supports_static_cache_flash()` in a module-level `pytestmark`, so the functional CUDA probe (build graph + create CUDA session) runs at import/collection, before fixtures — an xdist / lazy-CUDA-init hazard. The parity test's setup-time helper pattern is cleaner; consider matching it.
- **Global-state mutation** in `captured_attention_dispatch`: restores the ORT logger severity to a hardcoded `2` rather than the prior value, and redirects process fd 2 (blast radius is test-only, but it is not concurrency-safe).
- **Probe does not assert CUDA node *placement*** (acknowledged residual in the docstring; the known-answer value check mitigates the realistic case).
- **`_apply_opset_lowering` lacks the `original <= 23` guard** that `ort_inference.py` has, so a sub-model already at ≤23 would be "upgraded" to 23 with a misleading `Lowered 21→23` log (latent; all packages are 24 today).
- **Unenforced load-bearing invariant** `_PROBE_KEY_FILL == _PROBE_QUERY_FILL` — collapse to one constant or add an `assert`.

### 🔵 Nits
`_require_static_cache_attention` skips (not fails) — the `require_*` name implies otherwise; stale `name="static_cache_attention_probe"` in the *parametric* builder in the parity test; `FLASH_CACHE_DTYPE` governs all tensors, not just caches; double-negation guard in `_builder.py`; no `ai.onnx`-domain test for the `Attention` input-#6 branch; `_flash_capable_gpu` keys on torch's device view rather than ORT's active device.

### ❓ Open questions (grounded)
1. **onnx/onnx#8068 is still an open/unmerged errata** — the published opset-24 reference impl and spec prose still describe *top-left* alignment; microsoft/onnxruntime#28958 (the actual authority for these tests) merged very recently. The module docstrings cite #8068 as if it were settled spec. A one-line note ("unmerged errata; the authority here is the merged ORT kernel") would prevent confusion.
2. **`test_chunked_prefill_structurally_empty_rows_are_zero` pins behavior in the `nonpad < q_seq` (negative-offset) regime that #8068 itself labels out-of-contract.** It is a deliberate mobius guard (and never arises in production), but a future ORT could legitimately change it without violating #8068. Confirm you want it pinned.

### ✅ Praise
The known-answer probe design (identical Q/K → uniform softmax → mean-of-value-tags, which cleanly separates top-left `1.0` from bottom-right `2.0`, fp16-exact with `atol=0.1`), the convex-combination hull invariant in the chunked-prefill test, the `_ProbeOutcome` four-state enum, and the recursive `If`/`Loop`/`Scan` subgraph scan were all called out as genuinely well-reasoned.

<sub>Synthesized from a five-agent review (readability, correctness, adversarial, spec-deep, cross-module). GPU-gated tests were not executed here; M1/M2 and the open questions are the load-bearing items.</sub>

src/mobius/_builder.py
```diff
@@ -213,29 +213,95 @@ def forward(self, op, input_ids, attention_mask,
...
     return pkg
 
 
+# Attention input index for the optional ``nonpad_kv_seqlen`` operand. This
+# operand (external/static KV cache length) and the TensorScatter op are
+# defined only in opset 24, so a graph using either must not declare opset 23.
+_ATTENTION_NONPAD_KV_SEQLEN_INPUT_INDEX = 6
+
+
+def _apply_opset_lowering(pkg: ModelPackage, execution_provider: str) -> None:
```
justinchuby:
```suggestion
def _maybe_apply_opset_lowering(pkg: ModelPackage, execution_provider: str) -> None:
```

justinchuby: because it doesn't always

justinchuby (approved):

justinchuby (commented):
**Review verdict: 🟡 minor-issues (non-blocking).** Thorough, correctness-first PR — the opset-24 retention logic, the fail-closed/fail-open probe design, and the parity tests are all rigorous and well-reasoned. I traced the probe + tests against the production `components/_attention.py` static branch (Attention input #6 = `nonpad_kv_seqlen`, `is_causal=1`, no `attn_mask`/`past_key`) and they faithfully replicate it. A few non-blocking observations:

**1. Description ↔ code mismatch (no fix needed in code).** Change-summary item **(e)** says *"deleted `tests/_static_cache_support.py`"*, but the diff contains **no deletion** and that file exists on neither `main` nor the head ref (`gh api .../contents/tests/_static_cache_support.py` → 404 on both). The end state is correct (no shim), but the "deleted" claim is misleading — worth correcting the PR body so reviewers aren't hunting for a phantom removal.

**2. Nice latent-bug fix.** Routing the build-side `_apply_opset_lowering` through the recursive `_graph_requires_opset24` (and adding the per-submodel retention) closes a real hole: the *previous* `build_from_module` lowering had no TensorScatter/nonpad guard at all, so with `MOBIUS_ORT_LOWER_OPSET_FOR_EP=1` on a non-default EP it would have emitted an invalid opset-23 graph carrying `TensorScatter`. Latent only (flag default is `False`), but a genuine correctness improvement, and the build/inference gates now agree.

**3. Probe value-mismatch classification** (inline) — a wrong-but-not-top-left output is silently `NEEDS_FIX`; see inline note.

**4. Logger-severity restore nit** (inline).

Nothing here blocks merge; the gating story (probe-gated skips, self-enabling once ORT #28958 is pinned) is sound and CI-safe. 👍

src/mobius/_testing/ort_capabilities.py
```diff
@@ -0,0 +1,456 @@
...
+                    "absent from the probe session providers (%s); treating the "
+                    "static-cache CUDA path as unsupported.",
+                    session.get_providers(),
+                )
+                return _ProbeOutcome.PROBE_ERROR
+            try:
+                attn_output = session.run(None, _probe_feeds())[0]
+            except _EXPECTED_REJECT_ERRORS as exc:
+                return _classify_run_error(exc)
+            if not _probe_output_is_correct(attn_output):
```
justinchuby, line 391: Non-blocking nit on the fail-closed/loud contract. This branch maps **any** non-`2.0` output to `NEEDS_FIX` (logged at `debug`). The known-answer cleanly separates the *historical top-left* fallback (`1.0`) from correct (`2.0`), but it does not distinguish a top-left fallback from a genuine post-#28958 numerical regression (e.g. NaN/Inf, or a different wrong value): both would silently skip the whole suite as "needs #28958" rather than surfacing loudly like `PROBE_ERROR` does. Consider gating this `NEEDS_FIX` on the value actually being the top-left answer (`~_PROBE_VALUE_TAGS[0]`) and/or asserting `np.isfinite(attn_output).all()`, routing anything else to `PROBE_ERROR` at WARNING — so a real regression on a fixed build stays loud.

tests/static_cache_flash_e2e_test.py
```diff
@@ -0,0 +1,433 @@
...
+    saved_stderr_fd = os.dup(2)
+    with tempfile.TemporaryFile(mode="w+b") as capture_file:
+        try:
+            os.dup2(capture_file.fileno(), 2)
+            yield captured
+        finally:
+            # Flush C++ stderr, restore the real fd, then read back what we caught.
+            os.dup2(saved_stderr_fd, 2)
+            os.close(saved_stderr_fd)
+            ort.set_default_logger_severity(2)
```
justinchuby: Minor global-state nit: this restores the ORT default logger severity to a hard-coded `2` (WARNING) rather than the value that was in effect before the `set_default_logger_severity(0)` on entry. If a caller had configured a different default severity, it is silently reset. Capturing the prior severity on entry and restoring it here would make the context manager fully transparent.

src/mobius/_builder.py
titaiwangms: Addressed in 46c8506 — renamed `_apply_opset_lowering` → `_maybe_apply_opset_lowering` at the definition, the `build_from_module` call site, and all test references. Thanks!

tests/static_cache_flash_e2e_test.py
titaiwangms: Addressed in 46c8506 — the ORT default logger severity is now saved and restored through `_set_ort_logger_severity` so the global state is left as found. Thanks!

tests/static_cache_flash_e2e_test.py
```diff
@@ -0,0 +1,453 @@
...
+# call is invisible to the shadow (an unavoidable consequence of the missing getter).
+_ort_logger_severity = _ORT_DEFAULT_LOGGER_SEVERITY
+
+
+def _set_ort_logger_severity(severity: int) -> int:
+    """Set ORT's global default logger severity, returning the previous value."""
+    global _ort_logger_severity
+    prior = _ort_logger_severity
+    ort.set_default_logger_severity(severity)
+    _ort_logger_severity = severity
```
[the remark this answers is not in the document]
titaiwangms: This is a false positive: `_ort_logger_severity` **is** used — it's read inside `_set_ort_logger_severity` (`prior = _ort_logger_severity` at line 249) to save/restore the prior severity, after a `global _ort_logger_severity` declaration. Static analysis misses the read because it's rebound in the same function. Resolving as not-a-defect.

src/mobius/_testing/ort_capabilities.py
titaiwangms: Partially addressed: the probe now distinguishes the expected top-left fallback (~1.0, logged at debug) from any other wrong value, which is surfaced **loudly at WARNING** ('possible post-#28958 regression or ORT drift; investigate'). We did **not** route the unexpected case to `PROBE_ERROR` or add an explicit `np.isfinite` assertion — it still maps to `NEEDS_FIX` (suite skips) rather than erroring. Leaving this thread open for your call: is the WARNING-level loud path sufficient, or would you prefer routing non-top-left/non-finite results to `PROBE_ERROR`? Happy to make that change.

titaiwangms (commented):
## Review summary — static-cache Flash enablement / probe / CI

Reviewed at `d95460c` by a multi-model review team plus independent verification. **Clean, well-gated infrastructure; no Critical/Major findings specific to this PR. Mergeable.**

### Verified correct ✅
- **Recursive opset-24 guard** (`_graph_requires_opset24` + `_maybe_apply_opset_lowering`): correctly detects `TensorScatter` *and* `Attention` with a non-empty input #6 (`nonpad_kv_seqlen`) inside `If`/`Loop`/`Scan` bodies, decides per sub-model, and errs conservatively — an empty-name optional placeholder only *keeps* opset 24, never wrongly lowers to 23. The default-off flag means no shipped export changes.
- **Capability probe** (`supports_static_cache_flash`): genuinely fail-closed. The known-answer value check (`2.0`) closes the latent CPU-fallback fail-open where a CUDA build that declines the node would silently run on CPU with wrong (top-left) values; the `_ProbeOutcome` enum cleanly separates the expected pre-#28958 reject from unexpected probe errors (logged with `exc_info`). Strong unit-test coverage of the classifier.
- New GPU tests skip automatically unless the installed ORT can actually run the path, so CI stays green today and flips green automatically once an ORT release with microsoft/onnxruntime#28958 is installed — no code change needed.

### Notes (no action required for this PR)
- This PR enables the **maskless `is_causal=1`** path; the **bias-aware** sibling (#367) and the genai consumer (microsoft/onnxruntime-genai#2235) edit parallel, non-conflicting structures — confirmed compatible across branches.
- The recursive guard is future-proof but currently only matches `TensorScatter` / `Attention`-input#6; if other opset-24-only default-domain ops are introduced into these graphs later, the predicate set will need extending (a lowered-to-23 graph would otherwise be silently invalid).

Praise: the probe design (functional runtime check + known-answer value gate, not a version-string check) is exactly the right pattern for a capability that depends on an unreleased ORT kernel, and the fail-closed-but-loud behavior is well thought through.

Issue #369, opened 2026-06-22
Learnings: PersonaPlex/Moshi full-duplex S2S export + real-time ORT inference (#368)


justinchuby:
## Summary

Tracking/knowledge issue capturing the learnings from adding **PersonaPlex / Moshi**
full-duplex speech-to-speech support to mobius (Kyutai Mimi codec + Moshi temporal/depformer LM),
plus the real-time browser voice-chat demo. Work landed in **#368**.

The goal is to make these (sometimes hard-won) findings discoverable for the next
person who exports a streaming codec / dual-stream speech LM, or who needs ORT
real-time inference in Python.

---

## 1. Architecture (what the model actually is)

- **Mimi codec** (SEANet + codec transformer): encoder `(B,1,T)->(B,8,Tf)`,
  decoder `(B,8,Tf)->(B,1,T)`, 12.5 Hz frame rate, 8 codebooks.
  Decoder uses **ELU residual blocks (not SnakeBeta)** + downsample/upsample.
- **Temporal transformer**: dim 4096, 32 layers, 32 heads, head 128, RoPE θ=1e4,
  3000-frame sliding causal context. **RMSNorm(eps=1e-8, fp32)**, SwiGLU
  (linear_in 4096->2*11264, linear_out 11264->4096). Fused QKV `in_proj_weight (3*4096,4096)`.
  17-channel dual stream: ch0=text, ch1..8=agent audio, ch9..16=user audio,
  per-channel **delays** `[0,0,1,1,1,1,1,1,1,0,1,1,1,1,1,1,1]`.
- **Depformer**: dim 1024, 6 layers, 16 heads, **context 8, no RoPE**,
  `weights_per_step=16` (stacked per-substep linear weights via multi_linear).
  Resets KV **every temporal frame**; 16 substeps generate the 8 agent + 8 user codes.

## 2. Export gotchas (codec + LM)

- **LayerNorm epsilon**: Kyutai/Mimi codec transformer uses `eps=1e-5`
  (`create_norm_fn`), but mobius `LayerNorm` defaults to `1e-6`. The default was
  the single root cause of an early codec parity failure — pass `layer_norm_eps`
  explicitly. *(stored as a repo memory)*
- **Causal Conv1d padding**: Kyutai SEANet causal convs pad
  `left = effective_kernel - stride`, **not** `kernel - 1`. The latter is wrong for
  strided Mimi convs. *(stored as a repo memory)*
[...]

PR #371, opened 2026-06-22
Teach the direct GQA path to emit sliding-window local_window_size

justinchuby:
## Summary

Uniform sliding-window decoders (e.g. **Moshi/PersonaPlex temporal**, `sliding_window=3000`) take the **direct GQA path** in `TextModel.forward`, but that path built a global `GQAContext` that **never set `local_window_size`**. The window was silently dropped and every layer ran *full* causal attention — incorrect for these models, and it forfeits GQA's sliding-window kernel path.

This wires `config.sliding_window` → `GQAContext.local_window_size`, guarded to **uniformly-sliding** models.

## What changed

- **`models/base.py`** — `TextModel` now stores `self.config` and computes `_gqa_local_window_size()`:
  - Returns `config.sliding_window` when set and the schedule is uniform.
  - Returns `-1` (disabled) for full-attention models and for **mixed** `layer_types` (some full, some sliding) — those use custom per-layer `GQAContext`s (Gemma2/3/4, gpt-oss) and can't be expressed by one global window.
- **`attention-optimization` skill** — corrected/expanded:
  - `nonpad_kv_seqlens` is **prefill-only** (ORT *rejects* it together with `past_key`/`past_value`).
  - Opset-24 ONNX `Attention` has **no `past_present_share_buffer`** / no in-place KV buffer (`present = concat(past, new)` each step) — only GQA writes in place. This is the main residual reason GQA out-paces ONNX `Attention` during decode (context for #369).
  - Documented GQA `local_window_size` for sliding windows + the rewrite-path limitation.
- **Tests** — assert the emitted `GroupQueryAttention` carries `local_window_size == sliding_window` for uniform/all-sliding configs, and omits it for full-attention and mixed-`layer_types` configs.

## Why the direct path (not the rewrite)

The post-hoc `RotaryAttentionToGQA` rewrite only sees an already-baked float mask and cannot recover the window size, so windowing must be set on the **direct** GQA path. Uniform-sliding base-`TextModel` models always take the direct path on GQA-capable EPs; the rewrite only fires for non-standard-RoPE/DML cases that emit a padding-only mask. Documented as a known limitation.

## Validation

- **Semantics (standalone, ORT CPU):** proved `local_window_size=W` ⇒ query attends keys `[i-W+1, i]`, **exactly** matching HF `sliding_window=W` (max diff `8.8e-8`).
- **End-to-end (ORT CUDA, fp16):** ORT loads + runs the GQA node with `local_window_size`. Windowed vs full twin: **identical** for `seq ≤ W`, **diverges** for `seq > W` (window actively masking).
- **Moshi temporal:** every GQA node now emits `local_window_size=3000`.
- **Regression:** `1284 passed, 42 skipped` across `build_graph_test.py` + attention/component suites. Non-sliding models are unchanged (helper returns `-1`).
- Lint clean (`lintrunner`).

### DIFF
--- a/.agents/skills/adding-a-new-model/SKILL.md
+++ b/.agents/skills/adding-a-new-model/SKILL.md
@@ -112,6 +112,12 @@ class MyCausalLMModel(CausalLMModel):
         self.lm_head = Linear(config.hidden_size, config.vocab_size, bias=False)
 ```
 
+> If your text model **subclasses `TextModel`** (rather than `nn.Module`) and
+> overrides `__init__` with `nn.Module.__init__(self)` to swap in a custom
+> decoder layer, you must also set `self.config = config` in that subclass —
+> `TextModel.forward` reads `self.config`. See troubleshooting §7.
+
+
 #### Class metadata attributes
 
 Every registered model class should set two class-level attributes:
@@ -395,6 +401,25 @@ differs between loads, suspect `_init_weights` corruption. Set
 `torch.manual_seed(42)` before `from_pretrained` — if that makes outputs
 deterministic, `_init_weights` is the culprit. Then compare specific
 parameters between the loaded model and the safetensors checkpoint.
+
+### 7. `AttributeError: '<X>TextModel' object has no attribute 'config'`
+
+**Symptom:** Building (or running build-graph / GQA rewrite-rule tests for)
+a model raises `AttributeError: '…TextModel' object has no attribute
+'config'` from inside `TextModel.forward` (e.g. `_gqa_local_window_size`).
+
+**Root cause:** The base `TextModel.__init__` sets `self.config = config`,
+and `TextModel.forward` relies on it (sliding-window detection, etc.). A
+`TextModel` **subclass** that overrides `__init__` with
+`nn.Module.__init__(self)` — instead of `super().__init__(config)` — to swap
+in a custom decoder layer must re-establish the contract by setting
+`self.config = config` itself. Forgetting it crashes only that model.
+
+**Fix:** Add `self.config = config` right after `nn.Module.__init__(self)`
+in the subclass `__init__`. Real examples: `Glm4TextModel` (`models/glm.py`),
+`_LoRATextModel` (`models/phi.py`). Do **not** paper over it with
+`getattr(self, "config", None)` in the base — that silently disables
+config-driven features for the offending subclass.
 
 > For additional troubleshooting (gated attention split ordering, DeltaNet
 > scaling, identity node folding, fp32 upcast patterns, multi-token prefill,
--- a/.agents/skills/attention-optimization/SKILL.md
+++ b/.agents/skills/attention-optimization/SKILL.md
@@ -35,7 +35,7 @@ Use this skill when:
 | Pattern | Recommended mask type |
 |---------|----------------------|
 | Simple causal-only | No mask — use `is_causal=1` (enables Flash) |
-| Sliding window (simple model) | Bool mask (precise, less memory) |
+| Sliding window (simple, GQA path) | GQA `local_window_size` (fast path); bool mask only if you must use ONNX Attention |
 | KV-shared layers | Float additive bias |
 | Mixed head_dim (e.g. Gemma4) | Float additive bias |
 | Padding + causal | `nonpad_kv_seqlens` or bool mask |
@@ -280,6 +280,13 @@ use custom per-layer `GQAContext`s instead.
 > cache. Also, the post-hoc GQA rewrite (`RotaryAttentionToGQA`) cannot
 > recover a window from an already-baked float mask, so sliding windows
 > must be set on the **direct** GQA path (GQAContext), not via the rewrite.
+>
+> Decode trade-off: a non-default `local_window_size != -1` disqualifies the
+> dedicated `seq==1` **XQA** decode kernel (which requires `local_window==-1`,
+> see the CUDA cascade table above), so windowed GQA decode falls back to
+> Flash-decode/MEA. This is still on the fast GQA path and is the correct
+> trade for models that genuinely need the window; do not set it on
+> full-attention models.
 
 ## Key takeaways for model builders
 
--- a/.agents/skills/attention-optimization/SKILL.md
+++ b/.agents/skills/attention-optimization/SKILL.md
@@ -23,8 +23,8 @@ Use this skill when:
 | Scenario | Recommended | Why |
 |----------|------------|-----|
 | Causal only | `attn_mask=None` + `is_causal=1` | Enables Flash (fastest for prefill) |
-| Padding (batch>1) | `nonpad_kv_seqlens` (best) or bool mask | `nonpad_kv_seqlens` enables Flash + shared buffer with no mask |
-| Sliding window (simple) | Bool mask | Equally precise as float, uses less memory |
+| Padding (batch>1) | `nonpad_kv_seqlens` (prefill) or bool mask | `nonpad_kv_seqlens` enables Flash with no mask, but is rejected with past KV (decode) |
+| Sliding window (simple) | GQA `local_window_size` or bool mask | `local_window_size` keeps the fast GQA path; bool mask if you need ONNX Attention |
 | Complex (sliding+KV-shared+dual head_dim) | Float additive bias | Avoids mask construction bugs in multi-constraint patterns |
 | Custom pattern | Float additive bias | Arbitrary values |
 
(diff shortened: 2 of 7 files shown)


justinchuby:
@titaiwangms

titaiwangms (commented):
## Review: Wire `config.sliding_window` → GQA `local_window_size`

Reviewed by a 5-model review team (deep / critical / code / integration / readability) plus independent verification against the ORT kernel source, HF transformers, and live test runs at PR head `039b325`. **The core change is correct and well-tested.** One config-robustness issue and one path-divergence issue are worth addressing before/after merge.

### ✅ Verified correct (the load-bearing claim)
`local_window_size=W` ⇔ HF `sliding_window=W` is **bit-exact**, confirmed against three independent authorities:
- **HF** `masking_utils.py`: `kv_idx > q_idx - W` ∧ causal ⇒ keys `[i-W+1, i]` (W tokens incl. current).
- **ORT CPU** `gqa_attention_base.h`: masks `t < (i+1) - W` ⇒ keeps `[i-W+1, i]`; header comment: *"The window size includes new token. It only includes tokens on the left side."*
- **ORT CUDA** `group_query_attention_impl.cu:891`: passes `W-1` as flash left-window ⇒ same `[i-W+1, i]`.

Also confirmed the window is a **bit-exact no-op for seq ≤ W** (CPU kernel gates on `seq>W`), so Mistral/Moshi short-sequence goldens stay valid. Empirically: windowed twin `max|Δlogits| = 0.00e+00`; full-attention twin `= 2.3e-2`. The new `test_window_confines_receptive_field` (seq=24 > W=4) is exactly the right kind of test — it actually exercises masking, unlike the short-seq parity golden. The `L*(W-1)` receptive-field bound is correct.

Tests pass: `6 passed` (`-k "GQASliding or temporal_gqa_emits or window_confines"`).

---

### 🟠 Major — uniformity guard silently depends on HF having normalized the config
`_gqa_local_window_size()` decides uniformity from only `sliding_window` + `layer_types`, but the Qwen2/Qwen3 family expresses "window present in config but **not applied**" through other fields:
- `Qwen2Config.__post_init__`: `sliding_window = sliding_window if use_sliding_window else None` (default `use_sliding_window=False` → HF nulls it).
- `max_window_layers`: sliding applied only to upper layers.

`ArchitectureConfig.from_transformers` (`_configs/_base.py:578`) reads `sliding_window` **unconditionally** and never consults `use_sliding_window`/`max_window_layers`. So this PR is correct *only because* modern HF `__post_init__` already nulled the field. **Reachable bypass:** the raw-`config.json` fallback (`_config_resolver.py` → `_dict_to_pretrained_config` → generic `transformers.PretrainedConfig(**d)`) does **not** run Qwen2's `__post_init__`, so a config with `sliding_window=32768, use_sliding_window=false, layer_types=None` would emit `local_window_size` on **every** layer and silently cripple long-context attention. (No live regression for canonical Qwen2 today, since AutoConfig nulls it — hence Major, though one reviewer rated this Critical given Qwen2's traffic.)

**Suggested fix (cheap, removes the latent hole):** in `from_transformers`, set `sliding_window=None` when `getattr(config, "use_sliding_window", True) is False`; optionally derive `layer_types` from `max_window_layers`. Add a Qwen2-style regression test (`sliding_window=4096, use_sliding_window=False, layer_types=None` → attribute omitted) so the invariant is *enforced*, not *inherited*.

### 🟠 Major — EP/dtype path divergence
The window is wired only in the `if use_gqa:` branch (fp16+cuda). The fallback `else:` branch (CPU fp32, DML, non-fused-RoPE) builds only `create_padding_mask` — **full causal, no window**. So a sliding-window model's attention pattern now silently depends on the export target: fp16/cuda correct, other EPs wrong for `seq > W`. This is partly pre-existing (before this PR *all* paths dropped the window), but the PR makes the divergence live. Recommend either masking the fallback path with the sliding window, or emitting a loud warning/error when exporting a uniform-sliding model through a path that can't represent it. (Reasonable as a tracked follow-up if out of scope here.)

---

### 🟡 Minor
1. **Guard edge cases** (`base.py:192`): `any(t != "sliding_attention" for t in [])` is `False`, so an empty `layer_types` returns the window; a short/partial list (`len < num_hidden_layers`) is also treated as uniform. Harden to `if layer_types and len(layer_types) == config.num_hidden_layers and all(t == "sliding_attention" ...)`.
2. **No Mistral resolver→emit test.** Mistral-7B-v0.1 is the highest-traffic real model affected (genuinely uniform `sliding_window=4096` → correct). A one-line test asserting the resolver→GQA wiring emits `local_window_size=4096` would lock down the most-used path.
3. **fp16 sensitivity threshold** (`test_window_confines_receptive_field`): full-attn assert `>1e-2` currently passes at `2.3e-2` (~2.3× margin). The windowed half is bit-exact (0.0); only this half has finite margin — consider a larger window/seq or lower threshold for robustness. Non-blocking.
4. **SKILL.md sub-table is now self-contradictory:** the "Bool mask vs float bias → When to use each" sub-table still says `Sliding window (simple) | Bool mask`, contradicting the new top-table + takeaway #8 recommending `local_window_size`. A reader hitting that section first gets the old answer.
5. **SKILL.md omits the XQA decode penalty:** setting `local_window_size != -1` disables the XQA decode kernel (per the doc's own dispatch table), so decode falls to Flash/MEA. Worth one sentence in the new note. (For Moshi the measured latency impact was negligible, but it's operationally relevant in general.)

### ⚪ Nits
- `int(sliding_window)` silently truncates a float; consider validating integer-ness at config ingestion.
- `getattr(self.config, "sliding_window", None)` is defensive though `ArchitectureConfig.sliding_window` is a declared field — harmless.
- Docstring could note *why* the sentinel is `-1` (ORT's GQA "full causal" convention), not `None`.
- "weight-agnostic" in the test docstring applies only to the windowed half; the sensitivity half depends on seed 1234.
- Redundant `self.config = config` in several subclasses now that the base sets it (harmless, same object).
- SKILL.md wording: "IO-bound" → "aliased to the same device buffer"; "residual decode edge" → "remaining decode advantage".

### Heads-up (not this PR's bug)
ORT's **unfused** CUDA fallback kernel (`unfused_attention.cu:130`, `start = q_pos - W`) attends to **W+1** keys `[i-W, i]`, vs Flash/CPU's **W** keys — an ORT-internal off-by-one between kernels. Only the rare unfused fallback path (head_size>256, etc.) is affected, and it's a no-op for seq ≤ W, so it doesn't affect this PR's targets. Possibly worth an upstream ORT issue.

---

**Bottom line:** math/parity is correct and well-grounded; the function is clean and the tests are meaningful. Main asks: (1) honor `use_sliding_window` in the resolver + pin it with a Qwen2 regression test, and (2) decide how to handle the non-GQA-path divergence (fix or warn). Nice work keeping uniform-window models on the fast GQA path. 🚀

justinchuby:
Thanks for the thorough review — addressed in f741320:

**Major 1 — honor `use_sliding_window`:** `from_transformers` now routes `sliding_window` through a new `_resolve_sliding_window()` helper that nulls the window when `use_sliding_window is False`. This closes the hole where the raw `config.json` resolver (which bypasses HF's `__post_init__`) would leak a window onto Qwen2/Qwen3 models that keep a non-null `sliding_window` while disabled. Models without the flag (Mistral) default to `True` and are unaffected. Added resolver tests for True/False/absent.

**Major 2 — non-GQA path divergence:** Added a `logger.warning` in `TextModel.forward`'s non-GQA `else` branch (static cache / mRoPE / NoPE / non-fused RoPE) when the model declares a uniform window that path can't represent. Note: CPU fp32 actually takes the GQA path, so the uncovered cases are narrower than the static-cache/mRoPE branches — the warning flags those. Full fallback masking on those paths is left as a follow-up. Test exercises the static-cache trigger.

**Minor 1 — guard hardening:** `_gqa_local_window_size` now requires `layer_types` to be non-empty AND `len == num_hidden_layers` AND all `sliding_attention`; empty/partial schedules are treated as non-uniform (no longer vacuously accepted by `all()`). Added empty/partial edge-case tests.

**Minor 4/5 — SKILL.md:** Fixed the contradictory bool-vs-float sub-table (sliding window → GQA `local_window_size` on the fast path) and added the XQA decode trade-off note (`local_window_size != -1` disqualifies the seq==1 XQA kernel).

**Nit:** Added a docstring line explaining `-1` is ORT's documented sentinel for full-causal.

Re `max_window_layers` (partial-window upper-layer schedule): left as a documented limitation — it only matters when `use_sliding_window is True` with `max_window_layers < num_hidden_layers`, which doesn't occur in the models currently routed here. The uniformity guard keeps it safe (full attention) rather than wrong.

All 12 window tests pass; lint clean.

titaiwangms (approved):

justinchuby:
Wait for my personaplex test

PR #372, opened 2026-06-23
Add rank-4 RMSNormalization reshape rule for QNN HTP

shreyshah-microsoft:
## What
A rewrite rule that reshapes rank-4 `RMSNormalization` (query/key norm) to rank-3 for the QNN HTP, which miscomputes RMSNormalization over the last axis of a rank-4 tensor.

## Why
Gemma-4 (and Qwen3) apply RMSNormalization over the head dimension after reshaping q/k to `(batch, seq, heads, head_dim)`, so the norm runs on a rank-4 tensor. On the Hexagon HTP this computes incorrectly -- the graph finalizes but the q/k-norm output is numerically wrong (uncorrelated with CPU); rank-3 RMSNormalization runs correctly. Pairs with the QNN EP profile (#370): that targets the HTP, this makes the q/k-norm graph compute correctly there.

## How
`ReshapeRank4RMSNorm` reshapes `(B, S, H, Dh)` -> `(B, S*H, Dh)`, applies the identical RMSNormalization over the last axis, and reshapes back. Both reshapes use constant shapes (`[0, -1, head_dim]`, `[0, -1, heads, head_dim]`) so no `Shape` op is introduced and the graph stays static for the HTP. Numerically exact -- the normalized axis is unchanged. Gated on a new `supports_rank4_rmsnorm` capability (default `True`; `False` only for `qnn`), applied in the lowering stage alongside SeparateRoPE.

## Test
`_reshape_rmsnorm_test.py`: the rule reshapes a rank-4 RMSNorm to rank-3 (2 Reshapes added) and the rewritten graph matches the original on CPU (where rank-4 RMSNorm is correct); a rank-3 RMSNorm is left untouched.

### DIFF
--- a/src/mobius/_execution_providers.py
+++ b/src/mobius/_execution_providers.py
@@ -69,7 +69,7 @@ class EpCapabilities:
             Leave ``True`` for CUDA / DML EPs that ship the native kernel.
         supports_rank4_rmsnorm: ``False`` reshapes rank-4 ``RMSNormalization``
             (query/key norm over the head dimension) to rank-3 and back via
-            ReshapeRank4RMSNorm.  ``True`` leaves it unchanged.  Set ``False``
+            HtpRank4RMSNorm.  ``True`` leaves it unchanged.  Set ``False``
             only for the QNN HTP, which miscomputes rank-4 RMSNormalization.
         default_int4_accuracy_level: Default accuracy level for INT4
             quantization (0 = highest accuracy, 4 = fastest).
--- a/src/mobius/_execution_providers.py
+++ b/src/mobius/_execution_providers.py
@@ -67,6 +67,10 @@ class EpCapabilities:
             ``PackedMultiHeadAttention`` via InlinePass to a
             block-diagonal attention bias + standard ``Attention``.
             Leave ``True`` for CUDA / DML EPs that ship the native kernel.
+        supports_rank4_rmsnorm: ``False`` reshapes rank-4 ``RMSNormalization``
+            (query/key norm over the head dimension) to rank-3 and back via
+            ReshapeRank4RMSNorm.  ``True`` leaves it unchanged.  Set ``False``
+            only for the QNN HTP, which miscomputes rank-4 RMSNormalization.
         default_int4_accuracy_level: Default accuracy level for INT4
             quantization (0 = highest accuracy, 4 = fastest).
         provider_options: Default ORT GenAI provider options dict for this EP.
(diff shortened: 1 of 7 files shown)


src/mobius/rewrite_rules/_reshape_rmsnorm_test.py
```diff
@@ -0,0 +1,67 @@
...
+    node = helper.make_node("RMSNormalization", ["x", "weight"], ["y"], epsilon=1e-6)
+    graph = helper.make_graph(
+        [node], "rmsnorm",
+        [helper.make_tensor_value_info("x", TensorProto.FLOAT, shape)],
+        [helper.make_tensor_value_info("y", TensorProto.FLOAT, shape)],
+        [weight],
+    )
+    model = helper.make_model(graph, opset_imports=[helper.make_opsetid("", 23)])
+    model.ir_version = 10
+    return model
```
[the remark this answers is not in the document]
kunal-vaishnavi: I agree with this.

src/mobius/_execution_providers.py
```diff
@@ -67,6 +67,10 @@ class EpCapabilities:
             ``PackedMultiHeadAttention`` via InlinePass to a
             block-diagonal attention bias + standard ``Attention``.
             Leave ``True`` for CUDA / DML EPs that ship the native kernel.
+        supports_rank4_rmsnorm: ``False`` reshapes rank-4 ``RMSNormalization``
+            (query/key norm over the head dimension) to rank-3 and back via
+            ReshapeRank4RMSNorm.  ``True`` leaves it unchanged.  Set ``False``
+            only for the QNN HTP, which miscomputes rank-4 RMSNormalization.
```
kunal-vaishnavi, line 73: Is this just for QNN or for other hardware as well?

src/mobius/rewrite_rules/_reshape_rmsnorm_test.py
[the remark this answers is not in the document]
shreyshah-microsoft: Rebuilt the fixture with onnx_ir (`ir.Value` / `ir.Graph` / `GraphBuilder`, `op.Constant(value=ir.tensor(...))` for the weight) and serialize via `ir.serde.serialize_model` for ORT. Removed `onnx.helper`/`TensorProto`. Behavior should be unchanged.

src/mobius/_execution_providers.py
shreyshah-microsoft: QNN-specific. It's an HTP hardware miscompute of rank-4 `RMSNormalization` (the q/k-norm over the head dimension), not a general correctness issue. CPU/CUDA/DML/OpenVINO all compute rank-4 RMSNorm correctly, so `supports_rank4_rmsnorm` stays `True` by default and only the qnn profile sets it `False`.

kunal-vaishnavi (approved):

src/mobius/rewrite_rules/_htp_rank4_rmsnorm.py
justinchuby, line 1: Can you name this rule to have qualcomm hexagon htp specific names so that we know it is not generic? Is this going to be true for all qnn versions, or is this expected to be removed sometime?

PR #375, opened 2026-06-25
Load safetensors from local checkpoint directories

xiaoyu-work:
## Summary
- load model.safetensors or model.safetensors.index.json from a local checkpoint directory
- avoid HuggingFace Hub calls for local directories
- reject absolute/path-traversal shard names before reading local weights

## Validation
- PYTHONPATH=$PWD/src conda run -n olive python -m pytest src/mobius/_weight_loading_test.py -q -k 'LocalSafetensorsLoading'


justinchuby:
Could you check review comments? Thanks

justinchuby (approved):

src/mobius/_weight_loading.py
```diff
@@ -22,18 +22,23 @@
 import concurrent.futures
 import json
 import logging
+from pathlib import Path, PurePosixPath
```
justinchuby: Can you update this to be module only? Use pathlib.Path etc below

xiaoyu-work: @copilot fix this

justinchuby:
@copilot+claude-sonnet-4.6 fix lint

justinchuby (approved):

PR #376, opened 2026-06-25
Support Olive-packed quantized weights

xiaoyu-work:
## Summary
- recognize Olive quantization configs that use the symmetric field
- preprocess Olive uint8 qweight/qzeros tensors into MatMulNBits names and layout
- apply Olive preprocessing during mobius.build weight loading

## Validation
- PYTHONPATH=$PWD/src conda run -n olive python -m pytest src/mobius/_weight_utils_test.py -q -k 'PreprocessOliveWeights'


justinchuby (approved):

src/mobius/_weight_utils.py
```diff
@@ -658,6 +658,44 @@ def preprocess_gptq_weights(
...
+    """
+    blob_size = group_size * bits // 8
+    result: dict[str, torch.Tensor] = {}
+
+    for key, value in state_dict.items():
+        if key.endswith(".qweight"):
+            if value.dtype != torch.uint8:
+                raise ValueError(f"Olive qweight must be uint8 for {key}, got {value.dtype}")
+            if value.shape[-1] % blob_size != 0:
+                raise ValueError(
```
justinchuby, line 731: The loop renames every `*.qweight` → `*.weight`, including `model.embed_tokens.qweight` and `lm_head.qweight`. This silently mishandles the **RTN-quantized embedding/lm_head case**: the full INT4 Olive recipe (`tencent-HY-MT1.5-1.8B/cuda/..._int4.json`) runs `rtn` with `embeds: true, lm_head: true`, producing `model.embed_tokens.{qweight,qzeros,scales}`. Renaming `embed_tokens.qweight → embed_tokens.weight` feeds a uint8-packed `[N, n_blocks, blob_size]` tensor into the Embedding's `Gather`, which expects a float `[vocab, hidden]` table → silent corruption (quantized lookup would need `GatherBlockQuantized`).

This PR correctly handles the **body-quantized (gptq) case** where embeds/lm_head stay fp16. Suggest either:
1. Guard against it — `raise` a clear error if `*.embed_tokens.qweight` / `lm_head.qweight` are present (until `GatherBlockQuantized` is supported), or
2. Document explicitly that only body-quantized Olive models are supported (RTN-quantized embeddings/lm_head are not).

Verified empirically against a real MT2 Olive `gptq`→`rtn` output: the `rtn` stage emits `embed_tokens.qweight/qzeros/scales`, so a full-pipeline model fed to `mobius.build` would hit this path.

justinchuby: Addressed in 73fc3d4 (pushed to this branch). The RTN-quantized embedding/lm_head gap is now closed:

- New `QuantizedEmbedding` component emits `GatherBlockQuantized` (com.microsoft) and consumes Olive's 2-D uint8 `embed_tokens.qweight` directly — no more feeding a packed tensor into a plain `Gather`.
- `QuantizationConfig` parses `embeds`/`lm_head` → `quantize_embeddings`/`quantize_lm_head`, and reads `symmetric` as a fallback for `sym`.
- `preprocess_olive_weights` keeps the embed table 2-D, renames `qzeros`→`zero_points` (uint8-validated), and synthesises the tied quantized `lm_head` (3-D MatMulNBits layout) from the shared packed bytes. The float<->head Parameter alias is now only applied when both sides are unquantized.

Verified end-to-end on Hy-MT2-1.8B: `mobius build --config <gptq+rtn out> --dtype f16 --ep cuda` → 1 GatherBlockQuantized + 225 MatMulNBits + 32 GQA; the int4 model translates correctly via ORT GenAI CUDA. +18 unit tests.

justinchuby:
@copilot fix lint

justinchuby (approved):

PR #396, opened 2026-07-13
Fix GGUF→ONNX Q4 conversion for Qwen2/Qwen3 (missing biases + wrong permute)

justinchuby:
Converted Qwen2.5 Q4 GGUF models emitted garbage despite a correct MatMulNBits weight repack (layer-0 dequant matched GGUF exactly, max_abs_diff=0.0). Two independent bugs:

1. **Missing attention QKV biases** — `gguf_to_config` never inferred projection-bias flags, so `attn_qkv_bias` defaulted to False. Qwen2 carries large q/k/v biases (q≈79, k≈130); omitting the bias `Add` wrecks attention. Added tensor-presence inference (`_infer_attn_qkv_bias/_infer_attn_o_bias/_infer_mlp_bias`).
2. **Spurious/mis-shaped Q/K reverse-permute** — Qwen2/Qwen3 use NEOX rope and are NOT permuted by llama.cpp, but were mapped to the Llama interleaved-rope reverse-permute. `_reverse_permute` also used the forward reshape `(n_head,2,dim)` instead of the inverse `(n_head,dim,2)` (only correct at head_dim=4). Gated the permute on model_type (llama/mistral only) and fixed the reshape.

**Verified:** converted Qwen2.5-0.5B Q4 now produces coherent output ("The capital of France is Paris."); isolation confirmed both bugs are independently fatal. `pytest src/mobius/integrations/gguf/` → 151 passed.

### DIFF
--- a/src/mobius/integrations/gguf/_builder.py
+++ b/src/mobius/integrations/gguf/_builder.py
@@ -470,6 +470,7 @@ def _load_quantized_state_dict(
 
     num_heads = getattr(config, "num_attention_heads", None)
     num_kv_heads = getattr(config, "num_key_value_heads", None)
+    model_type = getattr(config, "model_type", None)
 
     # Detect Tencent's non-mainline Q1_0 layout once per file. Reading
     # such tensors requires a custom parser keyed on the explicit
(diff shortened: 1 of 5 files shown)


justinchuby:
@copilot+claude-sonnet-5 fix lint

PR #402, opened 2026-07-13
Emit MatMulNBits accuracy_level=4 (int8 path) — 2.3x CPU decode

justinchuby:
**Root cause of onnx-genai being ~3.6x slower than llama.cpp on CPU: MatMulNBits was emitted with no `accuracy_level`**, so ORT's MLAS kernel ran the fp32 dequant+GEMM path instead of the int8 dynamic-quant + int8 dot-product path (ARM SDOT / x86 AVX-VNNI) — the same class of kernel llama.cpp uses. ORT's parity claim holds; we just weren't opting in.

`default_int4_accuracy_level=4` already existed on the cpu/webgpu `EpCapabilities` but was **dead config**. This plumbs it via a new `_accuracy_level_attrs()` helper into both `MatMulNBits` emission sites in `components/_quantized_linear.py` (Q/K/V/O, MLP, tied + non-tied head). Emits when `>0`, omits at `0` (portable default preserved).

### Measured (Qwen2.5-0.5B Q4, CPU EP, decode tok/s, all coherent)
| config | tok/s |
|---|---:|
| baseline (accuracy_level missing) | 39.3 |
| **accuracy_level=4 (int8)** | **91.8** (2.33x) |
| quantized head + acc4 | **194.7** |
| ref: LM Studio CPU | 157 |

fp16/bf16 levels regress on M1 (no native GEMM). End-to-end verified: `--ep cpu` stamps all 168 nodes with accuracy_level=4; `--ep default` omits it.

lintrunner clean; pytest 518 passed (+2 tests).

### DIFF
--- a/src/mobius/components/_quantized_linear.py
+++ b/src/mobius/components/_quantized_linear.py
@@ -28,8 +28,13 @@ from mobius._build_context import ep_capabilities
 _MICROSOFT_DOMAIN = "com.microsoft"
 
 
-def _accuracy_level_attrs() -> dict[str, int]:
+def _accuracy_level_attrs(bits: int) -> dict[str, int]:
     """Return the ``accuracy_level`` attribute for ``MatMulNBits``, if any.
+
+    Only emitted for 4-bit weights: ``accuracy_level`` is sourced from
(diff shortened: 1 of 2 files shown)


justinchuby:
@jambayk could you review? Thanks!

PR #410, opened 2026-07-17
Add NVIDIA NeMo Sortformer speaker-diarization model

themason2011:
## Summary

Adds support for the NVIDIA NeMo **Sortformer streaming speaker-diarization** model (`nvidia/diar_streaming_sortformer_4spk-v2.1`) to mobius.

This is not a HuggingFace `transformers` architecture — it ships as a NeMo `.nemo` checkpoint with no `config.json`. The architecture was reconstructed from the checkpoint and NeMo source. The exported ONNX graph implements the **offline forward path** (`frontend_encoder` + `forward_infer`): mel-spectrogram features in, per-frame speaker-activity probabilities out.

## Architecture

- **FastConformer encoder** (NeMo `ConformerEncoder`): rel_pos Transformer-XL attention, 8x `dw_striding` conv subsampling, Macaron feed-forward blocks, batch-norm convolution module, xscaling.
- **Encoder projection** (512 -> 192) + **post-LayerNorm Transformer encoder** (18 layers).
- **Speaker sigmoid head** (`relu -> Linear -> relu -> Linear -> sigmoid`).

Module attribute names mirror the NeMo weight prefixes, so `preprocess_weights` only drops unused keys (mel preprocessor, `num_batches_tracked`, the streaming FIFO head). The mel preprocessor (STFT) is left out of the graph, matching the mobius audio convention.

## Changes

- `models/sortformer.py` — `SortformerConfig` (+ `from_nemo_yaml`), `SortformerDiarizationModel`, and a `build_sortformer(.nemo)` loader that extracts the archive, builds the graph, and applies checkpoint weights.
- `tasks/_diarization.py` — `DiarizationTask` (`input_features -> speaker_probs`), registered as the `diarization` task.
- Exports wired into `models/__init__.py` and the task registry.

## Testing

- **Unit tests** (`tests/build_graph_test.py::TestBuildGraphSortformer`): tiny-config graph build, I/O contract, initializers, task lookup, and an ORT forward run. All pass.
- **Integration parity** (`tests/sortformer_integration_test.py`, `-m integration`): compares the ONNX output against the real NeMo PyTorch reference. Offline output matches to **1.2e-7 (float32)**.
- ruff check + format clean.


src/mobius/models/__init__.py
```diff
@@ -13,6 +13,7 @@
     "BertModel",
     "Blip2Model",
     "BloomCausalLMModel",
+    "build_sortformer",
```
justinchuby: Looks like we don't need to expose this to public under `models.`?
```suggestion
```

src/mobius/models/sortformer.py
```diff
@@ -0,0 +1,736 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Low-level ONNX helpers
+# ---------------------------------------------------------------------------
+
+
+def _swish(op: OpBuilder, x: ir.Value) -> ir.Value:
+    """Swish / SiLU activation: ``x * sigmoid(x)``."""
+    return op.Mul(x, op.Sigmoid(x))
```
justinchuby: op.Swish is defined for opset 24 which we use (https://onnx.ai/onnx/operators/onnx__Swish.html). You may use it directly

src/mobius/models/sortformer.py
```diff
@@ -0,0 +1,736 @@
...
+        self.linear_pos = Linear(d_model, d_model, bias=False)
+        self.pos_bias_u = nn.Parameter([num_heads, self._head_dim])
+        self.pos_bias_v = nn.Parameter([num_heads, self._head_dim])
+
+    def _to_heads(self, op: OpBuilder, x: ir.Value) -> ir.Value:
+        # [B, T, D] -> [B, T, H, d_k] -> [B, H, T, d_k]
+        x = op.Reshape(x, [0, 0, self._num_heads, self._head_dim])
+        return op.Transpose(x, perm=[0, 2, 1, 3])
+
+    def _rel_shift(self, op: OpBuilder, x: ir.Value, t_dim: ir.Value) -> ir.Value:
```
justinchuby, line 355: This seems complicated. Potentially an optimization opportunity? Otherwise it looks good if we don't have better ops to do this

src/mobius/models/sortformer.py
```diff
@@ -0,0 +1,736 @@
...
+        bias_v = op.Unsqueeze(self.pos_bias_v, [0, 2])
+        q_u = op.Add(q, bias_u)
+        q_v = op.Add(q, bias_v)
+
+        matrix_ac = op.MatMul(q_u, op.Transpose(k, perm=[0, 1, 3, 2]))  # [B,H,T,T]
+        matrix_bd = op.MatMul(q_v, op.Transpose(p, perm=[0, 1, 3, 2]))  # [B,H,T,2T-1]
+        matrix_bd = self._rel_shift(op, matrix_bd, t_dim)  # [B,H,T,T]
+
+        scale = op.Constant(value=ir.tensor(np.array(self._head_dim**-0.5, dtype=np.float32)))
+        scores = op.Mul(op.Add(matrix_ac, matrix_bd), scale)
```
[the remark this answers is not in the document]
justinchuby: Is this relevant @themason2011

src/mobius/models/sortformer.py
```diff
@@ -0,0 +1,736 @@
...
+        self.query_net = Linear(d_model, d_model)
+        self.key_net = Linear(d_model, d_model)
+        self.value_net = Linear(d_model, d_model)
+        self.out_projection = Linear(d_model, d_model)
+
+    def _to_heads(self, op: OpBuilder, x: ir.Value) -> ir.Value:
+        x = op.Reshape(x, [0, 0, self._num_heads, self._head_dim])
+        return op.Transpose(x, perm=[0, 2, 1, 3])
+
+    def forward(self, op: OpBuilder, x: ir.Value) -> ir.Value:
```
justinchuby, line 534: Can the `Attention-24` op be used here?

src/mobius/models/sortformer.py
```diff
@@ -0,0 +1,736 @@
...
+            renamed[key] = value
+        return renamed
+
+
+# ---------------------------------------------------------------------------
+# Loader
+# ---------------------------------------------------------------------------
+
+
+def build_sortformer(
```
justinchuby:
```suggestion
```

Please check the nemo integration on how nemo is handled. The build function should live outside of the modeling code.

src/mobius/models/sortformer.py
themason2011: The `_rel_shift` complexity is inherent to the Transformer-XL relative-position mechanism NeMo uses (`RelPositionMultiHeadAttention.rel_shift`). There isn't a single ONNX op for it — the left-pad → reshape → drop-row → reshape → slice sequence is the canonical way to realign the `[B, H, T, 2T-1]` positional scores into `[B, H, T, T]` so entry `(i, j)` holds the bias for relative distance `i - j`. Kept as-is since it maps 1:1 to the reference and offline parity is bit-close to NeMo (max abs diff 1.2e-7). Happy to revisit if a fused op becomes available.

src/mobius/models/sortformer.py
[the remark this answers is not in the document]
themason2011: Yes — this is the relative-position score scaling and it is required. It reproduces NeMo's `RelPositionMultiHeadAttention`, which computes `scores = (matrix_ac + matrix_bd) / sqrt(d_k)`, where `matrix_ac = (q + u) @ kᵀ` and `matrix_bd` is the rel-shifted `(q + v) @ pᵀ`. The single `scale = d_k**-0.5` is applied to the summed terms, matching the reference exactly (offline parity max abs diff 1.2e-7 vs the NeMo golden). Note this rel-pos path is intentionally left as manual matmuls (not `op.Attention`) because the positional bias `matrix_bd` depends on `q` and the two learned biases `u`/`v`, so it can't be expressed as a static additive `attn_bias`.

src/mobius/models/__init__.py
themason2011: Fixed — removed `build_sortformer` from the public `mobius.models` namespace (dropped from `__all__` and the import in `models/__init__.py`). It now lives in the NeMo integration package (`from mobius.integrations.nemo import build_sortformer`), see the related thread on the loader relocation. The declarative model classes `SortformerConfig` / `SortformerDiarizationModel` remain exported under `mobius.models`.

src/mobius/models/sortformer.py
themason2011: Fixed in c235538 — replaced the manual `x * sigmoid(x)` helper with the opset-24 `op.Swish` op directly. Removed the `_swish` function; the Conformer feed-forward and convolution modules now call `op.Swish(...)`. Offline parity vs the NeMo golden is unchanged (max abs diff 1.2e-7).

src/mobius/models/sortformer.py
themason2011: Fixed in c235538 — `_TransformerAttention` now uses the opset-24 `op.Attention` op (passing the `[B, T, D]` Q/K/V projections with `q_num_heads`/`kv_num_heads`, which splits heads internally), replacing the manual reshape/transpose/matmul/softmax. No mask is applied (bidirectional, unpadded single-utterance input). Parity vs the NeMo golden is unchanged (max abs diff 1.2e-7). Note: the rel-pos Conformer attention is intentionally left manual — see the thread on line 402 for why its positional bias can't be a static `attn_bias`.

src/mobius/models/sortformer.py
themason2011: Fixed in c235538 — moved `build_sortformer` out of the modeling code into a dedicated NeMo integration package, `mobius.integrations.nemo` (`__init__.py` + `_sortformer.py`), mirroring the GGUF integration layout (`mobius.integrations.gguf.build_from_gguf`). The loader now holds only the `.nemo` archive/IO glue (untar, YAML, `torch.load`) and delegates graph construction to `SortformerDiarizationModel` + `build_from_module`, keeping `models/sortformer.py` purely declarative. Updated the integration test import accordingly.

src/mobius/integrations/nemo/_sortformer.py
justinchuby, line 1: It seems there isn't a similar file anywhere? I wonder if parts of this would fit in https://github.com/onnxruntime/mobius/blob/main/src/mobius/integrations/nemo/_config_mapping.py

justinchuby (approved):
Thanks! Could you address the copilot comments and ensure the unit test CI + lint CI are green?

justinchuby:
If possible, could you also add the L4/L5 tests?

src/mobius/models/sortformer.py
```diff
@@ -0,0 +1,736 @@
...
+        self.norm_feed_forward2 = LayerNorm(d_model, eps=1e-5)
+        self.feed_forward2 = _ConformerFeedForward(d_model, d_ff)
+        self.norm_out = LayerNorm(d_model, eps=1e-5)
+
+    def forward(
+        self, op: OpBuilder, x: ir.Value, pos_emb: ir.Value, t_dim: ir.Value
+    ) -> ir.Value:
+        # Macaron feed-forward (half-step residual)
+        ff = self.feed_forward1(op, self.norm_feed_forward1(op, x))
+        x = op.Add(x, op.Mul(ff, op.Constant(value_float=0.5)))
```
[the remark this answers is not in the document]
themason2011: Try to build sortformer using dtype=float16/bfloat16. If it doesn't work, we need to apply all data-type related suggestions from copilot. If it does work without changes, we can ignore and un-apply the previous comment

src/mobius/integrations/nemo/_sortformer.py
themason2011: Fixed this by moving it to config_mapping

themason2011:
> If possible, could you also add the L4/L5 tests?

lintrunner f --all-files

src/mobius/models/sortformer.py
```diff
@@ -0,0 +1,736 @@
...
+    def _to_heads(self, op: OpBuilder, x: ir.Value) -> ir.Value:
+        x = op.Reshape(x, [0, 0, self._num_heads, self._head_dim])
+        return op.Transpose(x, perm=[0, 2, 1, 3])
+
+    def forward(self, op: OpBuilder, x: ir.Value) -> ir.Value:
+        q = self._to_heads(op, self.query_net(op, x))
+        k = self._to_heads(op, self.key_net(op, x))
+        v = self._to_heads(op, self.value_net(op, x))
+        scale = op.Constant(value=ir.tensor(np.array(self._head_dim**-0.5, dtype=np.float32)))
+        scores = op.Mul(op.MatMul(q, op.Transpose(k, perm=[0, 1, 3, 2])), scale)
```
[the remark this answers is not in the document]
themason2011: @copilot what would you suggest after the changes?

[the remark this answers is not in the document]
themason2011: @copilot I meant what would you suggest for the current comment now that the previous code it was written for is outdated: "The attention scale constant is hard-coded as float32, which will make float16/bfloat16 exports invalid (Mul inputs must share dtype). Cast the constant to the score tensor dtype (via CastLike) before multiplying."

themason2011:
> If possible, could you also add the L4/L5 tests?

Added L4/L5 tests

src/mobius/models/sortformer.py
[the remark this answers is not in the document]
themason2011: Got it, then I'll resolve this.

themason2011:
Ran lintrunner with copilot

justinchuby (approved):

PR #446, opened 2026-08-01
Add NVFP4/FP8 (ModelOpt) weight reconstruction foundation for Qwen3.6

justinchuby:
## What

Foundation for consuming **NVIDIA ModelOpt** NVFP4/FP8 mixed-precision checkpoints (e.g. quantized **Qwen3.6**), porting the verifiable numeric core of ORT GenAI [#2350](https://github.com/microsoft/onnxruntime-genai/pull/2350) / [#2351](https://github.com/microsoft/onnxruntime-genai/pull/2351).

ModelOpt Qwen3.6 layout:
- **Routed MoE experts / shared expert / lm_head** → `W4A16_NVFP4`: block-16 E2M1 (fp4) weights, FP8-E4M3 block scales, per-tensor FP32 global scale (`weight_scale_2`).
- **Attention / linear-attn projections** → `FP8` (E4M3, per-tensor `weight_scale`).

## How

New `mobius/integrations/modelopt/` package with the loader's numeric core:
- `dequantize_nvfp4()` — `e2m1(code) × e4m3(block_scale[k//16]) × global_scale → bf16`
- `dequantize_fp8()` — `e4m3(weight) × weight_scale → bf16`
- `unpack_nvfp4_codes()`, `is_modelopt_quant_config()`, `FP4_E2M1_LUT`

Following ORT GenAI's loader, dense FP8 and NVFP4 shared-expert/lm_head weights are **reconstructed to BF16 exactly** (ORT has no FP8 attention GEMM), so the standard BF16 build path can consume the checkpoint. Fully unit-tested against hand-computed references — numeric-only, no runtime/GPU dependency.

`QuantizationConfig.from_transformers` now **fails loudly** (`NotImplementedError`) on ModelOpt schemes instead of silently routing packed E2M1/float8 weights through the INT4 `MatMulNBits` path (which would mis-dequantize them). The check runs before the `quant_method=="none"` early-return since ModelOpt may name its scheme only via `quant_algo`/`quant_cfg`.

## Scope / verification

- ✅ Verifiable here: dequant math (7 unit tests, hand-computed references) + config detection/guard (89 tests pass in `_configs_test.py` + module).
- ⏳ **Out of scope (documented follow-up):** full checkpoint weight-load wiring and **native routed-expert NVFP4 QMoE emission** (CUDA/Blackwell, `onnxruntime_USE_FP4_QMOE=ON`). That path is **not buildable or verifiable** in this environment (no Blackwell GPU, no FP4-QMOE ORT build, no real ModelOpt checkpoint), so it is intentionally gated rather than shipped unverified — the safe increment given the repo's "no silently-wrong weights" bar.

Please review — do not merge yet.


tianleiwu (commented):
### Summary

The numeric core is clean and the "fail loudly instead of mis-dequantizing" guard is the right call for this repo's no-silently-wrong-weights bar. Three items look worth resolving before this lands; details are inline.

**High priority**

1. **`dequantize_nvfp4` never validates the block-scale row dimension.** The block *count* is checked, but `shape[0]` is not, so a mismatched-N scale tensor silently broadcasts row 0 across all output rows. Verified against this head:
   ```python
   w      = np.full((4, 8), 0x22, dtype=np.uint8)                # N=4, K=16
   bs_bad = np.full((1, 1), 2.0, dtype=ml_dtypes.float8_e4m3fn)  # N=1  <-- wrong
   dequantize_nvfp4(w, bs_bad, 1.0).shape                        # -> (4, 16), no error
   ```
   This is the same silently-wrong-reconstruction class the block-size check was added to prevent.

2. **`is_modelopt_quant_config` misses `quant_algo == "MIXED_PRECISION"`**, which is what ModelOpt emits for exactly the NVFP4-MoE + FP8-attention Qwen3.6 layout this PR targets.

3. **`ml_dtypes` is not a declared dependency** but is now imported at module scope on a path reached by every `QuantizationConfig.from_transformers` call.

**Follow-up worth filing (outside this diff)**

`src/mobius/_weight_loading.py::_dequantize_fp8_weights` only looks for `weight_scale_inv`. ModelOpt names the per-tensor scale `weight_scale`, so a ModelOpt FP8 tensor takes the `else` branch — `logger.warning("FP8 weight '%s' has no scale_inv - casting without scaling")` — and yields unscaled weights with only a WARNING. That is the exact failure mode this PR guards against, and it stays reachable for any ModelOpt checkpoint whose `quantization_config` is not flagged by `is_modelopt_quant_config` (see item 2). Teaching `_dequantize_fp8_weights` about `weight_scale` is small, torch-native, and verifiable in-repo — arguably higher value than the currently-unreferenced numpy `dequantize_fp8`.

Also note the new numpy `dequantize_fp8` sits alongside that existing torch implementation in a torch-based loader pipeline; a line in the module docstring saying which one is intended to survive would help the next reader.

**What's good**

- E2M1 decode via `FP4_E2M1_LUT[codes & 0x7]` + sign mask is branch-free and format-faithful.
- Tests use hand-computed references rather than round-trip self-consistency, and the raw-`uint8` vs typed-`float8` equivalence test is exactly the right case for real checkpoints.
- Raising `NotImplementedError` with an actionable message beats silently routing packed E2M1 through `MatMulNBits`.

**Nit (line 45, outside the diff hunk):** `from_transformers`'s docstring still documents only the `None` return but the method now raises. Worth a `Raises:` section — ruff `D` with the Google convention is enabled in this repo.

src/mobius/integrations/modelopt/_dequant.py
```diff
@@ -0,0 +1,148 @@
...
+        ``ml_dtypes.bfloat16`` array ``[N, K]`` where
+        ``w = e2m1(code) * e4m3(block_scale[n, k // 16]) * global_scale``.
+    """
+    codes = unpack_nvfp4_codes(weight_u8).astype(np.int64)  # [N, K]
+    mag = FP4_E2M1_LUT[codes & 0x7]  # [N, K]
+    val = np.where((codes & 0x8) > 0, -mag, mag).astype(np.float32)  # [N, K]
+
+    block_scale = _to_float32(block_scale_e4m3)  # [N, K/16]
+    k = codes.shape[1]
+    n_blocks = block_scale.shape[1]
```
tianleiwu, line 97: **High:** the block *count* is validated but the block-scale **row** dimension never is, so a mismatched-N scale silently broadcasts instead of failing. Verified on this head:

```python
w      = np.full((4, 8), 0x22, dtype=np.uint8)                # N=4, K=16
bs_bad = np.full((1, 1), 2.0, dtype=ml_dtypes.float8_e4m3fn)  # N=1  <-- wrong
dequantize_nvfp4(w, bs_bad, 1.0).shape                        # -> (4, 16), no error
```

Row 0's scale gets applied to all `N` rows — the same silently-wrong-reconstruction class the `NVFP4_BLOCK_SIZE` check below was added to prevent. A 1-D `block_scale` is also not handled cleanly: `block_scale.shape[1]` raises `IndexError: tuple index out of range` rather than the intended `ValueError`.

Suggested, right after `block_scale = _to_float32(block_scale_e4m3)`:

```python
if block_scale.ndim != 2:
    raise ValueError(
        f"NVFP4 block scales must be 2D [N, K/16], got shape {block_scale.shape}."
    )
if block_scale.shape[0] != codes.shape[0]:
    raise ValueError(
        f"NVFP4 block scale rows {block_scale.shape[0]} != weight rows {codes.shape[0]}."
    )
```

src/mobius/integrations/modelopt/_dequant.py
```diff
@@ -0,0 +1,148 @@
...
+    """
+    if not isinstance(quantization_config, dict):
+        return False
+    method = str(quantization_config.get("quant_method", "")).lower()
+    if method == "modelopt":
+        return True
+    algo = str(
+        quantization_config.get("quant_algo") or quantization_config.get("quant_cfg") or ""
+    ).upper()
+    return "NVFP4" in algo or "FP8" in algo
```
tianleiwu, line 141: **High:** this misses `quant_algo == "MIXED_PRECISION"`, which is what ModelOpt writes whenever a checkpoint uses more than one format (`modelopt/torch/export/quant_utils.py`: *"If we have more than one quantization format, infer MIXED_PRECISION"*, handled in `convert_hf_config.py` together with a `quantized_layers` map). That is precisely the NVFP4-MoE + FP8-attention Qwen3.6 checkpoint this PR targets, and `"MIXED_PRECISION"` contains neither `NVFP4` nor `FP8`.

Detection then rests entirely on `quant_method == "modelopt"`, which contradicts the rationale given in `_quantization.py` for running the check before the `quant_method == "none"` early-return ("ModelOpt may name its scheme only via `quant_algo`/`quant_cfg`"). A config carrying only `quant_algo: "MIXED_PRECISION"` falls through to `return None` — the silent mis-route this guard exists to prevent.

```python
if str((quantization_config.get("producer") or {}).get("name", "")).lower() == "modelopt":
    return True
...
return "NVFP4" in algo or "FP8" in algo or algo == "MIXED_PRECISION"
```

`producer.name` is written by ModelOpt's exporter and is the most robust signal; `quantized_layers` is another mixed-precision-only key.

src/mobius/integrations/modelopt/_dequant.py
```diff
@@ -0,0 +1,148 @@
...
+The functions below are the numeric core of the loader and are format-faithful
+to ModelOpt's on-disk encoding (verified in ``_dequant_test.py``).
+
+E2M1 (fp4) magnitude table, indexed by the low 3 bits of each 4-bit code
+(bit 3 is the sign): ``{0, 0.5, 1, 1.5, 2, 3, 4, 6}``.
+"""
+
+from __future__ import annotations
+
+import ml_dtypes
```
tianleiwu, line 32: **High:** `ml_dtypes` is not declared in `pyproject.toml` — neither in `[project] dependencies` nor in any `optional-dependencies` group. It is only present transitively via `onnx_ir`. Because `_configs/_quantization.py` now imports this package, this module-level import runs on **every** `QuantizationConfig.from_transformers` call, i.e. on the core export path for all models, not just ModelOpt ones.

This is also the only top-level `import ml_dtypes` under `src/`: `src/mobius/_testing/ort_inference.py` imports it lazily inside functions, and `rewrite_rules/_static_empty_kv.py` goes through `ir_dtype.numpy()` instead.

Either add `ml_dtypes` to `[project] dependencies`, or keep this numeric module off the config import path (see the comment on `_quantization.py`).

src/mobius/integrations/modelopt/_dequant.py
```diff
@@ -0,0 +1,148 @@
...
+import numpy as np
+
+# E2M1 (fp4) decoded magnitudes, indexed by ``code & 0x7`` (bit 3 is the sign).
+FP4_E2M1_LUT = np.array([0.0, 0.5, 1.0, 1.5, 2.0, 3.0, 4.0, 6.0], dtype=np.float32)
```
tianleiwu, line 36: This is exported in `__all__` and `FP4_E2M1_LUT.flags.writeable` is `True`, so any caller (or a test) that writes into it silently corrupts every subsequent dequantization process-wide. Suggest freezing it:

```python
FP4_E2M1_LUT = np.array([0.0, 0.5, 1.0, 1.5, 2.0, 3.0, 4.0, 6.0], dtype=np.float32)
FP4_E2M1_LUT.setflags(write=False)
```

src/mobius/_configs/_quantization.py
```diff
@@ -53,6 +53,27 @@ def from_transformers(cls, hf_config) -> QuantizationConfig | None:
...
+        # NVIDIA ModelOpt NVFP4/FP8 checkpoints (e.g. quantized Qwen3.6) encode
+        # weights as packed E2M1 (fp4) / float8 with block + global scales — a
+        # layout the INT4 ``QuantizedLinear``/``MatMulNBits`` path below would
+        # silently mis-dequantize. Checked before the ``quant_method == "none"``
+        # early-return because ModelOpt may name its scheme only via
+        # ``quant_algo``/``quant_cfg`` (no ``quant_method``). Fail loudly: the
+        # reconstruction math is available in :mod:`mobius.integrations.modelopt`,
+        # but the full weight-load integration + native routed-expert NVFP4 QMoE
+        # emission (CUDA/Blackwell, ``onnxruntime_USE_FP4_QMOE=ON``) are not wired.
+        from mobius.integrations.modelopt import is_modelopt_quant_config
```
tianleiwu, line 65: `is_modelopt_quant_config` is a pure dict predicate with no numeric content, yet importing it pulls in `numpy` + `ml_dtypes` and inverts the layering (core config -> integration package) — which is why the import has to be function-local here.

Moving the predicate into `_configs`, or into a dependency-free `integrations/modelopt/_detect.py` that `__init__` re-exports, resolves both the layering inversion and the undeclared-`ml_dtypes`-on-a-core-path issue, and lets the import move to module scope.

Separate question on the same block: the substring match makes any `{"quant_algo": "FP8"}` config a hard error, but `_weight_loading._dequantize_fp8_weights` already reconstructs FP8 -> BF16 and the PR description says FP8 reconstructs exactly. Is blocking plain per-tensor FP8 intended, or is only the NVFP4 half actually unsupported?

src/mobius/integrations/modelopt/_dequant_test.py
```diff
@@ -0,0 +1,122 @@
...
+    raw = typed.view(np.uint8)
+    a = dequantize_nvfp4(weight_u8, typed, 1.0).astype(np.float32)
+    b = dequantize_nvfp4(weight_u8, raw, 1.0).astype(np.float32)
+    np.testing.assert_array_equal(a, b)
+    np.testing.assert_array_equal(a, np.full((1, 16), 3.0, dtype=np.float32))
+
+
+def test_unpack_nvfp4_codes_rejects_non_2d():
+    # Loader code should surface a shape mistake, not silently reinterpret it.
+    with np.testing.assert_raises(ValueError):
```
tianleiwu, line 93: Nit: these are the only two uses of `np.testing.assert_raises` in the repo — `pytest.raises` is used 151 times under `src/`. Prefer `pytest.raises(ValueError, match="must be 2D")`; matching the message also pins *which* validation fired, which matters now that `dequantize_nvfp4` has more than one `ValueError` path.

src/mobius/_configs_test.py
```diff
@@ -837,6 +837,24 @@ def test_from_transformers_fp8_returns_none(self):
         )()
         assert QuantizationConfig.from_transformers(hf) is None
 
+    def test_from_transformers_modelopt_nvfp4_raises(self):
+        """ModelOpt NVFP4/FP8 checkpoints fail loudly rather than mis-quantizing.
+
+        The INT4 path would silently mis-dequantize packed E2M1/float8 weights,
+        so ``from_transformers`` raises until the ModelOpt loader is wired.
+        """
+        import pytest
```
tianleiwu, line 846: Nit: `pytest` is already imported at module scope (line 11), so this local import can be dropped.

src/mobius/integrations/modelopt/_dequant.py
```diff
@@ -0,0 +1,131 @@
...
+    high nibble = odd ``K``) — the layout ModelOpt writes. Returns uint8 codes
+    ``[N, K]`` in ``0..15``.
+    """
+    packed = np.ascontiguousarray(packed_nk2).astype(np.uint8)
+    low = packed & 0x0F
+    high = packed >> 4
+    n = packed.shape[0]
+    # Interleave (low, high) along a new trailing axis, then flatten to [N, K].
+    codes = np.stack((low, high), axis=-1).reshape(n, -1)
+    return np.ascontiguousarray(codes)
```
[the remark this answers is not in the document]
tianleiwu: Half of this is addressed by d08c344 (the `ndim != 2` guard); the **dtype** half is still open — `.astype(np.uint8)` accepts and truncates any dtype:

```python
dequantize_nvfp4(w.astype(np.float32), bs_ok, 1.0)   # -> (4, 16), no error
```

Two follow-ons on the new guard:

- The docstring says the rank check catches "passing already-unpacked codes", but already-unpacked `[N, K]` codes are 2-D too, so they pass it — they are only caught downstream by the `NVFP4_BLOCK_SIZE` check. Worth rewording.
- `packed_nk2.ndim` is read before `np.asarray`, so a list/sequence input raises `AttributeError` instead of the intended `ValueError`.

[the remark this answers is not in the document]
justinchuby: Fixed both follow-ons in 1e7799c:
- **dtype half:** `unpack_nvfp4_codes` now rejects non-`uint8` input (`raise ValueError("NVFP4 packed codes must be uint8 ...")`) instead of `.astype(np.uint8)`-truncating it, so `dequantize_nvfp4(w.astype(np.float32), ...)` raises. Added `test_unpack_nvfp4_codes_rejects_non_uint8`.
- **AttributeError:** `np.asarray(packed_nk2)` now runs *before* the `ndim` check, so a list/sequence input raises the intended `ValueError`, not `AttributeError`.
- **docstring:** reworded — the rank guard no longer claims to catch already-unpacked `[N, K]` codes (they are 2-D and only fail downstream against `NVFP4_BLOCK_SIZE`).

src/mobius/integrations/modelopt/_dequant_test.py
```diff
@@ -0,0 +1,101 @@
...
+    weight_u8 = np.array([[byte0, byte1]], dtype=np.uint8)
+
+    block_scale = np.array([[2.0]], dtype=ml_dtypes.float8_e4m3fn)  # one block
+    global_scale = 0.5
+
+    out = dequantize_nvfp4(weight_u8, block_scale, global_scale)
+    assert out.dtype == ml_dtypes.bfloat16
+    # val * 2.0 (block) * 0.5 (global) == val
+    expected = np.array([[1.0, -0.5, 4.0, -2.0]], dtype=np.float32)
+    np.testing.assert_array_equal(out.astype(np.float32), expected)
```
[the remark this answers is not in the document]
justinchuby: Addressed. The fixtures already use one full `K=16` NVFP4 block (`test_dequantize_nvfp4_known_values` asserts the trailing 12 elements are zero). `K=4` now survives only in `test_dequantize_nvfp4_rejects_non16_block`, which deliberately asserts the non-16 block size is rejected.

src/mobius/integrations/modelopt/_dequant_test.py
```diff
@@ -0,0 +1,101 @@
...
+    # Block scales may arrive as raw uint8 e4m3 code bytes; result must match a
+    # typed float8 view.
+    k0 = _e2m1_code(0, 2)
+    weight_u8 = np.array([[k0 | (k0 << 4)]], dtype=np.uint8)  # K=2, both +1.0
+    typed = np.array([[3.0]], dtype=ml_dtypes.float8_e4m3fn)
+    raw = typed.view(np.uint8)
+    a = dequantize_nvfp4(weight_u8, typed, 1.0).astype(np.float32)
+    b = dequantize_nvfp4(weight_u8, raw, 1.0).astype(np.float32)
+    np.testing.assert_array_equal(a, b)
+    np.testing.assert_array_equal(a, np.array([[3.0, 3.0]], dtype=np.float32))
```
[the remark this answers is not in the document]
justinchuby: Addressed. `test_dequantize_nvfp4_raw_uint8_block_scale` now uses a full `K=16` block while still exercising the raw-`uint8` e4m3 code-byte path for the block scales.

src/mobius/integrations/modelopt/_dequant_test.py
```diff
@@ -0,0 +1,101 @@
...
+
+    out = dequantize_nvfp4(weight_u8, block_scale, global_scale)
+    assert out.dtype == ml_dtypes.bfloat16
+    # val * 2.0 (block) * 0.5 (global) == val
+    expected = np.array([[1.0, -0.5, 4.0, -2.0]], dtype=np.float32)
+    np.testing.assert_array_equal(out.astype(np.float32), expected)
+
+
+def test_dequantize_nvfp4_block_scale_repeat():
+    # Two 16-element blocks with distinct block scales exercise repeat_interleave.
```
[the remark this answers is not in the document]
justinchuby: Addressed. The comment now reads "exercise the `np.repeat` per-block scale broadcast", matching the implementation (`np.repeat`).

src/mobius/integrations/modelopt/_dequant.py
justinchuby: Fixed in 1e7799c. Right after `block_scale = _to_float32(...)` `dequantize_nvfp4` now validates rank and rows:

```python
if block_scale.ndim != 2:
    raise ValueError(f"NVFP4 block scales must be 2D [N, K/16], got shape {block_scale.shape}.")
if block_scale.shape[0] != codes.shape[0]:
    raise ValueError(f"NVFP4 block scale rows {block_scale.shape[0]} != weight rows {codes.shape[0]}.")
```

So the mismatched-`N` example now raises instead of broadcasting, and a 1-D `block_scale` raises a clear `ValueError` rather than `IndexError`. Added `test_dequantize_nvfp4_rejects_block_scale_row_mismatch` and `test_dequantize_nvfp4_rejects_1d_block_scale`.

src/mobius/integrations/modelopt/_dequant.py
justinchuby: Fixed in 1e7799c. Detection now recognises the mixed-precision signals:

```python
if str(quantization_config.get("quant_method", "")).lower() == "modelopt":
    return True
producer = quantization_config.get("producer")
if isinstance(producer, dict) and str(producer.get("name", "")).lower() == "modelopt":
    return True
if "quantized_layers" in quantization_config:
    return True
...
return "NVFP4" in algo or "FP8" in algo or algo == "MIXED_PRECISION"
```

So a config carrying only `quant_algo: "MIXED_PRECISION"` (or just the `producer.name`/`quantized_layers` stamps) is now detected as ModelOpt instead of falling through to `None`. Covered by `test_is_modelopt_quant_config` and `test_from_transformers_modelopt_nvfp4_raises`.

src/mobius/integrations/modelopt/_dequant.py
justinchuby: Fixed in 1e7799c: added `ml_dtypes` to `[project] dependencies` in `pyproject.toml`. I also removed it from the core config path entirely — the pure `is_modelopt_quant_config` predicate moved to a dependency-free `integrations/modelopt/_detect.py` and the numeric `_dequant` symbols are now lazily imported by the package `__init__`, so `from_transformers` no longer imports `numpy`/`ml_dtypes` (see the reply on `_quantization.py`).

src/mobius/integrations/modelopt/_dequant.py
justinchuby: Fixed in 1e7799c — froze the shared LUT:

```python
FP4_E2M1_LUT = np.array([...], dtype=np.float32)
FP4_E2M1_LUT.flags.writeable = False
```

Verified `FP4_E2M1_LUT.flags.writeable is False`, so a caller can no longer mutate it process-wide.

src/mobius/_configs/_quantization.py
justinchuby: Both addressed in 1e7799c.

**Layering / undeclared numpy:** `is_modelopt_quant_config` moved to a new dependency-free `integrations/modelopt/_detect.py` (pure dict predicate, no `numpy`/`ml_dtypes`). The package `__init__` re-exports it eagerly and loads the numeric `_dequant` symbols lazily via `__getattr__`, and `_configs/_quantization.py` now imports the predicate from `_detect` at module scope. Verified importing the package + calling the predicate does **not** import `_dequant`, so the core path stays numpy-free.

**Plain FP8 question:** blocking is intentional only for ModelOpt-*named* FP8. Plain HF per-tensor FP8 uses `quant_method="fp8"` with no `quant_algo`, so `is_modelopt_quant_config` returns `False` (→ `from_transformers` returns `None`) and it stays on the existing `_weight_loading._dequantize_fp8_weights` path. Only a config that names a ModelOpt scheme via `quant_algo`/`quant_cfg` (`FP8`/`NVFP4`/`MIXED_PRECISION`) or the `producer`/`quantized_layers` stamps is claimed — that FP8 is the ModelOpt attention-projection layout whose full weight-load integration is not yet wired. Added a `{"quant_method": "fp8"}` negative case to `test_is_modelopt_quant_config`.

src/mobius/integrations/modelopt/_dequant_test.py
justinchuby: Done in 1e7799c — switched to `pytest.raises(ValueError, match=...)` on both, pinning the specific validation (e.g. `match="must be 2D"`, `match="block scale rows"`, `match="block size must be"`).

src/mobius/_configs_test.py
justinchuby: Done in 1e7799c — dropped the redundant local `import pytest`; the test uses the module-scope import at line 11.

PR #449, opened 2026-08-03
fix(functions): add optional bias slot to PackedMultiHeadAttention fallback

justinchuby:
## Summary

Fixes the formal-input arity of the standard-ONNX fallback `ir.Function` for `com.microsoft::PackedMultiHeadAttention` by adding the optional `bias` slot.

## Root cause

ORT's `PackedMultiHeadAttention` has a 6–7 input **positional** signature:

```
query, key(opt), value(opt), bias(opt), token_offset, cumulative_sequence_length, attention_bias(opt)
```

Because `token_offset` / `cumulative_sequence_length` occupy positional slots 5 and 6, the optional `bias` at slot 4 must exist as a formal input (present-as-absent at call sites via `""`).

Call sites in mobius already correctly emit 6 inputs `(q, k, v, "", token_offset, cu_seqlens)`:
- `src/mobius/models/_qwen3_vl_vision.py`
- `src/mobius/rewrite_rules/_packed_attention.py`

But the fallback `ir.Function` declared **only 5 formals**, dropping the `bias` slot. This made onnx-genai's function-inline admission reject the call:

```
call.input.len()=6 > func.input.len()=5  →  FunctionArityMismatch
```

for the Qwen3.6-35B-A3B `vision_encoder`. onnx-genai is spec-correct here; the bug was purely the missing `bias` formal in mobius.

## Fix

- Insert `bias` as the 4th formal input, between `value` and `token_offset`, yielding the correct positional order: `query, key, value, bias, token_offset, cumulative_sequence_length` (6 formals).
- Add the matching `bias_input` parameter to the body `body(...)` in the same position. The body does **not** use `bias` — the fallback reconstructs the block-diagonal attention bias from `cumulative_sequence_length` — so it is present only to preserve positional slot alignment (treated like the already-unused `token_offset_input`).
- Update the module/function docstrings to list `bias: (optional) — unused in fallback` in the correct positional order.

## Test

Added `tests/packed_multi_head_attention_function_test.py`, which asserts:
- the function declares exactly 6 formal inputs in the exact positional order `query, key, value, bias, token_offset, cumulative_sequence_length` (not just the count — `bias` must be at index 3);
- a 6-input call node is admissible against it (`len(call inputs) <= len(func inputs)`);
- the function domain/name identity (`com.microsoft` / `PackedMultiHeadAttention`).

All 3 new tests pass; `tests/ep_optimization_test.py` (30 tests) passes as a regression check; ruff check + format clean.


justinchuby:
VERDICT: APPROVE

Independent review by Harry (reviewer, opus-4.8). I did not author this change. Reviewed at commit 864c04f on branch squad/packed-mha-bias-slot in a detached worktree off origin/main.

**1. Positional-wiring integrity — CORRECT.**
`build_function` calls `trace_function(gb.op, *trace_args)` and builds `graph_inputs` from the same `inputs` list, so body parameters map to formals strictly by order (verified by reading onnxscript._internal.builder.build_function). The body now takes `(op, query_input, key_input, value_input, bias_input, token_offset_input, cumulative_sequence_length_input)` and the formals list is `[query, key, value, bias, token_offset, cumulative_sequence_length]` — perfectly aligned. `token_offset` and `cumulative_sequence_length` still land on the right formals. The inner `op.Attention(...)` receives `query_batched, key_batched, value_batched, attention_bias`, where `attention_bias` is the block-diagonal mask RECONSTRUCTED from `cumulative_sequence_length_input` — NOT the new `bias_input`. No miscompute.

**2. bias is genuinely inert.** `grep` shows `bias_input` appears only as the parameter name (line 75) and in an explanatory comment (line 79). It is never fed to any op — exactly as inert as `token_offset_input`.

**3. Call-site parity — CORRECT (all 3 sites).** Note: the actual call sites are `src/mobius/components/_qwen3_vl_vision.py` and `_qwen25_vl_vision.py` (shared `_emit_packed_mha`), plus the rewrite rule — the paths in the review brief were slightly off, but all were located and checked:
- `rewrite_rules/_packed_attention.py:170` → `PackedMultiHeadAttention(q, k, v, None, token_offset, cu_seqlens_i32)`
- `components/_qwen3_vl_vision.py:252` → `(query_mha, key_mha, value_mha, None, token_offset, cu_seqlens_int32)`
- `components/_qwen25_vl_vision.py` → `(query_mha, key_mha, value_mha, None, token_offset, cu_seqlens_int32)`
All three emit 6 inputs with `None` (present-as-absent) in slot 4 (bias), matching the new 6-formal signature. No call site puts bias in the wrong slot.

**4. Test non-vacuous + mutation check.** `tests/packed_multi_head_attention_function_test.py` asserts the FULL positional order (`actual_order == EXPECTED_INPUT_ORDER`) and explicitly `func.inputs[3].name == "bias"`, not just the count. Mutation check performed: I temporarily moved the `bias` formal to last; the test FAILED with `At index 3 diff: 'token_offset' != 'bias'`. Reverted; 3 passed in ~2s. Guard is effective.

**5. Scope.** `git show --stat` confirms only 2 files changed (the function + its new test), 74 insertions — mobius-only, no onnx-genai changes, no scope creep. Docstrings (module + `bias` slot lines) updated correctly.

Test result: `3 passed in 1.71s`.

The fix correctly restores the 6th formal (bias at slot 4) so onnx-genai's function-inline admission accepts the 6-input call nodes, with zero behavioral change to the fallback decomposition. Approving.

justinchuby:
@copilot+claude-opus-5 please fix review comments

PR #457, opened 2026-08-04
world model framework

xiaoyu-work:
add world model framework init support.


rui-ren (approved):

rui-ren:
@xiaoyu-work which world model will we support?

justinchuby (approved):

xiaoyu-work:
> @xiaoyu-work which world model will we support?

this is a general framework for world model. I will start testing the smallest model maybe.

Issue #427, opened 2026-07-24
Add defensive ndim==2 guard in preprocess_olive_weights to reject Olive moe=True 3D expert checkpoints


titaiwangms:
## Summary

`preprocess_olive_weights` in `src/mobius/_weight_utils.py` assumes every Olive-quantized weight tensor is 2-D (Linear / Embedding). Once Olive's upcoming `moe=True` RTN quantization lands, it will emit **3-D fused-expert** weight tensors `(num_experts, out_features, in_features)`. The current code has no `ndim`/shape guard, so it will silently mis-reshape those 3-D tensors and produce garbage weights with no error.

This issue asks only for a **defensive guard** that rejects (or explicitly refuses) 3-D expert tensors. Implementing actual MoE quantization support in Mobius is **out of scope** and tracked separately.

## Trigger

microsoft/Olive PR [microsoft/Olive#2584](https://github.com/microsoft/Olive/pull/2584) (resolving [microsoft/Olive#2583](https://github.com/microsoft/Olive/issues/2583)) adds a `moe=True` flag to Olive's PyTorch RTN quantization pass. With that flag, MoE expert weights are stored as **3-D fused tensors** `(num_experts, out_features, in_features)` in Olive's `QuantTensor` format, using the same buffer naming Mobius already keys on: `<pname>_qweight` / `<pname>_scales` / `<pname>_qzeros`.

Mobius already consumes Olive's **non-MoE** (2-D) checkpoints through `preprocess_olive_weights`, called from `src/mobius/models/base.py` (~line 451). Nothing in that path distinguishes 2-D from 3-D tensors.

## Where the guard is missing

`src/mobius/_weight_utils.py`, in the generic `.qweight` branch:

```python
727:        elif key.endswith(".qweight"):
728:            if value.dtype != torch.uint8:
729:                raise ValueError(f"Olive qweight must be uint8 for {key}, got {value.dtype}")
730:            if value.shape[-1] % blob_size != 0:
731:                raise ValueError(...)
732:            new_key = key.replace(".qweight", ".weight")
736:            result[new_key] = value.reshape(value.shape[0], -1, blob_size).contiguous()
```

**Line 736** is the silent-corruption site. For a 3-D expert tensor `(num_experts, out_features, packed_in)`:

- `value.shape[0]` is `num_experts` (not `out_features`), and
- `value.reshape(value.shape[0], -1, blob_size)` collapses `out_features * packed_in` into the middle axis,

[...]

PR #458, opened 2026-08-06
Fix GGUF config mapping for scaled RoPE and hybrid attention

justinchuby:
## Summary

Fix GGUF -> ArchitectureConfig gaps that surface as deep model-construction errors instead of diagnosable GGUF metadata issues. These are general GGUF mapping bugs, not one-checkpoint quirks: any scaled-RoPE GGUF can set `rope_type` to `linear`/`yarn`/`longrope` while leaving `config.rope_scaling` unset, and any hybrid/sliding-window architecture needs the layer schedule and local RoPE fields populated before graph construction.

Changes:
- Populate `rope_scaling` (including `factor`) when GGUF declares `rope.scaling.type` as `linear`, `yarn`, or `longrope`; warn and use identity factor `1.0` if the factor key is absent instead of failing later.
- Populate local/sliding-window RoPE base frequency from `rope.freq_base_swa`, or default to `10000.0` for sliding-window GGUFs (matching HF Gemma3 defaults).
- Preserve GGUF `attention.sliding_window` and derive Gemma3 `layer_types` from HF's architectural default `sliding_window_pattern=6` (five sliding-window layers, then one full-attention layer).
- Fill other GGUF metadata/model gaps encountered after that: Gemma3 q/k norm config, Gemma3 default activation, Gemma3 GGUF norm tensor mappings, and quantized Gemma3 graph construction so `--keep-quantized` produces `MatMulNBits` instead of expanding projections to f16.

## Errors this prevents

These failures previously occurred far from GGUF parsing and did not name the missing metadata/config field:

- `TypeError: 'NoneType' object is not subscriptable` from `LinearRope.__init__` when `rope_type` was scaled but `config.rope_scaling` was `None`.
- `TypeError: unsupported operand type(s) for **: 'NoneType' and 'float'` from `_get_default_inv_freq` when sliding-window local RoPE theta was missing.
- `TypeError: 'NoneType' object is not iterable` from `gemma3_text.py` when `config.layer_types` was missing.
- `ValueError: Component 'model' has 124 initializer(s) without weights...` after graph construction when Gemma3 GGUF norm tensors were skipped/mis-mapped.
- Silent quantization loss: `--keep-quantized` built a ~53 GiB f16 model with `MatMulNBits=0`; after this change the same build is ~16 GiB with `MatMulNBits=434`.

I considered raising explicit config errors for absent `rope.scaling.factor`, but kept the warning + identity fallback because `1.0` is a safe no-scaling factor and lets partially specified GGUF metadata remain buildable. The missing local RoPE and layer schedule are now populated before the model code can fail with generic `TypeError`s.

## Regression tests

Added synthetic metadata/unit tests; no large GGUF is required. I also verified they fail without their corresponding implementation changes:

- Reverting `_config_mapping.py` and running `python -m pytest src\mobius\integrations\gguf\_config_mapping_test.py -q` fails all 5 new config tests (`rope_scaling`, local RoPE, sliding window/layer types).
- Reverting `_tensor_mapping.py` and running `python -m pytest src\mobius\integrations\gguf\_tensor_mapping_test.py -q` fails the Gemma3 norm mapping regression.
- Reverting `gemma3_text.py` and running `python -m pytest src\mobius\models\gemma3_text_test.py -q` fails the quantized-component and retied-head regressions.

## Validation

- `python -m ruff check src\mobius\integrations\gguf\_config_mapping.py src\mobius\integrations\gguf\_config_mapping_test.py src\mobius\integrations\gguf\_tensor_mapping.py src\mobius\integrations\gguf\_tensor_mapping_test.py src\mobius\models\gemma3_text.py src\mobius\models\gemma3_text_test.py`
- `python -m ruff format --check src\mobius\integrations\gguf\_config_mapping.py src\mobius\integrations\gguf\_config_mapping_test.py src\mobius\integrations\gguf\_tensor_mapping.py src\mobius\integrations\gguf\_tensor_mapping_test.py src\mobius\models\gemma3_text.py src\mobius\models\gemma3_text_test.py`
- `python -m pytest src\mobius\integrations\gguf\ src\mobius\_configs\ src\mobius\models\gemma3_text_test.py` -> `182 passed`
- Built the target model:
  `python -m mobius build-gguf C:\Users\justinchu\dev\models-gguf\gemma-3-27b-it-Q4_0.gguf --output C:\Users\justinchu\dev\models\gemma-3-27b-onnx --keep-quantized --dtype f16 --ep cuda`
- Verified output:
  - path: `C:\Users\justinchu\dev\models\gemma-3-27b-onnx\model.onnx`
  - total size: `16.079 GiB`
  - `onnx.checker.check_model(...)`: OK
  - `MatMulNBits`: 434
  - `GatherBlockQuantized`: 0 (`token_embd.weight` is Q6_K in this GGUF, so it does not match the graph's Q4_0 projection quantization config and remains f16)

## Why this matters

This came up while preparing a 27B int4 model (~15-16 GiB) to test virtual-memory management, weight offloading, weight streaming, and prefetch in `onnx-genai` on an 8 GB RTX 4060. Prior measurements used qwen2.5-0.5b, whose weights fit comfortably on the card and therefore cannot exercise PCIe saturation or memory-manager pressure.


justinchuby:
Performance update added in commit `8d79344` (same PR, because the perf defect blocks validating the same GGUF conversion path fixed by the config-mapping changes).

What changed:
- `_load_quantized_state_dict` now repacks independent quantized tensors through a bounded `ThreadPoolExecutor` (default `min(16, os.cpu_count())`, override with `MOBIUS_GGUF_REPACK_WORKERS`). Non-repacked/dequantized tensors stay serial to avoid expanding multiple multi-GB embeddings at once.
- Q4_0/Q4_1 nibble reorder now writes directly into the output array instead of allocating low/high nibble arrays plus a concatenate.
- Progress now reports repacked tensor count, in-flight work, packed GiB during repack, and external-data written GiB during ONNX save.

Profiling/measurement before optimizing:
- Synthetic Q4_0 raw repack profile (`8192x8192`, 37.7 MB raw, 5 iterations): `0.614s`; cProfile showed `_reorder_nibbles_gguf_to_ort` consumed `0.552s` of `0.608s` cumulative time (~91%).
- Thread scaling check over independent synthetic Q4_0 tensors: 1 worker `0.812s`, 16 workers `0.166s`, confirming NumPy releases enough GIL here for threads to help.

After changes:
- Same direct Q4_0 repack microbenchmark: `0.551s` (small vectorization win; `_reorder_nibbles_gguf_to_ort` still dominates per tensor).
- Synthetic `_load_quantized_state_dict` benchmark (112 tensors, 0.510 GiB raw):
  - 1 worker: `1.747s` (`0.313 raw GB/s`)
  - 2 workers: `0.976s`
  - 4 workers: `0.563s`
  - 8 workers: `0.445s`
  - 16 workers: `0.369s` (`1.485 raw GB/s`)

Gemma-3 27B rebuild after optimization:
- Command completed successfully in `81.122s` wall-clock.
- Output: `C:\Users\justinchu\dev\models\gemma-3-27b-onnx\model.onnx`
- Total size: `16.079 GiB`
- `onnx.checker.check_model(...)`: OK
- `MatMulNBits`: 434
- `GatherBlockQuantized`: 0 (`token_embd.weight` is Q6_K in this GGUF, so it remains f16 rather than using the graph's Q4_0 quantization config)

Validation after perf changes:
- `python -m pytest src\mobius\integrations\gguf\ src\mobius\_configs\ src\mobius\models\gemma3_text_test.py src\mobius\_model_package_test.py` -> `211 passed`
- Ruff check/format-check on touched files -> passed.

justinchuby:
Follow-up on sharding/write-side question, added in commit `389bb5d`.

Findings from reading the code:
- ONNX external-data path (`ir.save(..., external_data="model.onnx.data")`) does **not** support sharding today; it writes a single raw `model.onnx.data` file. So the current `--max-shard-size` restriction to safetensors is an API limitation/incidental implementation constraint, not an ONNX-format impossibility.
- Existing safetensors sharding in `onnx_ir.save_safetensors` split tensors into shards but wrote shards sequentially. It also failed in this environment with `safetensors==0.8.0` because the low-level `serialize_file` API now expects `TensorSpec` objects.
- Safetensors external data is standard ONNX external data in the sense that the ONNX initializers point at byte offsets inside the `.safetensors` files. I verified a sharded `build-gguf --external-data safetensors --max-shard-size 10KB` output with:
  - `onnx.checker.check_model(...)`: OK
  - `onnx.load(..., load_external_data=True)`: OK
  - `onnxruntime.InferenceSession(..., providers=["CPUExecutionProvider"])`: OK
  This is the closest practical proxy for onnx-genai loadability because onnx-genai ultimately loads the ONNX through ORT.

What changed:
- Wired `--max-shard-size` through `mobius build-gguf`; it was documented in the request but was not actually accepted by that subcommand.
- Replaced the safetensors save path in `ModelPackage.save` with a compatibility saver using the public `safetensors.numpy.save_file` API.
- Sharded safetensors files are written with bounded parallel shard writers. Default is up to 4 writer threads; override with `MOBIUS_SAFETENSORS_WRITE_WORKERS`.
- Added a regression test that saves sharded safetensors, checks the shard/index files, and loads the model back through ONNX external-data loading.

Important route decision for the 27B conversion:
- I am **not** switching the final Gemma3 27B artifact to safetensors by default because mobius already warns that safetensors offsets are not guaranteed 256-byte aligned and can cause CUDA CUBLAS misaligned-address failures. This model is specifically for CUDA/on-demand/streaming experiments, so the safer final artifact remains ONNX raw external data.
- The optimized ONNX external-data build is already practical: the 27B conversion completed in `81.122s`, output `16.079 GiB`, `MatMulNBits=434`, `onnx.checker` OK.

Updated validation:
- `python -m pytest src\mobius\integrations\gguf\ src\mobius\_configs\ src\mobius\models\gemma3_text_test.py src\mobius\_model_package_test.py` -> `212 passed`
- Ruff check and format-check on touched files -> passed.

justinchuby:
Update on ONNX external-data sharding and the safetensors detour:

I verified the installed `onnx_ir 0.2.1` source before building on sharding. The released ONNX external-data path does not expose a shard-size API:

- `onnx_ir.save(model, path, format=None, external_data=None, size_threshold_bytes=256, callback=None)` accepts one `external_data` path and no shard-size parameter.
- `onnx_ir.external_data.convert_tensors_to_external(tensors, base_dir, relative_path, callback=None)` also takes a single `relative_path`.
- `onnx_ir.external_data.unload_from_model(...)` documents and implements conversion to "a single data file".
- Searching the installed package for `shard` finds only the safetensors implementation under `onnx_ir._safetensors`, not the ONNX external-data implementation.

So the honest current status is: ONNX external-data sharding may exist in unreleased onnx-ir work, but it is not present in the released 0.2.1 API available to this mobius environment. Per the updated direction, I did **not** pursue safetensors as the solution and pushed `6e51b2d`, which reverts the safetensors sharding commit.

What remains in this PR:
- GGUF config/tensor/Gemma3 quantized export fixes.
- Parallel/vectorized quantized GGUF repacking.
- Progress reporting for both repack and `ir.save` writes using the onnx-ir callback.

Validation after the revert:
- `python -m pytest src\mobius\integrations\gguf\ src\mobius\_configs\ src\mobius\models\gemma3_text_test.py src\mobius\_model_package_test.py` → 211 passed.
- `python -m ruff check ...` and `python -m ruff format --check ...` on touched files → passed.
- `python -m lintrunner ...` could not run in this environment because `lintrunner` is not installed.

justinchuby:
Update after installing onnx-ir from git main:

- Installed `onnx-ir @ git+https://github.com/onnx/ir-py.git@main` for validation. It reports `onnx_ir.__version__ == 0.2.2` from commit `a86e5048c608c1fd73d7e7685886f2ad87abbd7f` and `ir.save(..., max_shard_size_bytes=...)` is present.
- Added `--max-shard-size` support to `mobius build-gguf` and wired `max_shard_size_bytes` through the ONNX external-data save path.
- Added a runtime guard: requesting ONNX sharding with onnx-ir 0.2.1 raises a clear error explaining that onnx/ir-py PR #427 is required. Once onnx-ir releases this, mobius should replace that guard/documentation with a normal dependency lower bound.
- Updated CLI/API docs so `--max-shard-size` is no longer described as safetensors-only.
- Kept progress logging on the onnx-ir save callback; with PR #427 the callback index/total are global across shards, so no separate counter/index file is needed.

Validation:
- Targeted tests: `python -m pytest src\mobius\integrations\gguf\ src\mobius\_configs\ src\mobius\models\gemma3_text_test.py src\mobius\_model_package_test.py` -> `214 passed`.
- Ruff check/format-check on touched files -> passed.
- Rebuilt Gemma3 27B after cleaning the output dir first:
  `python -m mobius build-gguf C:\Users\justinchu\dev\models-gguf\gemma-3-27b-it-Q4_0.gguf --output C:\Users\justinchu\dev\models\gemma-3-27b-onnx --keep-quantized --dtype f16 --ep cuda --max-shard-size 5GB`
  Wall-clock: `324.35s` in this run.
- Final sharded ONNX output:
  - `model.onnx`
  - `model.onnx-00001-of-00004.data` = 4.640 GiB
  - `model.onnx-00002-of-00004.data` = 4.607 GiB
  - `model.onnx-00003-of-00004.data` = 4.607 GiB
  - `model.onnx-00004-of-00004.data` = 2.225 GiB
  - Total = 16.079 GiB
  - `MatMulNBits=434`
  - `onnx.checker.check_model(...)`: OK
- Verified `onnx-genai` can load the sharded model by adding a temporary local test in the sibling repo that calls `onnx_genai_ort::Session::new(..., C:\Users\justinchu\dev\models\gemma-3-27b-onnx\model.onnx, SessionOptions::default())`; `cargo test -p onnx-genai-ort --test sharded_external_data_tmp` passed. The temporary test file was removed after the check.

Operational note: onnx-ir's sharded writer raises `FileExistsError` on shard-name collisions with files the in-memory model does not own, so repeated large conversions should clean the output directory first.

justinchuby:
Final artifact/loadability update:

After the latest correction, I stopped treating unreleased ONNX external-data sharding as a prerequisite for the artifact. The model now left on disk is plain ONNX external data with a single `model.onnx.data` file, built without `--max-shard-size`:

`C:\Users\justinchu\dev\models\gemma-3-27b-onnx\model.onnx`

Verification:
- Files now present:
  - `model.onnx` = 1,350,474 bytes
  - `model.onnx.data` = 17,263,722,496 bytes
  - recursive total = 17,265,072,970 bytes = 16.079352 GiB
- `MatMulNBits=434`
- `GatherBlockQuantized=0` because this GGUF's embedding is Q6_K and remains f16
- external data location in the ONNX proto: `model.onnx.data`
- `onnx.checker.check_model(...)`: OK
- `onnx-genai` graph load: passed via a temporary local test in the sibling repo calling `onnx_genai_ort::Session::new(..., C:\Users\justinchu\dev\models\gemma-3-27b-onnx\model.onnx, SessionOptions::default())`; the temporary test file was removed afterward.

Full `onnx-genai` CLI/model-directory loading is not complete yet because the output directory does not contain `tokenizer.json` (and no `genai_config.json`/metadata bundle was emitted by `build-gguf`). `cargo run -q -p onnx-genai-cli -- show C:\Users\justinchu\dev\models\gemma-3-27b-onnx` fails with `tokenizer.json not found`. So the ONNX graph itself is loadable by onnx-genai's ORT session layer, but the directory is not yet a complete generation-ready onnx-genai package.

Sharding note for reviewers: onnx/ir-py PR #427 is merged but not in PyPI's onnx-ir 0.2.1. It adds `ir.save(..., max_shard_size_bytes=...)` for plain ONNX external data; no index file is needed because the `.onnx` proto stores `location`/`offset`/`length`. The PR motivation is important here: "Safetensors sharding is incompatible with ORT CUDA kernels due to uncontrolled tensor alignment." Until that API is released and mobius can set a normal lower bound, sharding should be treated as requiring onnx-ir newer than 0.2.1 rather than assumed.

justinchuby:
Added the zero-point fix to this PR in commit `592db0a` rather than opening a second PR.

What changed:
- GGUF Q4_0 and Q8_0 are now treated as requiring explicit zero-point initializers even though they are symmetric formats.
  - Q4_0 emits packed zp=8 (`0x88` for pairs of 4-bit blocks; `0x08` when only one block is present).
  - Q8_0 emits zp=128.
- Because `QuantizedLinear` and `QuantizedEmbedding` key zero-point parameters off `not qc.sym`, this makes both `MatMulNBits` and `GatherBlockQuantized` receive explicit `zero_points` for GGUF Q4_0/Q8_0 instead of relying on provider defaults.
- The code comment names the CUDA corruption failure: `GatherBlockQuantized` CPU/CUDA defaults diverge when `zero_points` is omitted.

Regression coverage:
- Added a single-node `GatherBlockQuantized` runtime test with qweight nibble value 10 and explicit zp=8. It asserts the dequantized output equals hand-computed `(q - 8) * scale`, then reruns with zp=0 and asserts that output no longer matches. This checks the value, not merely that the input is wired.
- Existing synthetic GGUF builder tests now assert Q4_0 `MatMulNBits` nodes have four inputs and Q4_0 quantized embeddings emit `model.embed_tokens.zero_points == 0x88` with `GatherBlockQuantized` four-input wiring.

Quantization types checked:
- Direct GGUF repackers already produce zero-points for all supported direct formats: Q4_0 (8), Q4_1 (per-block affine), Q4_K (requantized per-block affine), Q8_0 (128), Q1_0 (1; Tencent Q1_0 uses its custom zp path).
- Q5_0/Q5_1/Q5_K/Q6_K are not direct MatMulNBits repack targets in this path; when they appear in mixed presets they are dequantized and requantized to the common target, which produces explicit zero-points when the target is asymmetric.

Validation:
- `python -m pytest src\mobius\integrations\gguf\ src\mobius\_configs\ src\mobius\models\gemma3_text_test.py src\mobius\_model_package_test.py` -> `216 passed`.
- `python -m ruff check src\mobius\integrations\gguf\_builder.py src\mobius\integrations\gguf\_builder_test.py` -> passed.
- `python -m ruff format --check src\mobius\integrations\gguf\_builder.py src\mobius\integrations\gguf\_builder_test.py` -> passed.
- Freshly converted `C:\Users\justinchu\dev\models-gguf\qwen2.5-0.5b-instruct-q4_0.gguf` with `--keep-quantized --dtype f16 --ep cuda`; the resulting graph has `GatherBlockQuantized` with 4 inputs, all `MatMulNBits` nodes with 4 inputs, and 170 zero-point initializers.
- Ran greedy generation through ONNX Runtime CUDA after adding the installed NVIDIA CUDA/cuDNN package DLL dirs to the process DLL search path. Output: `The capital of France is Paris. It is the largest city in`, confirming the freshly converted 0.5B Q4_0 produces correct CUDA output without post-hoc patching.

Upstream ORT issue:
- Reported the CPU/CUDA default divergence as https://github.com/microsoft/onnxruntime/issues/31692.

justinchuby:
Closing this PR rather than merging/rebasing it: it was accidentally based on another in-flight feature branch and includes unrelated commits. I rebuilt the urgent zero-point correctness fix from current origin/main in #459. The remaining Gemma3 config and repack performance work should be reapplied semantically as separate clean-main PRs.

PR #459, opened 2026-08-06
Emit GGUF zero points explicitly

justinchuby:
## Summary

This is the clean-main rebuild of the urgent correctness fix from #458. It contains only the GGUF zero-point fix, based on current `origin/main`.

GGUF Q4_0 and Q8_0 are symmetric formats, but their dequantization formulas still require non-zero zero points:

- Q4_0: `(q - 8) * scale`
- Q8_0: `(q - 128) * scale`

Mobius previously marked those GGUF imports as symmetric, so `QuantizedLinear` and `QuantizedEmbedding` omitted the optional `zero_points` input for `MatMulNBits` and `GatherBlockQuantized`. That is not portable: `GatherBlockQuantized` defaults diverge between ORT CPU and ORT CUDA when `zero_points` is omitted, corrupting CUDA embeddings before the first decoder layer runs.

This PR makes GGUF Q4_0/Q8_0 emit explicit zero-point initializers, so both `GatherBlockQuantized` and `MatMulNBits` receive the intended values instead of relying on EP defaults.

## Regression test

Added a single-node `GatherBlockQuantized` runtime regression:

- qweight nibbles are all `10`
- scales are known constants
- explicit zero point is `8`
- output is asserted against hand-computed `(q - 8) * scale`
- rerunning with zero point `0` is asserted not to match

I also temporarily flipped the test's zero point to `0`; it failed with a 100% output mismatch, so the test checks the value and not just the presence of the input.

Synthetic GGUF tests now also assert:

- Q4_0 `MatMulNBits` nodes have the explicit fourth input
- Q4_0 `GatherBlockQuantized` has the explicit fourth input
- Q4_0 embedding zero points are packed as `0x88`

## Quantization types checked

Direct GGUF repackers already produce explicit zero-points for supported direct formats:

- Q4_0: zp=8
- Q4_1: per-block affine zp
- Q4_K: requantized per-block affine zp
- Q8_0: zp=128
- Q1_0: zp=1; Tencent Q1_0 uses its custom zp path

Q5_0/Q5_1/Q5_K/Q6_K are not direct `MatMulNBits` repack targets here; when they appear in mixed presets they go through dequantize+requantize/native-block fallback paths, which produce explicit zero-points where needed.

## Validation

- `python -m pytest src\mobius\integrations\gguf\_builder_test.py -q` -> `33 passed`
- `python -m pytest src\mobius\integrations\gguf\ src\mobius\_configs\ src\mobius\_model_package_test.py` -> `304 passed`
- `python -m ruff check src\mobius\integrations\gguf\_builder.py src\mobius\integrations\gguf\_builder_test.py` -> passed
- `python -m ruff format --check src\mobius\integrations\gguf\_builder.py src\mobius\integrations\gguf\_builder_test.py` -> passed

Fresh conversion validated with `C:\Users\justinchu\dev\models-gguf\qwen2.5-0.5b-instruct-q4_0.gguf`:

```powershell
python -m mobius build-gguf C:\Users\justinchu\dev\models-gguf\qwen2.5-0.5b-instruct-q4_0.gguf --output C:\Users\justinchu\dev\models\qwen2.5-0.5b-q4_0-mobius --keep-quantized --dtype f16 --ep cuda
```

Converted graph verification:

- `GatherBlockQuantized` input count: 4
- all `MatMulNBits` input counts: 4
- zero-point initializers: 170

Generation checks on the freshly converted model, no post-hoc graph patching:

- ORT CPU via onnx-genai CLI: `The capital of France is Paris. It is the largest city in`
- ORT CUDA via onnx-genai CLI (`ONNX_GENAI_ORT_LIB_DIR` pointed at the installed onnxruntime-gpu package, CUDA/cuDNN DLL dirs on PATH): `The capital of France is Paris. It is the largest city in`
- native CUDA via onnx-genai CLI (`--features native-cuda`, CUDA/cuDNN DLL dirs on PATH): `The capital of France is Paris. It is the largest city in`

## Upstream ORT issue

Reported the provider default divergence here: https://github.com/microsoft/onnxruntime/issues/31692

## Relationship to #458

#458 was accidentally based on another in-flight feature branch and includes unrelated commits. This PR is the clean-main replacement for the urgent zero-point correctness fix only.

### DIFF
--- a/src/mobius/integrations/gguf/_builder.py
+++ b/src/mobius/integrations/gguf/_builder.py
@@ -528,7 +528,7 @@ def _detect_quant_params(gguf_model, gguf_arch: str) -> tuple[int, int, bool]:
         map_gguf_to_hf_names,
     )
 
-    # Symmetry of each supported GGUF quantization type.
+    # Whether the graph can omit zero_points for each supported GGUF type.
     #
     # Mainline Q1_0 (1-bit binary) is repacked into 2-bit MatMulNBits
     # with zp=1 — see _repack_q1_0. Tencent's custom Q1_0 (2-bit SEQ,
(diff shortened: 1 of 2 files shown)


justinchuby:
@copilot do not use onnx or helper module, use onnx-ir instead for building the test model. Also fix lint and unit tests

PR #465, opened 2026-08-08
Add Qwen Image Edit 2509 pipeline support

justinchuby:
## Summary

- add faithful `QwenImageEditPlusPipeline` detection for `Qwen/Qwen-Image-Edit-2509`, including its Qwen2.5-VL image-aware prompt encoder, Qwen Image transformer, normalized 3D VAE, processor assets, and FlowMatch Euler scheduler configuration
- align the denoiser with diffusers packed target/source latent tokens, non-contiguous text masks, text/image 3D RoPE, 256-wide timestep projection, target-token output slicing, and fp16 clipping
- add dedicated prompt-encoding, denoising, and edit-VAE tasks while preserving direct diffusers weight names
- emit onnx-genai image-edit runtime metadata for prompt templates, image sizing, latent packing, source conditioning, RoPE inputs, dynamic scheduler shifting, and component dataflow
- add official-config graph coverage plus reduced-config numerical, dtype, CUDA, real-source-image, and deterministic end-to-end diffusion-loop coverage
- rebase onto current `main` and adopt its fused `Swish` implementation; fold diffusers weight transposes immediately after loading and prevent inappropriate causal/GQA optimization of the bidirectional VAE reconstruction graph

## Validation

- rebased onto `origin/main` at `1e6f61f`
- `lintrunner -a` — full repository lint clean
- full non-integration suite — 3,872 passed, 58 skipped
- Qwen Image integration/CUDA selection — 6 passed
- schema/model coverage sweep — 923 passed, 217 skipped
- official `Qwen/Qwen-Image-Edit-2509` fp16 CUDA component CLI export with `--runtime onnx-genai` — succeeded and emitted `model.onnx`, FlowMatch metadata, tokenizer, Qwen2-VL image/video processor configs, and chat template
- RTX A1000, onnxruntime-gpu 1.28.0 CUDAExecutionProvider — profiled CUDA node execution on reduced configurations; fp32 CPU/CUDA parity passed; fp16 and bf16 outputs matched reduced diffusers references at `rtol=atol=1e-2`
- CUDA verbose partitioning — zero inserted Memcpy nodes

### Graph optimization evidence

A two-block fp16 CUDA Mobius graph after the standard optimization pipeline contains:

- 2 `Attention` nodes (one per layer); standard Attention is required because Qwen Image Edit needs an explicit padding bias, which correctly disqualifies Flash and contrib GQA
- 2 CSE-shared `Swish` nodes and 0 decomposed `Sigmoid` nodes
- 6 `SkipLayerNormalization` nodes
- 3 standard fused `Gelu` nodes
- 0 `Identity` nodes

After weights become constants, Mobius post-weight folding reduces `Transpose` from 39 to 8 and total nodes from 391 to 360; the eight remaining transposes are dynamic RoPE/attention layout changes, not weight-layout overhead. Diffusers pipeline loading now performs this fold for every component instead of leaving Linear weight transposes in the saved graph.

ORT CUDA `ORT_ENABLE_ALL` further produces 3 `FastGelu` nodes, keeps the 6 `SkipLayerNormalization` nodes, folds `Swish` into surrounding kernels, retains only 8 transposes, and inserts 0 memcpy nodes. These counts are locked by model regression coverage where the representation is controlled by Mobius.

The official-config, no-weight transformer graph remains 60 Attention layers with 0 Identity nodes; post-weight folding is validated on a reduced identical module tree because downloading the 57+ GB checkpoint is waived below.

## Golden provenance and semantics

The committed golden is a genuine independently generated **reduced Qwen Image Edit diffusion pipeline reference**, but it is **not generated with the official 20.4B checkpoint weights and is not an official-checkpoint perceptual output**.

Reference generation uses diffusers 0.39.0 with a seeded one-block `QwenImageTransformer2DModel` and seeded reduced `AutoencoderKLQwenImage` whose weights are copied into the separate Mobius ONNX graphs. The source is the real nonzero `testdata/pipeline-cat-chonk.jpeg`:

1. diffusers resizes and encodes the JPEG through the reduced Qwen Image VAE
2. it takes the posterior mode, applies Qwen latent mean/std normalization, and performs official 2x2 latent packing
3. it concatenates those source-image tokens after independently seeded target latent tokens
4. it invokes the diffusers transformer with deterministic prompt embeddings, a non-contiguous prompt mask, and image/text 3D RoPE
5. it advances the target through three independent `FlowMatchEulerDiscreteScheduler` steps
6. it unpacks, denormalizes, and decodes the final target through the reduced diffusers VAE

The test then independently runs the Mobius chain: ONNX VAE encoder -> separate NumPy 2x2 packing -> ONNX denoiser plus a separate FlowMatch scheduler -> NumPy unpacking -> ONNX VAE decoder.

- **L4 (`l4_noise_pred`)**: all 256 target-token flow-prediction values from the first actual diffusers transformer call. Both the diffusers reference and ONNX output are compared elementwise to the committed data.
- **L5 (`l5_final_latents`, `l5_final_image`)**: all 256 target latent-token values after three actual diffusers transformer/scheduler steps and all 768 clipped decoded image values from the final reduced diffusers VAE decode. Mobius final latent tokens and decoded image are compared elementwise.
- **Prompt scope**: prompt embeddings are deterministic synthetic tensors so this reduced numerical golden isolates the image-edit diffusion chain. The official Qwen2.5-VL prompt-encoder split is covered separately by graph/config/runtime-contract tests; no claim is made that the golden exercises official prompt-encoder weights.

This proves real-image VAE conditioning, packing, transformer invocation, multi-step scheduling, and decode semantics without fake token-generation data. It does **not** claim official-weight image quality or official-checkpoint end-to-end parity.

## Waivers

- **Official-weight L3/L4/L5 and perceptual E2E:** the checkpoint contains 20.4B parameters and over 57 GB of weights, exceeding the RTX A1000's 8 GB VRAM and the available validation budget. Official component configs and the official-size graph/export path are validated without weights; numerical parity and committed goldens use the genuine reduced pipeline described above.
- **ORT GenAI load:** onnxruntime-genai does not expose a diffusion/image-edit pipeline runtime. The supported onnx-genai iterative runtime sidecar is exported and tested instead.
- **Foundry Local / Olive quantization:** no Qwen Image Edit diffusion runtime or applicable quantization path is available in this environment; these are not claimed as validated.


src/mobius/models/qwen_image_test.py
```diff
@@ -0,0 +1,512 @@
...
+        expected = hf_model(
+            hidden_states=torch.from_numpy(feeds["sample"]),
+            timestep=torch.from_numpy(feeds["timestep"]),
+            encoder_hidden_states=torch.from_numpy(feeds["encoder_hidden_states"]),
+            encoder_hidden_states_mask=torch.from_numpy(feeds["encoder_hidden_states_mask"]),
+            img_shapes=[image_shapes],
+        ).sample[:, :4]
+
+    with tempfile.TemporaryDirectory() as directory:
+        path = f"{directory}\\model.onnx"
```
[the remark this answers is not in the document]
justinchuby: Fixed in 4cef061: all temporary model and profiling paths now use os.path.join, preserving TemporaryDirectory isolation on POSIX and Windows. Direct Ruff format/check and the Qwen Image test module pass.

src/mobius/tasks/_qwen_image_vae.py
```diff
@@ -74,3 +74,58 @@ def _build_decoder_graph(
...
+
+
+class QwenImageEditVAETask(QwenImageVAETask):
+    """Build VAE graphs with Qwen Image Edit latent normalization embedded."""
+
+    def _build_encoder_graph(self, module, config: QwenImageVAEConfig) -> ir.Model:
+        if config.latents_mean is None or config.latents_std is None:
+            raise ValueError(
+                "Qwen Image Edit VAE config requires latents_mean and latents_std"
+            )
```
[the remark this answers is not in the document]
justinchuby: Fixed in 4cef061: QwenImageEditVAETask now validates that both latent-stat arrays exist and have exactly z_dim entries before either graph is constructed, with a regression test for the mismatch case.

src/mobius/models/qwen_image_test.py
```diff
@@ -0,0 +1,512 @@
+# Copyright (c) Microsoft Corporation.
```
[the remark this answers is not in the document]
justinchuby: Applied the formatter and pushed 4cef061. Full `lintrunner -a` is clean, and direct Ruff format/check reports the reviewed files are formatted with no lint findings.

src/mobius/_diffusers_builder.py
```diff
@@ -318,16 +348,29 @@ def build_diffusers_pipeline(
 
         model_module = module_class(config)
 
-        sub_pkg = build_from_module(model_module, config, task_name)
+        if (
+            pipeline_class == "QwenImageEditPlusPipeline"
```
justinchuby: This doesn't look general enough

src/mobius/_diffusers_builder.py
```diff
@@ -344,4 +387,24 @@ def build_diffusers_pipeline(
...
+            if "scheduler" in pipeline_index
+            else {}
+        ),
+        processor_config=(
+            _load_optional_diffusers_json(model_id, "processor/preprocessor_config.json")
+            if "processor" in pipeline_index
+            else {}
+        ),
+        model_type=(
+            "qwen_image_edit" if pipeline_class == "QwenImageEditPlusPipeline" else "diffusers"
```
justinchuby: Also this

src/mobius/models/qwen_image.py
```diff
@@ -389,7 +515,7 @@ def __init__(self, config: QwenImageConfig):
         self.img_in = _Linear(config.in_channels, hidden_size)
         self.txt_norm = _RMSNorm(config.joint_attention_dim, eps=config.norm_eps)
         self.txt_in = _Linear(config.joint_attention_dim, hidden_size)
-        self.time_text_embed = _TimestepEmbedding(hidden_size, hidden_size)
+        self.time_text_embed = _TimestepEmbedding(256, hidden_size)
```
justinchuby: Why

src/mobius/integrations/onnx_genai/auto_export.py
```diff
@@ -106,6 +107,106 @@ def _write_clip_tokenizer(output_dir: str, source: str | None) -> str | None:
...
+                    "to": "text_encoder.inputs_embeds",
+                },
+            ]
+        )
+    phases = dict(pipeline.get("phases", {}))
+    for name in ("text_encoder_vision_encoder", "text_encoder_embedding", "vae_encoder"):
+        if name in names:
+            phases[name] = {"run_on": "prompt_only"}
+    pipeline["phases"] = phases
+    pipeline["image_edit"] = {
```
justinchuby: This `image_edit` block is descriptive metadata only: no executable dataflow connects the VAE encoder through latent packing/source-token concatenation to the denoiser, nor unpacks rank-3 target tokens before the rank-5 VAE decoder. The generic scheduler loop also assumes sample/output shapes match, while this denoiser consumes target+source tokens and returns target-only tokens. Please add runtime-supported transforms/edges and an actual package execution test, or reject this export rather than emit a non-runnable pipeline.

src/mobius/integrations/onnx_genai/auto_export.py
```diff
@@ -415,13 +521,22 @@ def write_onnx_genai_config(
...
             kwargs.setdefault(name, value)
+        if is_qwen_image_edit:
+            kwargs["text_encoder_output"] = "prompt_embeds"
+            kwargs["text_encoder_edges"] = [
+                ("prompt_embeds", "encoder_hidden_states"),
+                ("prompt_embeds_mask", "encoder_hidden_states_mask"),
+            ]
         # Classic text-conditioned diffusion (a text encoder is present) uses
         # classifier-free guidance by default; SD's canonical scale is 7.5.
         if guidance_scale is None and "text_encoder_filename" in kwargs:
```
justinchuby, line 469: This applies Stable Diffusion CFG (`7.5`) to Qwen Image Edit by default, but the official pipeline defaults guidance to disabled and optional true CFG requires independently encoded negative prompt embeddings/masks, a second transformer pass, and norm rescaling. Zeroing only the primary conditioning tensor runs a different algorithm. Please disable generic CFG for this model unless the Qwen-specific two-pass path is implemented.

src/mobius/integrations/onnx_genai/inference_metadata.py
```diff
@@ -1718,7 +1749,9 @@ def from_diffusers(cls, config: dict[str, Any]) -> SchedulerConfig:
         """
         raw_name = str(config.get("_class_name", ""))
         name = raw_name.lower()
-        if "eulerancestral" in name:
+        if "flowmatcheuler" in name:
+            kind = "flow_match_euler"
```
justinchuby, line 1753: The emitted scheduler identifiers do not match onnx-genai: the runtime registers `flow_matching`, not `flow_match_euler`, and accepts `flow`/`velocity` rather than diffusers’ `flow_prediction`. It also does not consume these dynamic-shifting fields as emitted, so package loading or scheduling is incorrect. Please map to supported runtime identifiers and implement the sequence-length-derived shift behavior, or emit an explicit correct sigma/timestep schedule.

justinchuby (commented):
Correctness review: three onnx-genai runtime findings.

src/mobius/models/qwen_image.py
justinchuby: Clarified in b922438. The 256-wide sinusoidal timestep embedding is fixed by the upstream Qwen Image transformer architecture; `config.timestep_scale` scales values but does not define the embedding width.

src/mobius/integrations/onnx_genai/auto_export.py
justinchuby: Fixed in b922438. `--runtime onnx-genai` now fails explicitly before emitting an unusable package. Direct ONNX component export remains supported.

src/mobius/_diffusers_builder.py
justinchuby: Fixed in b922438. Pipeline model-type overrides are now data-driven through `_PIPELINE_MODEL_TYPE_OVERRIDES`, so this path is no longer hard-coded to Qwen Image Edit.

src/mobius/_diffusers_builder.py
justinchuby: Fixed in b922438. Component-specific task class overrides now live in a data-driven `_PIPELINE_COMPONENT_TASK_CLASSES` mapping instead of a Qwen-specific conditional.

src/mobius/integrations/onnx_genai/auto_export.py
justinchuby: Fixed in b922438 by rejecting onnx-genai runtime export. We no longer advertise or emit an SD-style CFG path that would be semantically wrong for Qwen Image Edit; direct component export is unchanged.

src/mobius/integrations/onnx_genai/inference_metadata.py
justinchuby: The mapping is retained as forward-looking metadata, but b922438 rejects runtime export because onnxruntime-genai currently has no diffusion executor, iterative denoise loop, or scheduler consumer. The required FlowMatch/Qwen Image Edit orchestration is documented upstream in microsoft/onnxruntime-genai#2114: https://github.com/microsoft/onnxruntime-genai/issues/2114#issuecomment-5255356626.

PR #467, opened 2026-08-09
Fix prefill pruning for Gemma 4 models

sushraja-msft:
Gemma 4 models have shared KV cache 
- For the E2B model the KV cache of the first 15 layers are shared with the remaining 20 layers. 
- This means during prefill there is no need to compute beyond 15 layers, only decode needs to run the full graph. Effectively 2B model is a 0.6B model for prefill purposes. 
- Without this optimization prefill perf will suffer on CPU/GPU.
This change extends the prune_lm_head concept, now instead of pruning at the lm_head level - gemma models prune at a much higher layer, at level 15 onwards. 

The brings the prefill TTFT from 845ms on a Nvidia 4070 to 183ms.

Mobius is also fixed to generate graph capture compatible models for Gemma 4.


justinchuby (approved):
Thank you!

sushraja-msft:
Closing because the CI system doesnt quite work for Prs from forks

PR #470, opened 2026-08-10
Add vision and audio support for Gemma 3n

NishantN2005:
## What this does

Makes `google/gemma-3n-E4B-it` export as a real 4-model multimodal package
(`decoder` + `vision_encoder` + `audio_encoder` + `embedding`) instead of a
text-only decoder.

Before this, both `gemma3n` and `gemma3n_text` mapped to
`Gemma3nCausalLMModel`, so `Gemma3nCausalLMModel.preprocess_weights` **deleted**
every `vision_tower.`, `audio_tower.`, and `multi_modal_projector.` key. The
export succeeded and produced a package that could not accept images or audio at
all. `gemma3n` now maps to a new `Gemma3nMultiModalModel` with `task="gemma3n"`,
matching how `gemma3` / `gemma3_text` are already split.

## Two pre-existing text-path bugs fixed first

These affect the **already-shipping text export** and are independent of the
modality work, so they land as their own commits at the base of the branch. Both
silently produced wrong numbers rather than crashing.

- **KV layer sharing was not implemented** (`b9c939a`). E4B sets
  `num_kv_shared_layers=15`, so layers >= 20 reuse the KV of the last non-shared
  layer *of the same type* (sliding -> layer 18, full -> layer 19). Attention
  always recomputed `k_proj`/`v_proj`. The checkpoint does ship `k_proj` for all
  35 layers, so nothing failed to load -- the graph just computed KV that HF
  never uses, and the exported cache had 35 entries where HF effectively has 20.
- **Activation sparsity (`_gaussian_topk`) was missing** (`2989391`). For E4B
  layers 0-9 `activation_sparsity_pattern` is 0.95. HF computes
  `cutoff = mean + std * icdf(0.95)` over the last dim and applies
  `relu(gate_proj - cutoff)` before the activation. `icdf(0.95)` is a
  compile-time constant, so no ONNX erf/ppf is needed.

Both were previously invisible because `tests/synthetic_parity_test.py` pinned
`num_kv_shared_layers: 0` for gemma3n and never set a sparsity pattern. Those
escape hatches are removed here, so the existing HF-vs-mobius parity harness now
covers both.

## Architecture notes

Three places where gemma3n does *not* match the obvious existing component, and
reusing it would have been wrong:

- **The vision tower is MobileNet-V5-300m, not SigLIP.** HF routes
  `gemma3n_vision` to `TimmWrapperModel`, and timm is not a dependency. New
  `components/_mobilenetv5.py` implements the stem, EdgeResidual / UIB /
  UIB+MQA blocks, and the Multi-Scale Fusion Adapter, driven by a hard-coded
  per-stage spec table -- there is no config source for the block layout.
- **The 116 `bn.weight` tensors are timm `RmsNorm2d`, not BatchNorm.** The
  checkpoint ships **zero** `bn.bias` / `running_mean` / `running_var` /
  `num_batches_tracked`. Using the existing `components.BatchNorm2d` would have
  demanded four initializers that do not exist.
- **The audio tower reuses `_gemma4_audio.py` except for the norm.** gemma3n's
  SSCP conv blocks use `Gemma3nAudioCumulativeGroupNorm` -- group norm computed
  *cumulatively over the time axis* -- where gemma4 uses plain `LayerNormNoBias`.
  That needs `CumSum` over T. gemma3n also adds an explicit
  `relative_position_embedding.pos_proj` that gemma4 lacks.

`components/_rms_norm.py` gains a shared scale-free RMSNorm, promoted out of
`models/gemma4.py` rather than duplicated.


justinchuby:
Thanks! Could you resolve conflicts?

NishantN2005:
> Thanks! Could you resolve conflicts?

I do not have permission to push anymore, looks like I was removed from the "mobius-external-contributors" team.

justinchuby:
@NishantN2005 sorry that was due to a github team clean up. Could you fork this repo and create a new PR? Now that the repo is public you should be able to do so. Sorry for the churn!

justinchuby:
Replaced by #474

PR #477, opened 2026-08-11
Ensure model package type correctness

justinchuby:
Ensure model package type


rui-ren (approved):

justinchuby:
@copilot can you instead uphold the invariant that the graph is available, and have a test to guard all new models?

PR #478, opened 2026-08-12
Simplify composite ONNX GenAI metadata

justinchuby:
## Summary

Produce the canonical ONNX GenAI `pipeline.workflow` ABI across Mobius exports and validate it against [onnx-genai#828](https://github.com/justinchuby/onnx-genai/pull/828).

- remove serialized `model.io`, legacy phase/strategy duplication, and runtime-private scheduling/allocation policy;
- emit typed SSA workflows for decoder generation, static cache, speculative decoding, adapters, VLM, diffusion/image edit, video, speech, codec/TTS, masked diffusion, ESM-2, and ProtBert;
- describe graph-visible state semantics including append, indexed-scatter, and fixed-size recurrent replacement;
- emit canonical encoder embedding profiles and application-sourced opaque auxiliary inputs;
- assign deterministic symbols to anonymous dynamic ONNX dimensions instead of invalid YAML `null`;
- stop emitting duplicated workflow `ir_version` and `onnx_opsets`, leaving schema and ONNX artifacts authoritative;
- pin cross-repository CI to ONNX GenAI commit `509cd4e9c4471f4cbc59fe44b47168f0ae128fe3`;
- merge current `main` into the branch without rebasing.

## Validation

- 164 metadata producer tests passed, 4 skipped;
- all 13 generated package families match committed fixtures and pass the authoritative `validate_metadata` parser/semantic validator;
- 11 ONNX GenAI engine conformance tests passed using real generated Mobius packages;
- changed files pass lintrunner formatting.


justinchuby:
Cross-repo contract audit against onnx-genai `4c3c4b6` found these concrete mismatches in `487d6ca`:

1. Metadata emits `fp32` while the workflow schema/validator uses `float32` (also `float16`/`bfloat16`). Please emit canonical workflow dtype names.
2. Seeded sampler currently uses scalar `temperature`, `seed`, `counter`, and `next_counter`; the published contract requires batched `temperature[B]` and counter RNG `seed[B]`, `offset[B]`, `next_offset[B]`.
3. EOS graph has only `token_ids`/`eos_token_ids -> terminated`; the current termination role also requires `iteration` and `max_iterations`. I will split/relax the runtime contract so EOS predicate and loop max remain independently composable rather than forcing limit math into EOS.
4. Euler graph exposes `sample`, `derivative`, scalar `sigma`, `sigma_next`; the current role's `step`/`schedule` mapping cannot describe both sigma values. I will revise the semantic role to explicit current/next schedule values.
5. Masked update needs semantic mappings for `confidence` and `threshold`; the current role's integer `step` is not equivalent. I will revise this contract.
6. Speculative artifact consumes probabilities/uniform and returns acceptance mask/count, whereas the current verifier role expects score tensors, accepted tokens, and done. I will revise the role to primitive acceptance math; token correction/state update remains separate workflow components.
7. Token state update grows `[B,T] -> [B,T+1]`; current validation incorrectly requires identical current/next shapes. I will change it to dtype/rank compatibility and rely on state recurrence bounds.
8. `add_policy_components_to_workflow` currently emits ports only, without each component's required `policy` semantic mapping/effect declaration. Those must be emitted once the revised fields land.

Please continue making sampler RNG batched/counter-based and canonicalizing dtypes. I will push the contract corrections and runtime E2E fixtures to #828, then post the new SHA here for exact producer alignment.

justinchuby:
Runtime E2E review of `dd1cb838` found three remaining blocking mismatches in `build_decoder_workflow_metadata`:

1. Decoder `logits` is rank 3 `[B,T,V]`, but both sampler artifacts declare rank 2 `[B,V]`. onnx-genai now validates SSA edge contracts and will reject this package. Please ship/invoke an ONNX final-position selector (or make the sampler artifact explicitly accept rank-3 logits and select the last position); the runtime will not add host slicing.
2. Workflow loop semantics are **continue while `condition` is true**. The generated graph binds `condition: loop.done`, which stops after the first ordinary token and continues after EOS. Please invoke an ONNX boolean inverter and bind a `continue` value, or change the termination artifact to expose continuation explicitly.
3. `loop.iteration` is a fixed application input and is never loop-carried/updated. The termination graph's max-iteration predicate therefore does not advance. The host loop bound still caps execution, but the policy output is wrong. Please add an ONNX counter state-update component/carry or remove the redundant limit predicate from the termination artifact in a coordinated contract revision.

Also, emitted `tokens` is declared rank-1 `[B]` with `mode: append`; that only represents a batch-1 flat stream. For batched output the boundary must define `[B,T]` accumulation (or use per-row event emission with the serving contract). I am making the runtime fail rather than silently inspect only batch row 0.

justinchuby:
ONNX GenAI branch phi/effect-merge blocker is fixed at `ad8e5e4f1a766eb95787c97266a6ed554e1b28ec` on PR #828.

Emit branches with:
```yaml
outputs:
  selected.tokens:
    cases: { "true": accepted.tokens, "false": corrected.tokens }
effects:
  rng:
    incoming: rng.0
    cases: { "true": rng.accepted, "false": rng.corrected }
    produces: rng.joined
```
Include `default` in each output/effect mapping whenever the branch has a default node. All phi contracts must unify. Case-local successor tokens may differ (or match each other), but `produces` is a distinct joined token. Only phi outputs and explicit emits escape branch scope.

Exact contract/docs: `crates/onnx-genai-metadata/src/schema/ir.rs` and `docs/WORKFLOW_POLICY_COMPONENTS.md`; generated schema is updated. Speculative token/KV/RNG/emit E2E passes.

justinchuby:
ONNX GenAI preprocessing-to-workflow SSA blocker is fixed at `justinchuby/onnx-genai@f72782c64fbe04ac6e12593a69ecff47781f023b` (draft PR #828).

Producer target:
- declare adapter component `{ kind: adapter, abi: onnx-genai.image-preprocess, version: "1" }` and pin `manifest.adapter_abis["onnx-genai.image-preprocess"]: "1"`
- adapter input `encoded`: `uint8` rank 1 `[encoded_bytes]`
- add exact `TensorContract` to every `preprocessing.image.outputs[]`; `source` is processor-local and required; `name` is the workflow SSA value
- invoke with explicit port maps, e.g. `outputs: { pixel_values: image.pixel_values, grid: image.grid }`
- output cannot be optional in workflow metadata
- feed those SSA values to vision/embedding/decoder invokes; do not emit legacy `component.input` names

Canonical documentation: `docs/WORKFLOW_POLICY_COMPONENTS.md` (Versioned adapter invocation). Schema: `schema/inference_metadata.schema.json`. Runtime E2E covers encoded request image -> preprocessing adapter -> pixel/grid SSA -> vision -> embedding -> decoder -> post-adapter emit. Branch phi/effect support remains at prior commit `ad8e5e4f1a766eb95787c97266a6ed554e1b28ec`.

justinchuby:
Follow-up audit of current Mobius head `151014eb` found remaining generated-workflow blockers:

1. Decoder state declaration still says `state.token.initializer: request.<token_input>` (`[B,T]`) while the state contract/current value is `[B,1]`. ONNX GenAI now validates declared state initializers statically and against the concrete value at loop entry (`21f6994d`). Set the initializer to the setup-produced `[B,1]` token value (`token.setup`) or restructure setup.
2. `loop.iteration` is still a literal zero and is not updated/carried. The termination max-iteration input therefore remains semantically wrong. Use an ONNX counter/state-update component and carry its `[B]` output; no host loop-index fallback will be added.
3. Decoder setup samples/emits one token, then the loop runs `max_iterations` body executions. That can emit `max_output_tokens + 1`; setup EOS also cannot prevent the first body execution because the declared loop condition is produced in the body. Either make setup initialization-only and perform all output steps in the body, or ship ONNX policy components/dataflow that compute the remaining body budget and gate the first body.
4. The same setup-plus-body accounting should be checked in masked diffusion: setup performs one masked update, then the body may run the full requested iteration count.

TTS does not need a new host induction primitive: per the component-centric contract, initialize an integer state cell and invoke/carry a generic ONNX counter update inside the nested loop; use that SSA value for `step_index`/position selection.

Please generate one concrete package/YAML at the fixed head so I can run ONNX GenAI load+execution cross-repo.

justinchuby:
ONNX GenAI loop induction blocker is fixed in `justinchuby/onnx-genai@e877270123d38d49012e70ecbf7254e720cd3431` (PR #828).

Producers may now declare:
```yaml
iteration:
  value: loop.i
  contract: { dtype: int64, rank: 1, shape: [batch] }
```
and bind `loop.i` directly to solver/masked-update/termination `step`, RNG counters, schedule selectors, or TTS `step_index`. It is zero-based, body/condition-scoped, deterministic, and lexically distinct for nested loops. No ONNX counter component is needed for the control-flow induction variable; reverse/remaining values still require an ordinary ONNX component.

Please migrate decoder/diffusion/masked/TTS loop iteration inputs to this field. The separate decoder setup/body output-count and initializer issues from the prior comment still need producer fixes before cross-repo execution.

justinchuby:
Cross-repo execution against ONNX GenAI #828 found one remaining producer-side contract mismatch (still present at current head `8f96651c0031ed082229ee404210c50de4a55216`):

- Decoder/VLM workflows carry `decoder.setup.logits` as invariant `state.logits`, then update it from `decoder.body.logits` (`workflow_metadata.py` around lines 1150-1195 and 2041-2105). A normal decoder produces setup logits `[B, prompt_sequence, V]` and body logits `[B, 1, V]`, so the generic runtime correctly rejects the first recurrence as an invariant shape change.
- Please carry last-token logits instead: invoke `last_token_logits` after the setup decoder, initialize a rank-2 `[B,V]` state from that result, invoke it after each body decoder, and carry the rank-2 result; alternatively restructure the loop so full rank-3 logits are not a carried invariant.

Other exact `774448b` generated-package results:
- masked diffusion executes end-to-end with the actual Mobius policy artifacts;
- codec executes end-to-end;
- decoder executes when the synthetic decoder is constrained to `[B,1,V]`, confirming request binding, invoke, loop, KV growth, effects, and emit paths.

ONNX GenAI fixes discovered by this run: scalar defaults now materialize unbound symbolic axes as singleton tensors, and component shape symbols are invocation-local (with adapter allocation retaining package-scope symbols). Permanent conformance/regression tests are being pushed to #828.

justinchuby:
Fixed the decoder/VLM logits recurrence in `87bd5b4`: setup and body decoder outputs are normalized through the generated `last_token_logits` component, and only rank-2 `[B,V]` values are carried invariantly. I generated a decoder with setup `[B,T,V]` and body `[B,1,V]` and ran ONNX GenAI conformance at `2ad2e0d`; `mobius_decoder_workflow_executes` passed for 3 tokens. Diffusion also executed end-to-end for two Euler iterations on the same runtime. Muse metadata was regenerated and validates semantically; local SHA-256 is `f1a33d7f162c350e79a37d11d8be5be0690a3c8c6cc0e71ce2b428c368dccabf`.

justinchuby:
Remaining cross-repo blockers after `87bd5b4`: (1) real speculative rejection cannot be represented faithfully because `WorkflowNode::Emit` has no accepted-length/validity operand, while verifier KV state also needs rollback/truncation to the accepted prefix; the current fixed-shape branch fixture is schema-valid but is not a correct rejection execution. (2) Real Qwen3-TTS still needs talker/code-predictor KV recurrence and the trained group-0/predictor embedding transition; the current representative nested-loop fixture validates structure but is not a real-package runtime E2E. Independent decoder, masked diffusion, codec, and Euler diffusion runtime paths pass; VLM/Muse validates against the latest schema.

justinchuby:
Performance acceptance instrumentation is pushed in `e864abe`.

The new paired-run gate rejects non-identical model/runtime/EP/device/precision/batch/shape/sampling/RNG/KV/capture/warmup conditions, then checks throughput, TTFT, peak memory, H2D/D2H counts+bytes, device syncs, session/kernel boundaries, device residency, and required island capture/replay. Required plans cover decoder+min-p+termination, speculative accept/reject, and grammar-delimited islands.

Current measured upstream baseline at ONNX GenAI `8bacf8c` is **not release-passing**:
- H200 decoder steady throughput workflow/native: `0.903` (9.7% regression).
- H200 min-p throughput: `0.957` (within 5%), but warm TTFT `4.36/3.64 ms` (19.8% regression).
- Cold startup: decoder `467/49 ms`, min-p `231/18 ms`, rooted in first-run output-extent discovery/stable-binding construction.
- Synthetic decoder/min-p islands captured once/replayed 503 times; speculative verifier/policy also captured/replayed.

Real Mobius package/KV/per-row serving measurements are still absent. Additionally, `application_overridable` samplers remain excluded by `is_fusible_component`, so the actual Mobius overrideable sampler path cannot yet demonstrate the required fused island. PR remains draft/not performance-ready.

justinchuby:
Architecture cleanup progress:

- `66e5281`: removed the closed `PolicyRole` enum. Policy ONNX artifacts now embed versioned namespaced contract IDs, and workflow declarations derive ID/version from artifact metadata.
- `ada6303`: decoder and masked-diffusion loops now consume lexical loop induction directly; generated iteration state and `iteration_increment.onnx` are removed. Serialized invokes contain semantic tensor mappings only, with no effect-token fields/read-write chains.
- Full Mobius suite: 3694 passed, 52 skipped; Ruff passed.

Muse concise metadata currently measures **600 lines / 546 mapping fields / 16,339 bytes**, down from the reviewed 1,183-line source (**583 lines, 49.3% reduction**). A pre-lowering flattened-field baseline was not retained, so only the current 546-field count is evidence-backed.

Remaining producer work is blocked on exact current #828 contracts, reported at issuecomment-5274711168: per-row `Emit.valid_length[B]`/ragged append and logical state lengths, `continue_when` polarity, generic KV alias/group/sequence-axis semantics, and a shipped semantic-validator CLI. Current `8bacf8c` still requires one-element emit length, so removing speculative `ReduceMin` now would create metadata the shipped validator/executor rejects. Legacy schema fixtures/continue-Not paths will be deleted when those contracts land; PR remains draft.

justinchuby:
ONNX GenAI producer blockers are resolved at `justinchuby/onnx-genai@8215649100a0a27be15b04045fddde777c8248fc` on PR #828. Producer migration can resume.

Use `schema/inference_metadata.schema.json` from that commit and run the shipped package validator:
`cargo run -p onnx-genai-metadata --bin validate_metadata -- <generated-package-dir>`

Key migration: `continue_when` is pre-test/zero-trip; emit `valid_length: int64[B]` without reduction and consume `output.row.<row>`; declare serving `active/done/accepted_len/slot_ids`; bind cache state to a KV `service_group` with `sequence_axis`, `layout`, semantic `logical_lengths: int64[B]`, storage, and past/present aliases; request sources no longer duplicate the semantic role.

Full exact checklist and YAML are in `docs/WORKFLOW_POLICY_COMPONENTS.md`; the detailed coordination comment is on #828.

justinchuby:
Cross-repo execution against onnx-genai #828 imported `3b0445a` immutably. Decoder/masked/codec execute; four packages fail before workflow runtime:

- VLM `embedding/model.onnx`: `inputs_embeds` is declared float but produced from int64 (`emit_inputs_embeds`).
- Diffusion `text_encoder/model.onnx`: `encoder_hidden_states` is declared float but produced from int64.
- Speculative `proposer/model.onnx`: graph outputs `proposed_tokens`/scores have no producers.
- TTS `code_predictor/model.onnx`: missing initializer producer for `code_predictor.layers.0.input_layernorm.weight`.

All seven metadata files pass the callable validator, demonstrating why runtime E2E is now mandatory. ONNX CI/test wiring is ready and will pin the corrected producer commit once these graph defects are regenerated.

justinchuby:
Confirmed corrected producer head `92bc47ef`: all seven checked-in packages now execute under #828 (7/7), including VLM preprocessing, Euler loop, nested TTS, and speculative grammar forced-token rejection/correction. One remaining producer diagnostic: the TTS E2E emits ORT memory-pattern shape-reuse warnings as sequence shapes grow (`{1,1,6} != {1,1,2..4}`, `{1,6} != {1,2..4}`, later `{1,1,7}`). Execution is correct and ORT falls back, but shared `dim_param` names appear to claim equality across changing sequence dimensions. Please tighten those graph shape symbols in a follow-up to preserve allocator/capture quality.

justinchuby:
Final validation at `9c0cc54b`: ONNX GenAI metadata/schema validation passed; all 7/7 checked workflow packages execute (decoder, VLM, diffusion, masked diffusion, codec, TTS, speculative); deterministic fixture regeneration passed; Ruff passed; Mobius suite passed (3696 passed, 52 skipped). Muse metadata: 764 lines, 720 fields, 20,545 bytes, SHA-256 `5f4019a492f2278728743e9e2658be00455c0952076098b3cbfae5d0514c0191`. The Integration (fast) rerun remains queued without a runner; its prior failures were isolated DeepSeek numerical-parity tests unrelated to this metadata/serialization diff. No merge or HF upload performed.

justinchuby:
CI follow-up: the Integration (fast) rerun completed with the identical two pre-existing DeepSeek numerical-parity failures (`test_deepseek_v2_lite_prefill_logits_match`, `test_deepseek_non_mla_decoder_prefill_logits_match`). All other checks passed again, including ONNX GenAI metadata and L1/L3/L4/L5. The failures do not exercise package save, workflow metadata, TTS, or regenerated fixtures.

justinchuby:
Real Muse H200 follow-up is pushed at `c332bec` (producer dtype fix `d7927c3`).

Paired native result on ORT 1.28/H200, exact 68-token prompt, 128 new tokens, 1 warmup + 3 runs, CUDA Graph/shared KV: **63.3653 tok/s median, 49.0217 ms TTFT** (2.60% above the historical 61.76 tok/s baseline).

The real metadata workflow currently cannot complete in frozen ONNX GenAI `a341c463`; exact runtime blockers found while executing the 52-layer package:
1. text-only VLM cannot omit `request.image` (optional tensor defaults unsupported);
2. image adapter rejects derived `total_patches` before preprocessing;
3. CUDA loop carries/continue masks cannot be cloned/inspected without host materialization;
4. cuDNN fused GQA fails for the real prefill shape, requiring unfused fallback.

Mobius also fixed BF16 decoder logits → float32 sampler ABI and added the exact runnable native/workflow harness. I returned the PR to draft and have not replaced the HF metadata, because the currently frozen runtime cannot execute the published workflow without runtime changes.

justinchuby:
Publication safety update (`48a433b`): added a regression that constructs a 52-layer BF16 decoder artifact while deliberately passing the tiny 4-layer config, then requires metadata to derive all 106 inputs/105 outputs, 104 KV aliases, head_dim 128, vocab 202048, and absence of `position_ids` from the artifact.

A metadata-only atomic HF bundle is staged locally with corrected metadata, all seven referenced policy ONNX files, and README frozen-schema note. Static `validate_metadata` and real artifact admission pass. Its manifest keeps `upload_allowed=false`; upload remains withheld until real multi-token runtime E2E and paired H200 performance pass. The published `d7a34f` metadata remains explicitly invalid and must be replaced, not documented as a benchmark caveat.

justinchuby:
KV ABI correction pushed as `b6c7cba`: conventional past/present decoder artifacts now emit `paging: none`, `storage: shared_buffer`, `compaction: false`; paged storage is reserved for artifacts that actually expose page/block-table inputs. Muse corrected metadata now hashes `ad8e99db32c4d25c2f9218f781a6e53a7bde482e28d6f6b12ba568878340dcd5` with all 104 aliases. Frozen validator and full Mobius suite pass (3697 passed, 52 skipped). Upload remains withheld pending ORT 1.28 fixed-capacity alias/capture verification, real multi-token E2E, and paired performance.

justinchuby:
Producer update: fixed-capacity shared-buffer state is pushed in `b282b2c`, text-only Muse VLM branching in `790b2c9`, and refreshed deterministic fixtures in `94e3761`. Frozen `a341c463` validation passes for the real 52-layer package; full local suite: 3700 passed, 52 skipped; Ruff passes.

Real H200 ORT 1.28 workflow admission now loads every artifact and reaches loop execution. Remaining runtime blocker is exact: `workflow loop carry active cannot preserve inactive rows` / `cannot borrow bytes of a device-resident tensor; copy it to host first`. HF publication remains withheld (`upload_allowed=false`) pending runtime E2E and paired performance. Current unpublished metadata SHA-256: `24084e146c73e8ed9bff064d5adcaf94c26f40594f186bcb9142b2af0d5e6a66` (3117 lines, 2640 fields, 99584 bytes).

justinchuby:
Pushed `76b74ce` to fix the remaining decoder/VLM conformance regression: KV workflow state is bounded (allowing synthetic CPU present-cache growth) while fixed-capacity attention masks and `shared_buffer` serving aliases remain intact.

Validation:
- frozen `validate_metadata`: 7/7 packages
- ONNX GenAI workflow runtime, ORT 1.28 CPU: 7/7
- targeted metadata tests: 51 passed
- full Mobius non-integration suite: 3700 passed, 52 skipped
- PR `ONNX GenAI metadata`, Lint, and lintrunner checks: passed

Real Muse metadata was regenerated from the admitted 52-layer package: SHA-256 `96b5ebec38caaee97c8b26374cad83473f29409da5428fde2b6a6ffc6747839a`, 3325 lines, 2848 fields, 104888 bytes. HF upload remains withheld pending runtime device-resident inactive-row carry support, real multi-token E2E, and paired H200 measurement.

justinchuby:
Pushed `6097887` for true text-only VLM parity with the latest workflow presence contract:
- `request.image` is optional with `present_as: request.image_present`
- the true branch alone runs preprocess + vision
- the false branch emits empty image features
- removed the sentinel `request.has_media` input

Validation: presence-aware schema admission passes; 51 metadata tests pass; H200/ORT 1.28 real Muse completed the canonical 68+128-token workflow with the image omitted and produced the same 128 token IDs as the prior sentinel run. Regenerated unpublished metadata: `f36c4ddb5d990213a77a26a0dd4dbdce1a7550e4d4ce5e6f122e6e3ab9ecac3f`, 3314 lines, 2837 fields, 104690 bytes. HF upload remains withheld pending the runtime carry patch and paired release benchmark.

justinchuby:
Follow-up `ab3a1a6` prevents real-package termination/capacity drift: VLM workflows now derive `eos_token_id` and `context_length` from packaged `genai_config.json` before config fallbacks. Regression supplies tiny/wrong EOS `2` and context `128`, then verifies admitted package values `200001` and `131072` win. The corrected Muse bundle remains SHA-256 `f36c4ddb5d990213a77a26a0dd4dbdce1a7550e4d4ce5e6f122e6e3ab9ecac3f`; schema validation and 51 metadata tests pass.

justinchuby:
The generic adapter ABI is now published in ONNX GenAI commit `21a935c241f9fb8bb4b77e4df39b2716b0f70a26` (PR #828). Normative contract: https://github.com/justinchuby/onnx-genai/blob/21a935c241f9fb8bb4b77e4df39b2716b0f70a26/docs/WORKFLOW_POLICY_COMPONENTS.md#parameter-adapters-lora

Please persist/freeze producer artifacts and cross-repo fixtures against this exact commit:

```yaml
adapters:
  base_model_fingerprint: onnx-genai-targeted-base-v1:sha256:<64 lowercase hex>
  selection:
    row_ids: request.row_ids
    request_epochs: request.request_epochs
    adapter_ids: request.adapter_ids
    adapter_counts: request.adapter_counts
    scales: request.adapter_scales
    active: request.active
    max_adapters: 4
  application_capability: onnx-genai.adapters@1
```

Canonical files are under `adapters/<alias>/`, with lowercase SHA-256 of exact bytes. Formats: RFC 8785 float32 JSON (`targets.<weight_key>.{a,b}`), safetensors (`<weight_key>.a/.b`), or upstream ORT `.onnx_adapter` (`TORT`, v1). Artifact indices are contiguous from zero. The targeted-base fingerprint hashes RFC 8785 canonical JSON over sorted `(component,parameter)` records containing exact initializer dtype/shape/logical little-endian bytes and sorted direct-consumer identity/attributes.

Selection is immutable SSA: IDs `[B,K]`, counts `[B]`, scales `[B,K]`, row IDs/epochs `[B]`, optional active `[B]`; unused slots are exactly `-1/0`. Composition is ordered `base + Σ scale*(alpha/rank)*B*(A*x)`. Compaction carries all selection/state together by `(row_id,request_epoch)`; epoch changes on slot reuse.

Please generate exact executable fixtures for heterogeneous B>1 rows, no-adapter row, composed adapters, compaction/reorder, slot reuse, eviction/reload, capture replay, plus invalid base fingerprint/shape/checksum/selection cases, and report artifact hashes for byte-identity consumption.

justinchuby:
**Supersedes the earlier 21a935c adapter draft. Do not freeze fixtures against that commit.**

The canonical ABI integrating native LoRA PR #318 (`813a9b53`) and #374 (`326fddcf`) is ONNX GenAI commit `903a2d1aad6e58ecd966db70e8b2fd310fce146a` on PR #828. Normative contract and exact reuse/adapt/retire matrix: https://github.com/justinchuby/onnx-genai/blob/903a2d1aad6e58ecd966db70e8b2fd310fce146a/docs/WORKFLOW_POLICY_COMPONENTS.md#parameter-adapters-lora

Key producer changes before fixture freeze:
- emit one **top-level** `InferenceMetadata.adapters`; there is no `workflow.adapters`;
- emit authoritative `target_manifest.targets[]` with stable `id`, generic `component`, exact base `parameter`, optional `output_value`/resolved `output_slice`, `activation_dtype`, dimensions, and optional Phase-1 `graph_inputs.{a,b,scale}`;
- declare source format/capability: `hf_peft` pairs checksummed `adapter_config.json` + safetensors; `ort_genai` is checksummed upstream `TORT` v1; Mobius is not required to emit FlatBuffers;
- artifact `bindings[]` reference manifest target IDs and weight keys, with optional per-target rank/alpha;
- request SSA is `slot_ids[B]`, `request_epochs[B]`, `segments[B,K]`, `adapter_counts[B]`, `scales[B,K]`, optional `active[B]`; unused slots are exactly `-1/0`;
- K=1 preserves Phase-2 segment routing; K>1 is ordered additive composition. Compaction carries slot ID, epoch, segments, counts, scales, active flag, and model state together.

Please update persistence and generate exact fixtures against `903a2d1a` for PEFT and/or ORT sources, heterogeneous B>1, base-only, ordered composition, compaction/reorder, slot reuse, eviction/reload, capture replay, and invalid fingerprint/shape/checksum/ID cases. Report producer commit plus exact file hashes before ONNX GenAI consumes/finalizes them.

justinchuby:
Exact producer ABI is now finalized at ONNX GenAI `d9482bca0ad6ccec907a2b49faf033c2006a9fa9` (supersedes `903a2d1a` for fixture pinning).

```yaml
adapters:
  base_model_fingerprint: onnx-genai-targeted-base-v1:sha256:<hex>
  target_manifest:
    targets:
      - id: <stable generic target id>
        component: <pipeline component or model>
        parameter: <exact immutable initializer>
        node_name: <exact ONNX projection node>
        output_name: <exact ONNX output value>
        activation_dtype: float16
        input_features: <K>
        output_features: <N>
        output_slice: { role: <producer label>, offset: <u>, width: <u> } # optional
        graph_inputs: { a: <optional A input>, b: <optional B input>, scale: <optional scale input> } # optional Phase-1 seam
  discovery_fallback: disabled # tooling_only only while importer resolves the manifest
  selection:
    slot_ids: <int64[B] SSA>
    request_epochs: <int64[B] SSA>
    segments: <int64[B,K] SSA>
    adapter_counts: <int64[B] SSA>
    scales: <float32[B,K] SSA>
    active: <bool[B] SSA> # optional
    max_adapters: <K>
  application_capability: onnx-genai.adapters@1
  portable_fallback: true
  artifacts:
    <alias>:
      index: <contiguous segment id>
      identity: <stable identity>
      version: <version>
      base_model_fingerprint: <same fingerprint>
      rank: <default rank>
      alpha: <default alpha>
      dtype: <factor dtype>
      provenance: { producer: mobius, source: <optional URI>, revision: <optional immutable revision> }
      weights:
        - format: hf_peft
          loader_capability: onnx-genai.adapters.hf-peft@1
          location: adapters/<alias>/adapter_model.safetensors
          sha256: <exact bytes>
          config_location: adapters/<alias>/adapter_config.json
          config_sha256: <exact bytes>
          scale_encoding: alpha_over_rank
        # ORT alternative: format=ort_genai, loader_capability=onnxruntime.lora-adapter@1,
        # location=...onnx_adapter, sha256=..., scale_encoding=baked
      bindings:
        - target: <manifest target id>
          weight_key: <canonical A/B key>
          rank: <optional override>
          alpha: <optional override>
```

Important #374 detail now explicit: TORT scale is baked, so ORT sources require `scale_encoding: baked` and must not receive alpha/rank again. PEFT is primary/provenance-preserving with `alpha_over_rank`. Exact node/output names and labeled fused slices are retained for graph validation; execution consumes the resolved bindings, never model-family discovery.

Normative docs/matrix: https://github.com/justinchuby/onnx-genai/blob/d9482bca0ad6ccec907a2b49faf033c2006a9fa9/docs/WORKFLOW_POLICY_COMPONENTS.md#parameter-adapters-lora

justinchuby:
The final ONNX GenAI adapter ABI superseding the earlier `d9482bca` pin is now published at `justinchuby/onnx-genai@793bfe9b5489257a8ae4126ba66b760e3b2cfe19` (PR https://github.com/justinchuby/onnx-genai/pull/828).

Please serialize/freeze producer fixtures against these exact public target names:
- `component`
- `initializer`
- optional `layer_index`
- optional target `rank` / `alpha`
- optional `output_slice { role, offset, width, rank, alpha }`
- optional Phase-1 `graph_inputs { a, b, scale }`

Effective artifact binding rank/alpha is checked against both target and output-slice policy. `component + initializer` is also the canonical targeted-fingerprint identity. This matches the authoritative exact-graph manifest direction in producer `7bb3fac`; model-family discovery remains producer/import tooling only.

justinchuby:
Final adapter consumer pin correction: use Mobius `62c69f51cbc9495c1c91a69221424e709880f5ad` against ONNX GenAI `2af34dcad0e429604ca66d7ba1388ee2e688756e`; do not use `d9bf9ffc` or `9a3f523`.

The final committed YAML is [here](https://github.com/onnxruntime/mobius/blob/62c69f51cbc9495c1c91a69221424e709880f5ad/tests/fixtures/onnx_genai_workflows/adapter/inference_metadata.yaml) and uses top-level `adapters`, `target_manifest`, SSA `selection`, artifact `bindings`, structured provenance, loader capabilities, and scale encoding. SHA-256 is `c63c9a3ac59f87383044fe3e47fa7b7b615c823763a0fda8b07e764e4d729e0e` (stale d9bf YAML was `8701473391a93707ac86d78a00ba7fd4387109d2576c3cabb09d88ee32e7bec4`).

The final [conformance test](https://github.com/onnxruntime/mobius/blob/62c69f51cbc9495c1c91a69221424e709880f5ad/tests/onnx_genai_workflow_conformance.rs) supplies `slot_ids`, `request_epochs`, `adapter_segments`, `adapter_counts`, `adapter_scales`, and `active` via `.with_input`; SHA-256 `15ae3864e7003e0989cf76a060fdcff0472897ceafe633a92e808d9afde3b17d` (stale d9bf test was `8297818c0166ce338c6a70a2efa865bf54787370586c84d7dd164220b7096446`). `AdapterSelection` exists only as a test-side lowering helper; it is never passed out-of-band to the runtime.

.github/workflows/main.yml
```diff
@@ -15,6 +15,68 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  onnx-genai-metadata:
```
justinchuby, line 18: Remove or made separate

PR #487, opened 2026-08-14
Add NVIDIA Nemotron 3.5 Lightning export support

justinchuby:
## Summary

- add pinned support for `nvidia/NVIDIA-Nemotron-3.5-Lightning-30B-A3B-BF16@d468880b6ad3c6e0d21377ce7242adaea4cc884d` using the real `nemotron_h` architecture, not an alias
- normalize legacy/current layer vocabularies; preserve FP32 router correction bias, routed-expert accumulation, and Mamba recurrent state; exclude only auxiliary `mtp.*` tensors
- add pinned reduced-real L4/L5 coverage across Mamba, MoE, attention, convolution/SSM/KV state, FP16 CUDA, and fresh Olive Q4 final-package reload
- emit graph-derived ORT GenAI metadata without model/runtime/topology capability gates; ORT GenAI owns downstream acceptance
- provide Option A: pinned BF16 checkpoint -> FP16 portable ONNX -> Olive Q4 -> direct cached ONNX Runtime generation
- inherit #496's quantized-by-default GGUF import and `--dequantize` opt-out unchanged
- inherit the seven generic runtime/export skill updates from merged #500 (`8143a7cc06a26bb32a6f4e99f0000b06e5ba3ae0`) exactly once; this PR retains only 19 lines of Nemotron-specific `moe-models` guidance
- retain merged stacked PR #498 exactly once as the final Qwen3.8 commit; #498 was merged into #487's branch, not `main`

## Exact-head evidence

- base: `cf77d0f7fb28adebbeb3aaecb25fe2e51ee0ddd0`
- head: `8ccfc25977d9551c7dfc07088eea1827a5fcb8b8`
- commit graph: 11 Nemotron commits + one merged Qwen3.8 commit + one rebase import-fix commit
- focused GenAI/GGUF/NeMo/Parakeet/Nemotron/Qwen regressions: **599 passed, 6 deselected**; the only additional selector was reproduced identically on detached exact base
- Qwen3.8 image/video/mixed HF-vs-ONNX integration: **1 passed**
- canonical unfiltered: **4,116 passed, 58 skipped, 1 subtest passed, 9 failures**
- canonical excluding only the nine selectors reproduced on detached exact base: **4,116 passed, 58 skipped, 1 subtest passed**
- reduced-real CPU L4 + L5, FP16 CUDA, and fresh Olive Q4 final package: **4 passed**
- initialized `lintrunner`; `lintrunner -a -r origin/main` and `git diff --check`: clean
- independent exact-head review: no significant findings
- seven merged #500 skill paths: zero diff from `main`; `moe-models`: 19 Nemotron-specific added lines

The nine exact-base failures are: GGUF Q4_0 zero-point dequantization; two Attention softcap decomposition selectors; Qwen Image FP16 and BF16 CUDA low-precision selectors; and four newly merged upstream-patch asset tests. All reproduce on detached `cf77d0f`; this PR does not hide or modify them.

## CUDA and Olive

- FP16 CUDA prefill max abs: `0.0078125`, cosine `0.999997079`
- cached-step max abs: `0.00292969`, `0.0078125`, `0.00390625`, `0.00390625`
- cached-step cosine: `0.999999642`, `0.999999821`, `0.999999702`, `0.999999702`
- exact HF/ONNX generated IDs: `[12, 13, 12, 12]`
- weighted graph uses standard `Attention`, `Scan`, and `Conv`, with no `GroupQueryAttention`, `LinearAttention`, or `CausalConvWithState`
- CUDA profile: 960 CUDA / 342 CPU events
- fresh Olive 0.13 Q4: 17 `MatMulNBits`, smaller package, CUDA reload, finite logits, exact generation
- BF16 remains rejected after corrected reduced-real evidence reached about `0.855469` max abs; supported Option A converts source BF16 weights to FP16

The old A10 `max_abs=0.309412` cached-step failure was caused by provider-dependent fused hybrid cache kernels, not Olive: it occurred identically before and after quantization. A fully standard ONNX cache graph restores every reused-state step below `1e-2`; no tolerance or golden was changed.

## ORT GenAI metadata policy

Mobius emits faithful graph-derived filenames, semantic inputs/outputs, representable KV/cross/conv templates, actual `logits`/`logits_Q4` names, and the global cache-slot count. It does not reject export based on the current runtime registry, version, topology support, or cache executor capability. NemotronH `ssm_state` has no current config template, so that unrepresentable field is omitted without blocking artifacts; downstream load/generation remains optional evidence.

GGUF config extraction also validates special-token IDs against vocabulary bounds, drops serialized null/out-of-range sentinels, and emits a deduplicated EOS/EOT/EOM stop-token union. Current-main `generation_config.json` token-ID precedence and upstream asset corrections remain intact.

## Resource/runtime waivers

- no full-30B parity or full Olive claim: the source checkpoint is 65.8 GB and the validation host has an 8 GB RTX A1000; the reduced fixture retains production dimensions and representative complete Mamba/MoE/attention layers
- no BF16 execution claim: corrected real-weight evidence fails the numerical gate; use FP16
- ORT GenAI/Foundry load-generation is optional downstream evidence and never gates Mobius export; the direct ONNX Runtime path is validated end to end

Relates to #483.

Cross-cutting skill guidance merged in #500.

### DIFF
--- a/.agents/skills/ort-genai-config/SKILL.md
+++ b/.agents/skills/ort-genai-config/SKILL.md
@@ -291,15 +291,6 @@ global cache-slot count. Preserve intrinsic schema/config validation, but do
 not gate or reject export based on the current GenAI model registry, runtime
 version, topology support, or cache executor capability.
 
-NemotronH is unsupported in ORT GenAI 0.15.2 because sparse global layer
-indices mix attention key/value caches with Mamba `conv_state` and
-`ssm_state`. Do not hard-code a model-type or runtime-version rejection.
-Inspect the actual decoder graph and reject only when it requires cache inputs
-the generated config has no template for (currently `ssm_state`). This keeps
-the guard structural: remove it when config emission can represent that graph
-contract. Record the tested-version waiver and provide a direct ONNX Runtime
-loop for current users.
-
 ### "input_ids size exceeds max length"
 
 For image prompts, the tokenized input_ids (including image_pad tokens) can
--- a/.agents/skills/ort-genai-config/SKILL.md
+++ b/.agents/skills/ort-genai-config/SKILL.md
@@ -293,10 +293,12 @@ version, topology support, or cache executor capability.
 
 NemotronH is unsupported in ORT GenAI 0.15.2 because sparse global layer
 indices mix attention key/value caches with Mamba `conv_state` and
-`ssm_state`. Do not hard-code a model-type rejection based only on that runtime
-version: emit a structurally honest config so downstream releases can evolve,
-record the tested-version waiver, and provide a direct ONNX Runtime loop for
-current users.
+`ssm_state`. Do not hard-code a model-type or runtime-version rejection.
+Inspect the actual decoder graph and reject only when it requires cache inputs
+the generated config has no template for (currently `ssm_state`). This keeps
+the guard structural: remove it when config emission can represent that graph
+contract. Record the tested-version waiver and provide a direct ONNX Runtime
+loop for current users.
 
 ### "input_ids size exceeds max length"
 
--- a/.agents/skills/ort-genai-config/SKILL.md
+++ b/.agents/skills/ort-genai-config/SKILL.md
@@ -291,11 +291,12 @@ global cache-slot count. Preserve intrinsic schema/config validation, but do
 not gate or reject export based on the current GenAI model registry, runtime
 version, topology support, or cache executor capability.
 
-NemotronH is one such unsupported contract in ORT GenAI 0.15.2: sparse global
-layer indices mix attention key/value caches with Mamba `conv_state` and
-`ssm_state`. The generic decoder schema cannot bind all three cache families.
-Reject `--runtime ort-genai` before creating output and direct users to a
-token-by-token ONNX Runtime loop until the runtime adds a dedicated model type.
+NemotronH is unsupported in ORT GenAI 0.15.2 because sparse global layer
+indices mix attention key/value caches with Mamba `conv_state` and
+`ssm_state`. Do not hard-code a model-type rejection based only on that runtime
+version: emit a structurally honest config so downstream releases can evolve,
+record the tested-version waiver, and provide a direct ONNX Runtime loop for
+current users.
 
 ### "input_ids size exceeds max length"
 
--- a/.agents/skills/ort-genai-config/SKILL.md
+++ b/.agents/skills/ort-genai-config/SKILL.md
@@ -291,6 +291,12 @@ global cache-slot count. Preserve intrinsic schema/config validation, but do
 not gate or reject export based on the current GenAI model registry, runtime
 version, topology support, or cache executor capability.
 
+NemotronH is one such unsupported contract in ORT GenAI 0.15.2: sparse global
+layer indices mix attention key/value caches with Mamba `conv_state` and
+`ssm_state`. The generic decoder schema cannot bind all three cache families.
+Reject `--runtime ort-genai` before creating output and direct users to a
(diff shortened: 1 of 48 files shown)


.agents/skills/ort-genai-config/SKILL.md
```diff
@@ -289,6 +289,12 @@ the exported contract. If the runtime cannot route required feature inputs,
 position-ID rank, cache state, scheduler, or multimodal metadata, reject export
 before writing artifacts and report the exact runtime version/limitation.
 
+NemotronH is one such unsupported contract in ORT GenAI 0.15.2: sparse global
+layer indices mix attention key/value caches with Mamba `conv_state` and
+`ssm_state`. The generic decoder schema cannot bind all three cache families.
+Reject `--runtime ort-genai` before creating output and direct users to a
```
justinchuby: Don't reject. We should allow downstream to evolve without us having to change anything

.github/workflows/gpu_l4_golden_parity.yml
```diff
@@ -86,7 +86,7 @@ jobs:
           AFFECTED='${{ inputs.affected_models }}'
           if [ -z "$AFFECTED" ] || [ "$AFFECTED" = "[]" ]; then
             echo "Running all L4 golden comparison tests"
-            pytest tests/e2e_golden_test.py \
+            pytest tests/e2e_golden_test.py tests/nemotron_h_real_weight_test.py \
```
justinchuby: Why is model specific change?

src/mobius/__main__.py
```diff
@@ -360,6 +369,20 @@ def _resolve_static_cache_task(model_type: str) -> ModelTask:
     _save_package(pkg, output_dir, args, optimize, component_filter)
 
 
+def _preflight_ort_genai_runtime(args, model_type: str) -> None:
```
justinchuby: No

.agents/skills/ort-genai-config/SKILL.md
justinchuby: Removed the NemotronH/runtime-version-specific rejection. The remaining guard inspects the actual decoder inputs and only rejects a graph when config emission has no template for its `ssm_state`; the skill now says to remove that structural guard when the schema can represent the contract.

src/mobius/__main__.py
justinchuby: Removed the CLI preflight and model-name rejection entirely. Compatibility is now evaluated from the built graph contract before package artifacts are saved, so downstream support is not blocked by a hard-coded model or runtime version.

.github/workflows/gpu_l4_golden_parity.yml
justinchuby: Replaced the model-specific workflow path with the generic `tests/*_real_weight_test.py` extension point in both L4 and L5. Affected-model detection now infers model types from that naming convention and maps model-specific example assets explicitly.

tests/arch_validation_test.py
```diff
@@ -26,20 +26,26 @@
...
     _try_load_config_json,
 )
 from mobius._registry import registry
+from mobius._testing.golden import discover_test_cases
 from mobius.tasks import get_task
 
 logger = logging.getLogger(__name__)
 
+_PINNED_REVISIONS = {case.model_id: case.revision for case in discover_test_cases()}
+_DECLARED_DTYPES = {case.model_id: case.dtype for case in discover_test_cases()}
```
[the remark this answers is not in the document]
justinchuby: Cached `discover_test_cases()` once in `_TEST_CASES` and now derive both revision and dtype mappings from that single parse.

PR #491, opened 2026-08-14
Fix Olive quantized-weight suffix mismatch in preprocess_olive_weights

titaiwangms:
## Problem

`preprocess_olive_weights()` in `_weight_utils.py` assumed Olive's quantized checkpoints use a **dotted** suffix convention (`<pname>.qweight` / `.scales` / `.qzeros`). Olive actually uses an **underscore** suffix appended directly to the owning parameter name — e.g. `model.layers.0.mlp.gate_proj.weight_qweight`, or for fused MoE parameters with no nested `Linear`, `model.layers.0.mlp.experts.gate_up_proj_qweight` (see `olive/common/quant/state_dict.py`, `QWEIGHT_SUFFIX = "_qweight"` etc.).

Because of this mismatch, every `endswith`/`replace` check in the function silently failed to match, so real Olive-quantized checkpoints fell through to the "pass through unchanged" branch instead of being renamed/reshaped for `MatMulNBits` / `GatherBlockQuantized` / `QMoE` consumption. This surfaced as `"N initializer(s) without weights"` errors when trying to load an Olive `Rtn`/`Gptq`-quantized checkpoint into a Mobius-built graph.

This exact mismatch was flagged narrowly in #427 (for a defensive `ndim==2` guard on the older `moe=True` 3-D path) and left unresolved ("parked"). #451 added QMoE emission for Qwen3.5/3.6-MoE on top of the same incorrect dotted-suffix assumption, so it also never actually exercised a real Olive checkpoint end-to-end.

## Repro

Chaining Olive's `Rtn(moe=true)` -> `MobiusBuilder` against `yujiepan/qwen3.5-moe-tiny-random` (or `optimum-intel-internal-testing/tiny-random-qwen3.5-moe`) reproduces the failure directly.

## Fix

- Added a `_rename()` helper that strips the underscore suffix, drops a redundant trailing `.weight` component (present for regular `nn.Linear`/`nn.Embedding` params, absent for bare fused-MoE params like `gate_up_proj`), and appends the correct dotted target name.
- Kept the `embed_tokens` special case separate (its packed weight keeps the `qweight` name, not `weight`, since `QuantizedEmbedding`/`GatherBlockQuantized` consume it directly).
- Added a previously-missing `_scales` rename branch for the general (non-embedding) case.
- Updated `_weight_utils_test.py`, `models/qwen35_test.py`, and `components/_moe_test.py` fixtures, which had been written against the same incorrect dotted-suffix convention.

## Scope / what this does NOT fix

This fixes weight **naming**, not quantization **coverage**. `Qwen35Attention`, `GatedDeltaNet`, and the `MLP(config)` call site in `Qwen35DecoderLayer` still hardcode plain `Linear` regardless of `config.quantization` — only the MoE fused-expert path (added in #451) threads quantization through for this model family. So a quantized Qwen3.5/3.6 checkpoint will still fail with a shape mismatch on `self_attn`/`linear_attn`/dense `mlp` weights after this fix (confirmed via the same repro, isolated with `modules_to_not_convert`). That's a separate, larger gap — will file a follow-up issue for it.

## Testing

- `pytest src/mobius/_weight_utils_test.py src/mobius/models/qwen35_test.py src/mobius/models/deepseek_test.py src/mobius/components/_moe_test.py` — 117 passed.
- Repro confirmed to now get past the suffix/rename stage and fail only at the (separately tracked) `linear_class` gap.

### DIFF
--- a/src/mobius/_weight_utils.py
+++ b/src/mobius/_weight_utils.py
@@ -716,19 +716,29 @@ def preprocess_olive_weights(
 ) -> dict[str, torch.Tensor]:
     """Rename and reshape Olive-packed quantized weights.
 
-    Olive stores quantized weights with uint8 packing:
-      - ``*.qweight``: [N, packed_K] uint8
-      - ``*.scales``: [N, n_blocks]
-      - ``*.qzeros``: [N, ceil(n_blocks * bits / 8)] uint8, asymmetric only
+    Olive stores the quantized state of a parameter named ``<pname>`` (e.g.
+    ``"weight"`` for ``nn.Linear``/``nn.Embedding``, or ``"gate_up_proj"`` for
+    a fused-3D MoE expert parameter) as sibling buffers on the owning module,
+    using an *underscore* suffix convention (see Olive's
+    ``olive/common/quant/state_dict.py``) — **not** a dotted one:
+      - ``<pname>_qweight``: [N, packed_K] uint8 (always present)
+      - ``<pname>_scales``: [N, n_blocks] (always present)
+      - ``<pname>_qzeros``: [N, ceil(n_blocks * bits / 8)] uint8 (asymmetric only)
+
+    e.g. ``model.layers.0.mlp.gate_proj.weight_qweight`` (regular ``nn.Linear``)
+    or ``model.layers.0.mlp.experts.gate_up_proj_qweight`` (fused MoE param,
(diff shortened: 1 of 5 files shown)


titaiwangms:
Consolidated into #495 along with #493 and #494 to simplify review.

PR #493, opened 2026-08-14
Thread linear_class through Qwen3.5 self_attn/linear_attn/shared_expert

### LINKED ISSUE #492: Qwen3.5/3.6: quantization not wired for self_attn/linear_attn/dense MLP (only MoE experts are)
## Summary

For the Qwen3.5/3.6 hybrid DeltaNet + full-attention MoE architecture, **only the MoE fused-expert path** (added in #451) actually threads quantization (`linear_class`/`QuantizedLinear`) through into the built ONNX graph. Every other Linear-like weight in the model is hardcoded to plain float `Linear`, regardless of `config.quantization`:

- `Qwen35Attention` (`components/_attention.py`, full-attention layers): `q_proj`/`k_proj`/`v_proj`/`o_proj` are always plain `Linear`. Constructor has no `linear_class` parameter at all (unlike the generic `Attention` class in the same file, which does accept one).
- `GatedDeltaNet` (`components/_gated_deltanet.py`, the majority of layers under `full_attention_interval`): `in_proj_qkv`/`in_proj_z`/`in_proj_b`/`in_proj_a`/`out_proj` are always plain `Linear`. No `linear_class` parameter at all.
- `Qwen35DecoderLayer.__init__` (`models/qwen35.py`): constructs `MLP(config)` without passing a `linear_class`, even though `MLP` (`components/_mlp.py`) already supports one. This affects both the dense `mlp` and the MoE block's `shared_expert`.

## Why this matters

Mobius does not trace/export the PyTorch model — it rebuilds a fresh ONNX graph node-by-node in Python. The choice to emit `MatMul` (plain, dense float initializer) vs `MatMulNBits`/`GatherBlockQuantized` (quantized, packed uint8 qweight+scales+zero_points) per weight is a **hardcoded, compile-time decision** at graph-construction time in the model-definition code, not something inferred from the checkpoint. So even a fully-quantized Olive checkpoint (see #491 for the naming-convention fix that makes Olive keys resolve correctly) fails to load into these modules with a shape mismatch, e.g.:

```
ValueError: Weight shape mismatch for 'model.layers.3.self_attn.k_proj.weight': model expects [16, 32], got [16, 2, 8]
```

because the graph's initializer slot for `k_proj.weight` was built assuming a dense float `[out, in]` shape, and can never accept packed quantized data.

## Repro

Olive `Rtn(bits=4, group_size=16, sym=true, moe=true, modules_to_not_convert=["linear_attn","shared_expert"])` -> `MobiusBuilder` against `optimum-intel-internal-testing/tiny-random-qwen3.5-moe` (after applying #491's suffix fix), excluding `linear_attn`/`shared_expert` to isolate: fails at `self_attn.k_proj` with the shape mismatch above. Without the exclusion, fails first at `linear_attn.in_proj_a`/similar DeltaNet projections for the same reason.

## Suggested fix

Thread `linear_class` (matching the pattern `models/base.py`'s generic `DecoderLayer`/`Attention`/`MLP` already use) through:
1. `Qwen35Attention.__init__` -> `q_proj`/`k_proj`/`v_proj`/`o_proj`
2. `GatedDeltaNet.__init__` -> `in_proj_qkv`/`in_proj_z`/`in_proj_b`/`in_proj_a`/`out_proj`
3. `Qwen35DecoderLayer.__init__` -> pass `linear_class` through to its `MLP(config)` construction (both dense MLP and MoE `shared_expert`)

Open design question for (2): `GatedDeltaNet` also owns `dt_bias`, `A_log`, and `conv1d.weight` — small recurrent-state-related tensors that are likely inappropriate to quantize even once `linear_class` threading exists. Need a decision on whether these should always stay unquantized regardless of `config.quantization`.

## Related

- Prerequisite/companion fix: #491 (fixes the Olive checkpoint key-naming mismatch; without it, this issue's repro fails even earlier at the naming stage instead of surfacing the shape mismatch).
- Related but distinct from #427's originally-scoped concern (which was about the 2-D vs 3-D RTN `moe=True` guard, not this general `linear_class` gap).

titaiwangms:
## Stacked on #491

This PR is stacked on #491 (the Olive suffix fix) — its base branch is `fix/olive-suffix-mismatch`, not `main`. Please merge #491 first; this diff will shrink to just this PR's own commit once that lands and the base is retargeted to `main`.

## Problem (#492)

For Qwen3.5/3.6's hybrid DeltaNet + full-attention MoE architecture, only the fused MoE expert path had real quantization wiring. Three call sites hardcoded plain `Linear` regardless of `config.quantization`, unlike the generic `Attention`/`MLP` classes used elsewhere in Mobius:

- `Qwen35Attention` (full-attention layers): `q_proj`/`k_proj`/`v_proj`/`o_proj`
- `GatedDeltaNet` (linear-attention layers): `in_proj_qkv`/`in_proj_z`/`in_proj_b`/`in_proj_a`/`out_proj`
- `Qwen2MoELayer`'s `shared_expert`/`shared_expert_gate` (reused by `Qwen35MoEBlock`) — the always-active shared expert isn't part of the fused QMoE op, so it needs its own quantization wiring separate from the routed experts

Since Mobius rebuilds the ONNX graph from scratch (rather than tracing the PyTorch model), the initializer slots for these modules were always built expecting dense float weights, so a quantized checkpoint's packed weights could never load — surfacing as a shape mismatch (e.g. `model.layers.3.self_attn.k_proj.weight: model expects [16, 32], got [16, 2, 8]`).

## Fix

- Added an optional `linear_class` constructor parameter to `Qwen35Attention` and `GatedDeltaNet`, matching the pattern already used by the generic `Attention`/`MLP` classes. `GatedDeltaNet`'s `dt_bias`/`A_log`/`conv1d.weight` are small recurrent-state parameters and are intentionally left unquantized regardless of this argument (open design question from #492, resolved here: always float).
- Added an optional `linear_class` parameter to `Qwen2MoELayer` for `shared_expert`/`shared_expert_gate`. Defaults to `None` (plain `Linear`), so existing `Qwen2MoEDecoderLayer` (Qwen2-MoE) callers are unaffected.
- Added a `_linear_factory(config)` helper in `qwen35.py` (mirrors the existing pattern already used in `deepseek.py`) and threaded it through `Qwen35DecoderLayer`/`Qwen35MoEDecoderLayer`'s `self_attn`/`linear_attn`/`mlp` construction.

## Testing

- `pytest src/mobius/models/qwen35_test.py src/mobius/models/_qwen35_mtp_test.py src/mobius/components/_attention_test.py src/mobius/components/_gated_deltanet_test.py src/mobius/components/_moe_test.py src/mobius/components/_mlp_test.py src/mobius/_weight_utils_test.py src/mobius/models/deepseek_test.py` — 198 passed.
- **End-to-end**: Olive `Rtn(bits=4, group_size=16, sym=true, moe=true)` -> `MobiusBuilder` against `optimum-intel-internal-testing/tiny-random-qwen3.5-moe`, **without** excluding `linear_attn`/`shared_expert` via `modules_to_not_convert`, now succeeds end-to-end: `self_attn`/`linear_attn`/`shared_expert` weights save as quantized `u8` tensors, and a full ONNX model + `genai_config.json` are produced.

Closes #492.

### DIFF
--- a/src/mobius/components/_attention.py
+++ b/src/mobius/components/_attention.py
@@ -467,13 +467,23 @@ class Qwen35Attention(nn.Module):
     - Per-head Q/K RMSNorm with +1 offset (OffsetRMSNorm)
     - Partial RoPE (rotary_embedding_dim < head_dim)
     - Output gating: attn_output * sigmoid(gate)
+
+    Args:
+        config: Architecture configuration.
+        linear_class: Factory callable ``(in_features, out_features, bias=...)``
+            for creating projection layers. Defaults to ``Linear``. Pass a
+            quantized-linear factory (see ``make_quantized_linear_factory``)
+            to emit ``MatMulNBits`` for q/k/v/o projections instead.
     """
 
     def __init__(
         self,
         config: ArchitectureConfig,
+        linear_class: type | None = None,
     ):
         super().__init__()
+        if linear_class is None:
+            linear_class = Linear
         self.hidden_size = config.hidden_size
         self.head_dim = config.head_dim
         self.num_attention_heads = config.num_attention_heads
@@ -486,22 +496,22 @@ class Qwen35Attention(nn.Module):
         self._rope_interleave = config.rope_interleave
 
         q_dim = self.num_attention_heads * self.head_dim
-        self.q_proj = Linear(
+        self.q_proj = linear_class(
             self.hidden_size,
             q_dim * 2,
             bias=config.attn_qkv_bias,
         )
-        self.k_proj = Linear(
+        self.k_proj = linear_class(
             self.hidden_size,
             self.num_key_value_heads * self.head_dim,
             bias=config.attn_qkv_bias,
         )
-        self.v_proj = Linear(
+        self.v_proj = linear_class(
             self.hidden_size,
             self.num_key_value_heads * self.head_dim,
             bias=config.attn_qkv_bias,
         )
-        self.o_proj = Linear(
+        self.o_proj = linear_class(
             q_dim,
             self.hidden_size,
             bias=config.attn_o_bias,
--- a/src/mobius/components/_gated_deltanet.py
+++ b/src/mobius/components/_gated_deltanet.py
@@ -102,10 +102,23 @@ class GatedDeltaNet(nn.Module):
     The forward pass is a clean sequence of op calls.
 
     The recurrent state replaces the KV cache for these layers.
+
+    Args:
+        config: Architecture configuration.
+        linear_class: Factory callable ``(in_features, out_features, bias=...)``
+            for creating the ``in_proj_qkv``/``in_proj_z``/``in_proj_b``/
+            ``in_proj_a``/``out_proj`` projections. Defaults to ``Linear``.
+            Pass a quantized-linear factory (see
+            ``make_quantized_linear_factory``) to emit ``MatMulNBits`` for
+            these projections instead. ``dt_bias``, ``A_log``, and
+            ``conv1d.weight`` are small recurrent-state parameters and are
+            never quantized regardless of this argument.
     """
 
-    def __init__(self, config: ArchitectureConfig):
+    def __init__(self, config: ArchitectureConfig, linear_class: type | None = None):
         super().__init__()
+        if linear_class is None:
+            linear_class = Linear
         self.hidden_size = config.hidden_size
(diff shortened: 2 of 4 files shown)


titaiwangms:
Consolidated into #495 along with #491 and #494 to simplify review.

PR #505, opened 2026-08-17
Fix QMoE scale precision: scales silently forced entire MoE FFN onto CPU EP

titaiwangms:
## Bug

When exporting a hybrid MoE model (e.g. `Qwen/Qwen3.6-35B-A3B`) via the native QMoE-emission path in `MoELayer` (`_moe.py`) at a non-float32 precision (e.g. `fp16` for CUDA), the `fc1_scales`/`fc2_scales` parameters were pinned with `_keep_float32 = True`, which `_cast_module_dtype()` (`_builder.py`) honors by skipping them during the FLOAT->target-dtype cast pass. Every other parameter (weights, activations) was correctly cast to FLOAT16, but the scales stayed FLOAT32.

ONNX Runtime's `com.microsoft::QMoE` kernel (`quant_type="int"`) requires its `T2` type constraint (used by `fc1_scales`/`fc2_scales`) to be **exactly equal to `T`** (the activation dtype) — confirmed by reading `onnxruntime/contrib_ops/{cuda,cpu}/moe/moe_quantization*.{cc,h}`. There is no registered kernel variant for `T2=FLOAT32` while `T=FLOAT16`.

Because of this mismatch, ONNX Runtime silently fails to find *any* matching QMoE kernel — on **either** CPU or CUDA EP — and falls back to placing every QMoE node on `CPUExecutionProvider`, inserting `InsertedPrecisionFreeCast_*` bridging nodes. Since QMoE is the entire MoE FFN compute (the dominant cost for a large MoE model), this silently forced 100% of the MoE compute onto CPU even when the workflow explicitly targeted CUDA — and also broke `enable_cuda_graph` in ONNX Runtime GenAI, since not all decoder nodes were assigned to the same EP.

## Fix

- `_moe.py`: stop pinning `fc1_scales`/`fc2_scales` to FLOAT32; let them be downcast to the model's target export precision (FP16/BF16) like every other float parameter.
- `_qmoe_fusion.py` (the separate "dense-fallback -> QMoE" rewrite rule): had the same class of bug — scales and router logits were hardcoded to FLOAT/FLOAT32 regardless of the model's activation dtype. Fixed to use the activation dtype instead.
- Updated `_moe_test.py` and `_qmoe_fusion_test.py` to reflect the corrected (dtype-matching) behavior, and added explicit FP32/FP16/BF16 coverage in `_qmoe_fusion_test.py`.

## Validation

- `python -m pytest src/mobius/components/_moe_test.py src/mobius/rewrite_rules/_qmoe_fusion_test.py -v` — 39 passed.
- `ruff` — passed.
- Real-model validation: re-ran the Olive `Rtn(moe=true)` + `MobiusBuilder(precision=fp16, CUDA EP)` export against `Qwen/Qwen3.6-35B-A3B`. All 430 QMoE scale initializers in the exported graph are now FLOAT16 (previously FLOAT32).
- Loaded the re-exported model with `onnxruntime` (verbose session logging): **zero** "CUDA kernel not found in registries" messages for QMoE, and all 40 QMoE nodes are now correctly assigned to `CUDAExecutionProvider` (previously all 40 fell back to CPU).
- `onnxruntime_genai.Model(...)` now loads successfully with `enable_cuda_graph=1` in `genai_config.json`, with no manual workaround needed (previously required manually stripping `enable_cuda_graph` from the generated config).

Related: builds on top of the export fixes merged in #495.

### DIFF
--- a/src/mobius/components/_moe.py
+++ b/src/mobius/components/_moe.py
@@ -371,7 +371,6 @@ class MoELayer(nn.Module):
             dtype=ir.DataType.UINT8,
         )
         self.fc1_scales = nn.Parameter([self.num_experts, fc1_out, hidden_size // block_size])
-        self.fc1_scales._keep_float32 = True
         self.fc2_experts_weights = nn.Parameter(
             [self.num_experts, hidden_size, intermediate_size * bits // 8],
             dtype=ir.DataType.UINT8,
@@ -379,7 +378,6 @@ class MoELayer(nn.Module):
         self.fc2_scales = nn.Parameter(
             [self.num_experts, hidden_size, intermediate_size // block_size]
         )
-        self.fc2_scales._keep_float32 = True
         if quantization.sym:
             self.fc1_experts_zero_points = None
             self.fc2_experts_zero_points = None
--- a/src/mobius/components/_moe_test.py
+++ b/src/mobius/components/_moe_test.py
@@ -244,8 +244,8 @@ class TestMoELayer:
 
         _cast_module_dtype(layer, ir.DataType.FLOAT16)
         assert layer.gate.weight.dtype == ir.DataType.FLOAT16
-        assert layer.fc1_scales.dtype == ir.DataType.FLOAT
-        assert layer.fc2_scales.dtype == ir.DataType.FLOAT
+        assert layer.fc1_scales.dtype == ir.DataType.FLOAT16
+        assert layer.fc2_scales.dtype == ir.DataType.FLOAT16
 
     def test_expert_major_packing_matches_static_64_expert_top6_reference(self):
         """Packed QMoE math matches the existing loop-over-experts semantics."""
--- a/src/mobius/rewrite_rules/_group_query_attention_test.py
+++ b/src/mobius/rewrite_rules/_group_query_attention_test.py
@@ -327,12 +327,14 @@ class TestGroupQueryAttentionRules:
             # computed in FLOAT32 for routing-score numerical stability.
             assert node.inputs[1] is not None
             assert node.inputs[1].dtype == ir.DataType.FLOAT16
-            # fc1_scales/fc2_scales (indices 3, 6) are quantization scale
-            # tensors, independent of the hidden_states/router dtype.
+            # fc1_scales/fc2_scales (indices 3, 6) must match the
+            # hidden_states/activation dtype (FLOAT16 here): QMoE's kernel
+            # registration requires T2 (scales) to exactly equal T
+            # (activation), so scales are no longer pinned to FLOAT32.
             assert node.inputs[3] is not None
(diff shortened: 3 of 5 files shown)


titaiwangms:
## Full review pass — findings & fixes

Ran a 4-reviewer pass (code / critical / deep / integration) over this PR. Summary:

**Confirmed & fixed (this commit):**
- **CI-breaking regression** (integration-reviewer, confirmed by actually running the test): `_group_query_attention_test.py::test_deepseek_v2_lite_int4_uses_one_qmoe_per_moe_layer` was a pre-existing test not touched by the original fix — it hardcoded the *old* buggy `FLOAT` expectation for QMoE scale dtypes. Fixed the assertions + stale comment.
- **Vacuous test parametrization** (deep-reviewer): the new BF16/FLOAT/FLOAT16 parametrized tests in `_qmoe_fusion_test.py` built synthetic `MatMulNBits` fixtures with scales hardcoded to FLOAT16 regardless of the parametrized `activation_dtype`, so they didn't really exercise mixed dtype packing. Fixed by threading `activation_dtype` into the fixture.
- **Coverage gap in the new dtype guard** (deep-reviewer): `fuse_dense_moe_to_qmoe`'s guard skipped fusion whenever `layer.hidden.dtype is None`, which is common for untyped graph-internal values in real models (not just the schema-invalid case it was meant to catch). Added a fallback to an existing expert's `MatMulNBits` scale dtype, plus a regression test (`test_fuses_when_hidden_dtype_is_untyped`).
- Minor: added the missing `fc2` scale content assertion to `test_scales_match_activation_dtype` (previously only `fc1` was checked).

**Verified correct, no fix needed:**
- deep-reviewer independently confirmed the core fix's premise against ORT source: `T2 == T` (scales must match activation dtype) is the only kernel-registered combination for `quant_type="int"` on both CPU and CUDA. Also flagged that WebGPU's QMoE registration is *untyped*, meaning the pre-fix bug could have silently produced wrong results there (not just failed to dispatch) — good independent confirmation this was a real bug worth fixing.

**Discussed, deferred as follow-up (not blocking this PR):**
- critical-reviewer raised two Major points: (1) no numerical-representability check before the FP16/BF16 scale downcast (extreme-dynamic-range scales could silently overflow/underflow), and (2) the "validated end-to-end" claim in the PR description isn't covered by the repo's own test suite (tests check IR-level dtypes/NumPy math, not real ORT `InferenceSession` CUDA kernel dispatch). Both are reasonable but larger-scope asks than this bugfix; flagging here for a follow-up issue rather than blocking merge, since the manual end-to-end validation (~30x speedup, all 40 QMoE nodes on CUDA EP) is described in the PR body.

Full suite after fixes: **1963 passed, 2 skipped**. `ruff check` + `ruff format --check` clean.

justinchuby (approved):

PR #511, opened 2026-08-17
Fix Qwen3.5/3.6-MoE VL decoder QMoE quantization support

titaiwangms:
## Summary

`Qwen35MoEVL3ModelCausalLMModel.preprocess_weights` (the 3-model-split vision-language wrapper for Qwen3.5/3.6-MoE) previously had **zero quantization awareness**, unlike its text-only sibling `Qwen35MoECausalLMModel`. An Olive-quantized VL checkpoint would be silently mishandled: quantized MoE expert weights would be routed as if float, producing a broken/incomplete ONNX export with no error raised.

This is one of the pieces needed to support end-to-end quantized export of `Qwen/Qwen3.6-35B-A3B` (256-expert MoE VL model) via `MobiusBuilder`, alongside microsoft/Olive#2628 and microsoft/Olive#2457.

## Changes

- `use_qmoe` detection (via `_supported_qmoe_quantization`), mirroring the text-only sibling.
- A guard that raises a clear `ValueError` when packed-but-unsupported quantized expert weights are found, instead of silently falling back to the (incompatible) dense per-expert path.
- `preprocess_olive_weights` + `pack_qmoe_expert_weights` calls to rename and pack Olive-quantized decoder MoE expert weights into the native QMoE ONNX parameter layout (`target_moe_path=".mlp"`, matching the text-only sibling's `Qwen35MoEBlock` placement).
- Explicit `NotImplementedError` guards for:
  - `quant_method` in `{"gptq", "awq"}` — not yet supported by this VL wrapper (only Olive-quantized checkpoints are handled today).
  - `quantize_embeddings` / `quantize_lm_head` — no quantized `Embedding`/`Linear` module variants are wired up anywhere in this 3-model split, so these flags previously produced orphaned initializers with no error.
  
  Both guards fail loudly instead of silently producing a broken export.
- Honors `quantization.tie_word_embeddings` (not just `config.tie_word_embeddings`) when deciding whether to duplicate `lm_head` from `embed_tokens`, for internal consistency with the recovered "effective tie" value.

## Tests

New `TestQwen35MoEVL3ModelQMoEExport` class covering:
- QMoE packing/binding (Olive-quantized)
- Dense fallback (unquantized)
- Unsupported-ABI guard (packed-but-non-QMoE quantization)
- Vision/embedding/MTP-routing preservation (unaffected by decoder quantization)
- GPTQ/AWQ rejection (clear error)
- Quantized-embeddings/lm_head rejection (clear error)

```
python -m pytest src/mobius/models/qwen35_test.py -v
10 passed
```

Full model suite: `python -m pytest src/mobius/models/ -q` → 401 passed, 3 pre-existing unrelated failures (confirmed present on clean `main`: `deepseek_v4_flash_test.py::test_ort_genai_sidecar_metadata`, `qwen_image_test.py::test_cuda_matches_cpu`, `world_model_test.py::test_mlp_world_model_matches_pytorch`).

`ruff check` / `ruff format --check`: clean.

## Review

This change went through a full internal review pass (readability, correctness, adversarial/critical, spec-adherence, and cross-module integration review, plus independent test execution) before being opened. Findings from that review (GPTQ/AWQ gap, quantized-embedding binding gap) are addressed via the explicit guards above rather than full implementations, since neither is needed by the current Olive→Mobius pipeline for Qwen3.6-35B-A3B.


src/mobius/models/qwen35.py
```diff
@@ -759,16 +806,31 @@ def preprocess_weights(
...
-                elif suffix.endswith(".mlp.experts.down_proj"):
+                elif not use_qmoe and suffix.endswith(".mlp.experts.down_proj"):
                     prefix = target[: -len("experts.down_proj")]
                     for i in range(value.shape[0]):
                         renamed[f"{prefix}experts.{i}.down_proj.weight"] = value[i]
                 else:
                     renamed[target] = value
+
+        if quantization is not None and quantization.quant_method == "olive":
+            renamed = preprocess_olive_weights(
```
justinchuby: Do we need to call this for olive for all models? Or is there a way to reduce duplication?

titaiwangms: Done

src/mobius/components/_moe.py
```diff
@@ -11,9 +11,12 @@
 import onnx_ir as ir
 from onnxscript import OpBuilder, nn
 
-from mobius._configs import ArchitectureConfig, QuantizationConfig
+from mobius._configs import ArchitectureConfig
+from mobius._weight_utils import supported_qmoe_quantization
```
justinchuby: as _supported_qmoe_quantization

src/mobius/_weight_utils.py
```diff
@@ -917,3 +953,163 @@ def preprocess_awq_weights(
...
+    )
+    if (
+        use_qmoe
+        and quantization is not None
+        and quantization.quant_method not in qmoe_quant_methods
+    ):
+        methods = tuple(qmoe_quant_methods)
+        if methods == ("olive",):
+            raise NotImplementedError(
+                f"{model_name} currently only supports QMoE export for "
```
justinchuby: Is model name needed if only for errors? Can we say "This model..." given the user understands which model is passed in?

justinchuby (approved):

src/mobius/components/_moe.py
titaiwangms: Done — imported directly as `_supported_qmoe_quantization` instead of the extra assignment.

src/mobius/_weight_utils.py
titaiwangms: Done — dropped the `model_name` parameter; the messages now say "This model ...".

PR #518, opened 2026-08-19
fix(gguf): correct Gemma-2 GGUF export (config, norms, tokenizer BOS)

justinchuby:
## Summary
Building a **standalone Gemma-2-2b-it int4 CUDA text export** from a llama.cpp Q4_K_M GGUF (`build-gguf --keep-quantized`) surfaced a cascade of Gemma-2 GGUF-path bugs. Each is fixed and unit-tested. With all five fixes, the export decodes **coherently on the native CUDA EP, byte-identical to the reference HF f32 export**.

## Bugs fixed
| # | Symptom | Fix | File |
|---|---------|-----|------|
| 1 | `AttributeError: query_pre_attn_scalar` — GGUF path handed Gemma2 a plain `ArchitectureConfig` | Add `_gemma2_postprocess` (softcapping + `query_pre_attn_scalar`), register for arch `gemma2` | `_config_mapping.py` |
| 2 | `rotary_emb is None` — Gemma2 GGUF omits all `rope.*` keys (llama.cpp hardcodes θ=10000) → `rope_type=None` → RoPE disabled | Force `rope_type="default"` in the postprocessor | `_config_mapping.py` |
| 3 | `52 initializers without weights` — `_GEMMA2_EXTRAS` used fictional tensor names | Map real llama.cpp names (`ffn_norm`/`post_attention_norm`/`post_ffw_norm`); dedicated `gemma2` dispatch branch (gemma v1 stays on Llama base) | `_tensor_mapping.py` |
| 4 | Garbage output — double norm offset. llama.cpp bakes `w_gguf = w_hf + 1`, and `OffsetRMSNorm` re-adds `+1` | `_process_gemma` now **subtracts 1** (was adding). Verified **byte-exact** vs reference HF export (norm corr = 1.0000) | `_tensor_processors.py` |
| 5 | **Single-token repetition loop** (`"HelHelHel…"`). transformers' GGUF loader returns a tokenizer (Unigram, for Gemma) whose post-processor does **not** prepend `<bos>`; Gemma requires BOS | `_ensure_bos_post_processor` restores BOS from GGUF `add_bos_token`/`bos_token_id` metadata, with an encode-probe to avoid a doubled BOS | `_tokenizer.py` |

Bug 5 was the final blocker: with correct weights/config/norms, the model still looped because the emitted tokenizer dropped BOS. Restoring it yields coherent decode.

## Validation
- `pytest src/mobius/integrations/gguf/` — all directly-affected tests pass (config/tensor/tokenizer). Added a focused regression test `test_restores_bos_post_processor_when_primary_path_omits_it`.
- End-to-end: the resulting Gemma-2-2b (head_dim=256) export decodes coherently on the native CUDA EP and matches the reference HF f32 export token stream.

## Notes
- The `_process_gemma` sign convention is **shared by gemma/gemma2/gemma3**; validated on gemma2 — gemma3 inherits the corrected convention.
- 3 pre-existing failures (`ir.save() unexpected kwarg max_shard_size_bytes` in `_builder_test`/`_mmproj_test`) are an **onnx-ir version mismatch on this base**, reproduce without this change, and are untouched here.

> Authored by the Squad Coding Agent (Deckard). User pre-authorized this Mobius export fix.

### DIFF
--- a/src/mobius/integrations/gguf/_config_mapping.py
+++ b/src/mobius/integrations/gguf/_config_mapping.py
@@ -28,7 +28,7 @@ from typing import Any
 
 import numpy as np
 
-from mobius._configs import ArchitectureConfig, Gemma4Config
+from mobius._configs import ArchitectureConfig, Gemma2Config, Gemma4Config
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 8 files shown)


justinchuby:
@copilot please fix ci

PR #520, opened 2026-08-19
Correct upstream runtime assets that ship broken

justinchuby:
## The problem

Weights are not the whole package. The chat template, tokenizer config and processor config decide whether an agent client can actually drive the model. When one of them is wrong the export is faithful and the deployment is still broken.

`meta-models/Muse-Glimmer-30B` has such a defect. Its chat template derives a namespace from every offered tool name and renders it as a wildcard:

```jinja
{%- set tns = fn.name.split('.')[0] -%}
{%- set rns.recipients = rns.recipients + ['"' + tns + '.*"'] -%}
```

`split('.')[0]` on an unqualified name returns the name itself, so a client offering plain tool names gets a preamble reading:

```
# Valid recipients: "self", "read.*", "bash.*", "glob.*", "user".
```

Nothing named `read` is addressable any more, only things *under* `read`. The model does what the preamble asks and supplies a second segment, usually by borrowing a parameter name, and emits `read.filePath`. Observed against every OpenCode session: the first tool call of each task failed and the agent recovered on retry, so every task paid for a wasted round trip.

Fixing the file in an exported package by hand does not stick. The defect returns on the next re-export and on every other machine.

## The change

A `mobius.upstream_patches` subpackage owns this class of correction.

```
data/<owner>/<model>/
    patch.json       what to change, and which upstream revision it was written against
    README.md        the symptom, the defect, and why this is the right fix
    <target>.patch   a unified diff
```

Two decisions carry the design.

**A correction is a diff, not a copy.** Checking in a whole corrected file would pin every unrelated byte in it to whatever upstream happened to have that day. A diff states only the change, and it is applied strictly -- the hunk lands at the line it names with the context it names, no fuzz and no offset search.

**A correction is selected by content hash, not by model id.** A source may be a Hub repo, a snapshot directory or a local checkout; the sha256 of the file is the same in all three, and there is no id to thread through the export. It also means a correction can only ever rewrite bytes it recognises.

Together these make going stale safe. When upstream republishes the file the hash stops matching, the correction stops being applied, and nothing is corrupted. `python -m mobius.upstream_patches.check_upstream` reports that drift so it can be re-reviewed -- upstream may have fixed it, or restructured around it.

A second operation kind, `sync_chat_template`, rewrites the `chat_template` key duplicated into `tokenizer_config.json`. Packages whose two copies disagree fail in a way that depends on which loader you use. It runs only when the template it mirrors was actually corrected.

Both export paths -- `onnx_genai.inference_metadata._copy_runtime_assets` and `ort_genai.auto_export` -- apply corrections after materializing runtime assets.

## Verification

- 15 new unit tests covering the diff applier (context mismatch, multi-hunk, missing trailing newline, empty context line) and the applier (idempotence, unrelated keys surviving the JSON rewrite, a package upstream already fixed being left alone).
- Applied to a real `meta-models/Muse-Glimmer-30B` snapshot through `_copy_runtime_assets`: output is byte-identical to the template that was verified end to end through OpenCode against the INT4 CUDA export, with a read task and a multi-tool debug task (read, edit, bash, todowrite) completing with no invalid tool errors.
- `check_upstream` confirms the recorded hash still matches upstream `main`, i.e. the defect is live.
- `ruff check` and `ruff format` clean.
- 9 failures in `ort_genai/auto_export_test.py::TestWriteProcessorConfig` are pre-existing (`AutoProcessor` import), reproduced identically on a clean tree.

## Follow-up

The defect has been reported upstream. This correction should be deleted once it lands there.

### DIFF
--- a/src/mobius/upstream_patches/_patches.py
+++ b/src/mobius/upstream_patches/_patches.py
@@ -1,0 +1,163 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+"""Corrections to runtime assets that ship broken from upstream model repos.
+
+A model package is more than weights. It carries a chat template, a tokenizer
+config, a processor config -- files that decide whether an agent client can
+actually drive the model. When one of those is wrong the export is faithful
+and the deployment is still broken, and the defect comes back every time
+somebody re-exports.
+
+The corrections live in ``data/<owner>/<model>/`` as unified diffs written
+against a named upstream revision, next to a ``README.md`` explaining the
+symptom that motivated them. A patch is selected by the sha256 of the file
+already in the package, not by a model id, so it finds its target whether the
+source was a Hub repo, a snapshot directory, or a local checkout -- and it can
+only ever rewrite bytes it recognises.
+
+That hash, plus the strictness of the diff applier, is how upstream is
+tracked: when upstream republishes the file, the patch stops matching and
+stops being applied instead of silently corrupting a package.
+"""
+
+from __future__ import annotations
+
+import dataclasses
+import hashlib
+import json
+import logging
+import pathlib
+from typing import Any
+
+from mobius.upstream_patches._diff import PatchError, apply_unified_diff
+
+_LOGGER = logging.getLogger(__name__)
+
+_DATA_ROOT = pathlib.Path(__file__).parent / "data"
+
+
+@dataclasses.dataclass(frozen=True)
+class AssetPatch:
+    """One upstream correction, loaded from ``data/<owner>/<model>/patch.json``."""
+
+    directory: pathlib.Path
+    upstream_repo: str
+    upstream_revision: str
+    summary: str
+    operations: tuple[dict[str, Any], ...]
+
+    @property
+    def name(self) -> str:
+        return self.upstream_repo
+
+
+def _load(directory: pathlib.Path) -> AssetPatch:
+    with (directory / "patch.json").open(encoding="utf-8") as handle:
+        raw = json.load(handle)
+    upstream = raw["upstream"]
+    return AssetPatch(
+        directory=directory,
+        upstream_repo=upstream["repo"],
+        upstream_revision=upstream["revision"],
+        summary=raw.get("summary", ""),
+        operations=tuple(raw["operations"]),
+    )
+
+
+def available_patches(root: pathlib.Path | None = None) -> list[AssetPatch]:
+    """Return every correction shipped with this package."""
+    base = _DATA_ROOT if root is None else root
+    return [_load(path.parent) for path in sorted(base.glob("*/*/patch.json"))]
+
+
+def _sha256(path: pathlib.Path) -> str:
+    return hashlib.sha256(path.read_bytes()).hexdigest()
+
+
+def _apply_diff(package: pathlib.Path, patch: AssetPatch, operation: dict) -> str | None:
+    target = package / operation["target"]
+    if not target.is_file():
+        return None
+    if _sha256(target) != operation["upstream_sha256"]:
+        return None
+
+    diff = (patch.directory / operation["patch"]).read_text(encoding="utf-8")
+    try:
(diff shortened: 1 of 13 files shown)


src/mobius/upstream_patches/data/meta-models/Muse-Glimmer-30B/README.md
```diff
@@ -0,0 +1,47 @@
...
+The model does what the preamble asks and supplies a second segment,
+usually by borrowing a parameter name, and emits a call to `read.filePath`.
+The harness rejects it because no such tool was offered.
+
+Observed against every OpenCode session: the first tool call of each task
+failed, the agent recovered on retry, and every task paid for the round trip.
+
+## The fix
+
+Treat the dot as the signal it is. A name that contains one is namespaced and
```
justinchuby, line 31: Should it start with `functions.` like gpt oss?

justinchuby: No — and the template answers this itself, a few lines below the block I patched. That paragraph is upstream's, untouched by this PR:

```
(If the tool namespace is not specified, invoke the function directly as
`example_function_name` rather than `example_tool_name.example_function_name`)
```

So upstream's own instruction to the model is that an unqualified name is invoked **bare**. The bug was never that bare names lack a namespace; it was that `render_system_meta` contradicted this paragraph by advertising `"read.*"` while the paragraph told the model to call `read`. Prefixing `functions.` would preserve the contradiction and just move it — the recipients line would read `"functions.*"`, the model would emit `to=functions.read`, and the harness would reject it for the same reason it rejected `read.filePath`. We'd be trading one invented segment for another.

The `functions.` convention is specific to OpenAI harmony, and this model isn't harmony lineage:

- no `<|channel|>`, no `<|constrain|>`, no `namespace functions { ... }` anywhere in the template (`grep -c` = 0 for all three);
- BOS is `<|begin_of_text|>`, not harmony's `<|start|>`-only framing;
- tools are declared as `// Tool metadata` JSON lines and invoked through an `<atem:function_calls>` XML block, not harmony's TypeScript-namespace preamble with `<|constrain|>json`.

`functions.` in gpt-oss isn't decoration — it's the name of the TypeScript namespace harmony emits, and calls resolve against it. There's no such namespace here to resolve against, so the prefix would be a bare string the model has no training signal for.

Rendered with two unqualified tools on the patched template, the preamble now agrees with the paragraph:

```
# Valid recipients: "self", "read", "bash", "user".
```

And a client that *does* offer namespaced names still gets `"ns.*"` exactly as before, so this stays a correction of upstream's mangling rather than a new convention.

src/mobius/upstream_patches/_patches.py
```diff
@@ -0,0 +1,163 @@
...
+def _sync_chat_template(package: pathlib.Path, operation: dict) -> str | None:
+    """Copy a patched template over the duplicate embedded in a JSON config."""
+    target = package / operation["target"]
+    source = package / operation["source"]
+    if not target.is_file() or not source.is_file():
+        return None
+
+    template = source.read_text(encoding="utf-8")
+    with target.open(encoding="utf-8") as handle:
+        config = json.load(handle)
```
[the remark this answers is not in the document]
justinchuby: Fixed in 15bf27f. You're right that this broke the contract `_apply_diff` sets: that one never raises — a hash mismatch returns `None`, a diff that won't apply is logged and skipped — because correcting upstream assets is opportunistic and must not fail an export that would otherwise succeed. Letting `json.load` propagate meant a malformed or non-UTF-8 `tokenizer_config.json` could take the whole export down over a *duplicate* copy of a file we had already corrected in place.

Now logged and skipped, catching `JSONDecodeError` and `UnicodeDecodeError`. Covered by `test_tolerates_an_unreadable_tokenizer_config`, which also asserts the unreadable config is left exactly as found rather than truncated — worth pinning because the write path opens with `"w"`, so returning early rather than failing partway through is what keeps it intact.

PR #521, opened 2026-08-19
Honor generation config token IDs in ORT GenAI export

titaiwangms:
## Summary

- load optional Hugging Face `generation_config.json` when writing ORT GenAI artifacts
- prefer generation-config BOS/EOS/PAD IDs over model-config values
- preserve multi-EOS stop tokens such as Qwen3.6's `[<|im_end|>, <|endoftext|>]`

Qwen3.6 defines scalar EOS `248044` in its text model config but `[248046, 248044]` in its generation config. Ignoring the latter causes generated ORT GenAI packages to miss `<|im_end|>` as a stopping token.

## Testing

- `python -m pytest -q src/mobius/integrations/ort_genai/auto_export_test.py` (112 passed)
- `ruff check src/mobius/integrations/ort_genai/auto_export.py src/mobius/integrations/ort_genai/auto_export_test.py`
- `ruff format --check src/mobius/integrations/ort_genai/auto_export.py src/mobius/integrations/ort_genai/auto_export_test.py`

### DIFF
--- a/src/mobius/integrations/ort_genai/auto_export.py
+++ b/src/mobius/integrations/ort_genai/auto_export.py
@@ -188,6 +188,19 @@ def _select_ort_model_type(
     return _resolve_ort_genai_model_type(hf_model_type or "unknown")
 
 
+def _load_generation_config(model_id: str):
+    """Load optional Hugging Face generation settings without requiring the file."""
+    import transformers
+
+    try:
+        return transformers.GenerationConfig.from_pretrained(model_id)
+    except OSError:
+        logger.debug(
+            "No generation_config.json found for %s; using model config token IDs", model_id
+        )
+        return None
+
+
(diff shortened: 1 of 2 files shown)


justinchuby (approved):
Just making sure: we are loading all the possible information from the generation config with this change?

titaiwangms:
Not all generation settings. This change intentionally loads only `bos_token_id`, `eos_token_id`, and `pad_token_id`, since those are model-level token IDs represented directly in the ORT GenAI config and the concrete bug here is losing Qwen3.6’s multi-EOS list. Sampling controls such as `do_sample`, `temperature`, `top_k`, `top_p`, and length limits continue to use Mobius defaults or caller-supplied `GeneratorParams`; importing them here would be a broader generation-policy behavior change. We can map additional fields separately if that is desired.

PR #378, opened 2026-06-25
Add component inspection API

xiaoyu-work:
## Summary
- add mobius.inspect_components and ComponentInfo
- report ModelPackage component names/kinds without building graphs or loading weights
- include source_path hints for Qwen VL components

## Validation
- PYTHONPATH=$PWD/src conda run -n olive python -m pytest src/mobius/_inspect_test.py -q

PR #525, opened 2026-08-20
Fix Olive-quantized Qwen3 MoE export

titaiwangms:
## Summary

- preserve packed Olive/GPTQ/AWQ fused expert tensors through generic MoE weight renaming so the QMoE packer receives the original expert-major layout
- keep unquantized fused-expert splitting unchanged and centralize packed-sidecar suffix detection
- emit the supported ORT GenAI `qwen3` runtime type for `qwen3_moe`, preserving Qwen3 reasoning token metadata
- add weight-preprocessing, graph, runtime-config, and regression coverage

This completes the generic Qwen3-MoE KQuant path built on #517. Before this change, packed `gate_up_proj_qweight` and `_scales` sidecars collided after being split into the same per-expert `.weight` keys, causing a deterministic weight-shape mismatch during export.

## Validation

- `220 passed` — focused MoE, weight utility, Qwen3.5, and QMoE fusion tests
- `94 passed, 8 skipped` — quantized/MoE graph tests
- `192 passed` — ORT GenAI integration tests
- Ruff check and format passed on all changed Python files
- full `Qwen/Qwen3-30B-A3B` Olive KQuant + Mobius export completed successfully
- exported 16 GB package loaded with ORT GenAI 0.16.0-dev on an A100 80 GB
- Qwen3 reasoning IDs resolved correctly (`bor=151667`, `eor=151668`)
- CUDA greedy generation produced coherent output and correctly answered `17 * 23 = 391`


titaiwangms:
The current all-files **Lint** failure is inherited from `main` after #378: `src/mobius/_inspect_test.py` is not Ruff-formatted. I opened the formatting-only #526 instead of mixing that unrelated change into this PR. The Qwen3-MoE changed files pass Ruff/lintrunner locally and the separate `lintrunner` check is green.

src/mobius/models/moe_test.py
```diff
@@ -0,0 +1,359 @@
...
+        params = dict(model.named_parameters())
+        for name in (
+            f"{_LAYER}mlp.gate.weight",
+            f"{_LAYER}input_layernorm.weight",
+            f"{_LAYER}post_attention_layernorm.weight",
+            "model.norm.weight",
+            "model.embed_tokens.weight",
+            "lm_head.weight",
+        ):
+            assert params[name].dtype.is_floating_point(), f"{name} must stay float"
```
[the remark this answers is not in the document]
titaiwangms: This assertion is valid with the pinned `onnx-ir`: `params[name].dtype` is `onnx_ir.DataType`, and `DataType.is_floating_point` is a bound method. I verified `callable(ir.DataType.FLOAT.is_floating_point) == True` and `ir.DataType.FLOAT.is_floating_point() == True`; the test also passes in the focused suite and all Ubuntu CI versions. No change is needed here.

src/mobius/_weight_utils.py
```diff
@@ -1121,18 +1139,24 @@ def preprocess_quantized_weights(
                 "supported_qmoe_quantization) for MoE models instead."
             )
 
-    def _is_packed_key(key: str, float_key: str) -> bool:
+    def _is_packed_sidecar_of(key: str, float_key: str) -> bool:
+        """Whether ``key`` is a packed sidecar of the specific ``float_key``.
+
+        Unlike the module-level :func:`is_packed_quant_key` suffix predicate,
+        this matches the *exact* sidecar keys of one named float parameter.
```
justinchuby, line 1156: What is a sidecar?

src/mobius/_weight_utils.py
```diff
@@ -24,6 +24,32 @@
 
 logger = logging.getLogger(__name__)
 
+# Key suffixes marking a *packed quantized* sidecar tensor rather than a float
+# weight. Olive appends them with an underscore directly to the parameter name
+# (``<pname>_qweight``, see Olive's ``olive/common/quant/state_dict.py``), while
+# GPTQ/AWQ store dotted sibling buffers on the owning module
+# (``<module>.qweight``). Both conventions occur in raw HF checkpoints.
+OLIVE_PACKED_QUANT_SUFFIXES = ("_qweight", "_scales", "_qzeros")
```
justinchuby: Use a frozen set?

justinchuby (approved):

src/mobius/_weight_utils.py
titaiwangms: Added a definition where the term is first introduced (module-level comment above `OLIVE_PACKED_QUANT_SUFFIXES`): a *sidecar* is an auxiliary tensor (`qweight`/`scales`/`qzeros`) that rides alongside a quantized parameter's canonical `.weight` key instead of replacing it — one logical weight is split across several sidecar keys until unpacked. See 2f100bf2.

src/mobius/_weight_utils.py
titaiwangms: Done — `OLIVE_PACKED_QUANT_SUFFIXES` and `DOTTED_PACKED_QUANT_SUFFIXES` are now `frozenset[str]`. Kept `PACKED_QUANT_SUFFIXES` (the combined one used by `str.endswith()`) as a `tuple`, since `endswith()` requires a tuple, not a set — added a comment explaining why. See 2f100bf2.

PR #526, opened 2026-08-20
Restore formatting and Windows asset-patch tests

titaiwangms:
## Summary

- apply the repository Ruff formatter to two long calls added by #378
- write the asset-patch SHA fixture as exact UTF-8 bytes so Windows newline translation cannot change LF to CRLF before hashing
- restore the all-files lint job and the four asset-patch tests failing across all Windows Python versions on `main` and dependent pull requests

## Validation

- `9 passed, 1 subtests passed` — `src/mobius/upstream_patches/_patches_test.py`
- Ruff check and format passed for both changed test files
- `git diff --check`

Only test formatting/fixture construction changes; runtime behavior is unchanged.


justinchuby:
thanks!

justinchuby (approved):

PR #529, opened 2026-08-21
feat(gguf): auto-detect + export Qwen3.5/3.8 MTP self-speculative head from GGUF

justinchuby:
## What

Recovers the **Multi-Token-Prediction (MTP / "nextn") self-speculative head** that the Qwen3.8-27B ONNX export was silently dropping, and emits it **automatically whenever the source GGUF ships the head** — no CLI flag, no config toggle. The MTP sidecar is a purely-additive artifact (`mtp/model.onnx` + shared embedding/lm_head + `SpeculatorConfig` metadata) that existing consumers ignore, so it is driven entirely by source-tensor presence: **emit MTP iff available.**

## Auto-detect (presence-driven, no flag)

The decision is made by `has_mtp_head(config)` — true when the config carries `_gguf_mtp_block_indices`, which `gguf_to_config` populates from the source GGUF. **Exact presence check:**

- Metadata key `<arch>.nextn_predict_layers > 0` (for Qwen3.8-27B: `qwen35.nextn_predict_layers = 1`, `block_count = 65`), which yields `_gguf_mtp_block_indices = [decoder_layers .. decoder_layers + nextn - 1]` (here `[64]`), and
- the trailing block's `blk.<N>.nextn.*` tensors — `nextn.eh_proj` (the `fc`), `nextn.enorm`, `nextn.hnorm`, `nextn.shared_head_norm` — plus that block's own attention (`attn_q/k/v/output`, `attn_q_norm/k_norm`, `attn_norm`, `post_attention_norm`) and FFN (`ffn_gate/up/down`) sublayer.

Behavior:
- **Source has the nextn head** -> always emit the `mtp/model.onnx` sidecar + the backbone `hidden_states.<N-1>` seed output + a `SpeculatorConfig` (`proposal_type: mtp`) block in `inference_metadata.yaml`.
- **Source has no nextn head** -> emit nothing; text-only output is byte-for-byte identical to today.
- `--static-cache` skips the head (it needs the dynamic concat-grow cache), logged as info; those exports are unchanged.

MTP dims are derived entirely from GGUF/source metadata (hidden size, heads, head_dim, KV heads, rope, quant) — no hardcoding.

## Root cause

The dense Qwen3.5/3.8 GGUF (e.g. `unsloth/Qwen3.8-27B`) ships a trained MTP head as the **trailing `blk.<N>` block** (for 27B: `blk.64`), tagged by `<arch>.nextn_predict_layers`. `gguf_to_config` read that key **only to subtract the block** from `num_hidden_layers`, and the text-only backbone build never made a second pass — so the head tensors were dropped.

## Changes

- **`integrations/gguf/_config_mapping.py`** — surface `_gguf_nextn_predict_layers` / `_gguf_mtp_block_indices` (the trailing block previously only subtracted).
- **`integrations/gguf/_mtp.py`** (new) — `has_mtp_head` (presence check), `map_gguf_mtp_to_hf_names` (GGUF `blk.<N>.nextn.*` + attn/ffn -> `Qwen35MtpModel` stems), `derive_mtp_config` (**all dims inherited from the backbone config — no hardcoding**), `build_mtp_head_from_gguf` (reuses `Qwen35MtpModel`/`Qwen35MtpTask` and the existing quantized/dequantized loaders via an injectable `name_mapper`; `fc` is quantized on the same main-layer path). The two `pre_fc_norm_*` `OffsetRMSNorm` `+1` offsets missed by the generic norm strip are handled explicitly.
- **`integrations/gguf/_builder.py`** — steps 4b (seed output) and 10 (attach `pkg.mtp_head`) gate on `has_mtp_head(config)` alone (auto-detect); `static_cache` skips the head via a logged info. The quantized/dequantized loaders gained optional `name_mapper` / `warn_unmapped` params (backward compatible).
- **`__main__.py`** — when `pkg.mtp_head` is present, save it into `output_dir/mtp/` and write speculator metadata. (No new CLI flag.)
- **`integrations/onnx_genai/inference_metadata.py`** — `write_mtp_speculator_metadata` (`proposal_type: mtp`, `target_hidden_output`, `mtp_hidden`, single-layer KV).
- **`integrations/gguf/_mtp_test.py`** (new) — mapping, config derivation, tiny-GGUF head emission, **and auto-detect** (nextn GGUF auto-emits the head with `mtp_hidden` output; plain GGUF omits it).

## Validation (partial — stated explicitly)

- **Unit + tiny-GGUF: 9 MTP tests PASS**, incl. auto-detect through `build_from_gguf` (source with nextn -> `pkg.mtp_head` attached; source without -> absent). Emitted head initializers include `fc.weight_t`, `pre_fc_norm_embedding.weight`, `pre_fc_norm_hidden.weight`, `norm.weight`, `layers.0.self_attn.q_proj.weight_t`; graph output `mtp_hidden`; every initializer has backing weights. `_config_mapping` / `_tensor_mapping` suites still pass (74 total).
- **Real 16 GB GGUF `blk.64` shape contract VERIFIED** against the metadata-derived head — all 15 head tensors map, backbone excluded:
  - `attn_q` = (12288, 5120) = (2*NH*HD, H) -> **doubled-Q output gating**
  - `attn_k`/`attn_v` = (1024, 5120) = (NKV*HD, H); `attn_output` = (5120, 6144) = (H, NH*HD)
  - `nextn.eh_proj` = (5120, 10240) = (H, 2H) — the `fc` GEMM (Q8_0 while projections are Q4_0; mixed quant handled per-tensor)
- **Not run:** a full 27B int4 head export through onnxruntime (heavy; the local venv's older `onnx_ir` lacks `max_shard_size_bytes`, which also breaks the pre-existing `test_default_quantized_package_save_reload` — unrelated to this change).

## Real artifact produced + two export-path fixes

Producing the real int4/bf16 27B artifact surfaced two defects that made the MTP sidecar silently degenerate. Both are now fixed (commit `987c8d04`):

1. **`Qwen35TextModel.forward` ignored `output_layer_indices`** — it always returned a 2-tuple, so the backbone never emitted the `hidden_states.{N-1}` seed the head consumes; `speculator.target_hidden_output` pointed at a non-existent output. Added the same capture path as `base.py` `TextModel` (returns the selected per-layer post-residual hidden states so `HybridCausalLMTask` emits `hidden_states.{idx}`).
2. **`build_from_gguf` dropped the `_gguf_*` nextn metadata under `--dtype`** — an explicit dtype (or quantization) triggers `dataclasses.replace`, which returns a fresh config without the plain `_gguf_*` attributes, so `has_mtp_head` saw nothing and skipped the head. Now captured into locals right after `gguf_to_config` (like `_gguf_model_type`) and re-attached before step 4b. Regression test added (`test_mtp_survives_dtype_replace`).

`write_mtp_speculator_metadata` also now emits `embedding_weights`, `lm_head_weights`, and `vocab_size`.

### Full export — VALIDATED end-to-end
Command:
```
python -m mobius build-gguf /home/justinchu/qwen38-gguf/Qwen3.8-27B-Q4_0.gguf \
  --output /home/justinchu/qwen38-27b-int4-mtp-cuda --dtype bf16 --runtime onnx-genai
```
- **Layout:** `model.onnx` (2.2 MB) + `model.onnx.data` (17.4 GB), `mtp/model.onnx` (38 KB) + `mtp/model.onnx.data` (354 MB), `inference_metadata.yaml`, `tokenizer.json`. Total **17 GB**.
- **Main model exposes** `hidden_states.63` = `[batch, sequence_len, 5120]` — exactly `speculator.target_hidden_output`.
- **MTP head I/O:** inputs `inputs_embeds`, `hidden_states`, `attention_mask`, `position_ids`, `past_key_values.0.{key,value}`; outputs `mtp_hidden` + `present.0.{key,value}` (single full-attention layer KV). `fc.weight_t` = (10240, 5120); q/k/v/o + gate/up/down projections are int4 `MatMulNBits` (UINT8 packed, block-32) following the main-layer path; `eh_proj` stays bf16 (Q8_0 in source). Shared embedding/lm_head are reused from the main model (named in the speculator block), not duplicated.
- **SpeculatorConfig block:** `proposal_type: mtp`, `num_speculative_tokens: 1`, `model_path: mtp/model.onnx`, `mtp_hidden_output: mtp_hidden`, `kv_mode: hidden_threaded`, `embedding_weights: model.embed_tokens.weight`, `lm_head_weights: lm_head.weight`, `target_hidden_output: hidden_states.63`, `hidden_size: 5120`, `vocab_size: 248320`.
- **Tests:** 53 pass (MTP + qwen35 backbone + qwen35-mtp suites), including the new seed-output and dtype-replace regressions.

## Runtime schema-conformance fix (metadata block)

The first emitted metadata used a `speculator:` block whose field names the onnx-genai runtime cannot parse, so `InferenceMetadata.speculative` deserialized to `None` and the sidecar was silently ignored (text-only decode). Now conformed exactly to the authoritative schema (onnx-genai `crates/onnx-genai-metadata`: `schema/generation.rs` `SpeculatorConfig`, `parser.rs` `resolve_mtp`, `config.rs` `validate_resolved_mtp_config`) — commit `b6747e63`:

- Top-level key `speculator` -> **`speculative`** (the runtime key; alias `speculator_config`). `speculator` was unknown -> dropped.
- `model_path` -> **`model`**; `hidden_size` -> **`target_hidden_size`**.
- `kv_mode: hidden_threaded` (engine-internal enum) -> **`proposal_local`** (only valid `MtpKvMode` for this k=1 head).
- `embedding_weights`/`lm_head_weights` flat strings -> nested **`embedding`/`lm_head`** `MtpTargetInitializer` objects `{source: target_initializer, name: ...}`.
- Added **`target_hidden_layout: BSH`** (rank-3 `[batch,seq,hidden]` seed) + **`hc_mult: 1`** (`resolve_mtp` requires `hc_mult>0`; `validate_resolved_mtp_config` pins it to 1 for BSH).
- Omit `mtp_state_output` (an `hc_mult==1` head has no recurrent Hyper-Connection state; the sidecar emits only `mtp_hidden` + `present.0.{key,value}`).

Emitted block:
```yaml
speculative:
  proposal_type: mtp
  num_speculative_tokens: 1
  model: mtp/model.onnx
  target_hidden_layout: BSH
  hc_mult: 1
  mtp_hidden_output: mtp_hidden
  kv_mode: proposal_local
  embedding: {source: target_initializer, name: model.embed_tokens.weight}
  lm_head: {source: target_initializer, name: lm_head.weight}
  target_hidden_output: hidden_states.63
  target_hidden_size: 5120
  vocab_size: 248320
```
Test `TestMtpSpeculatorMetadata` asserts the exact keys/values and validates the emitted YAML against onnx-genai's published `inference_metadata.schema.json`. (The existing 17 GB artifact's metadata was hand-patched for the benchmark; not re-exported.)

## Why

Unblocks **MTP self-speculative decode** for Qwen3.8-27B batch=1 decode on H200. The engine side already landed in **onnx-genai #1598** (recurrent-state commit-by-accepted-prefix); `mtp.rs`/`MtpDecodeSession` + `SpeculatorConfig` consume exactly this sidecar. Any Qwen3.5/3.8 GGUF that ships the head now exports it end-to-end automatically.


justinchuby:
## Review

Conflicts resolved and pushed as `3de1b7b`. Reviewed the change; it is sound and I recommend merging. One finding worth acting on separately, detailed below.

### Conflict resolution

15 blocks across 8 files. Twelve were formatting-only — main had run `ruff format` over code this branch also touched (quoted vs bare annotations, list concat vs unpacking, wrapping, one docstring rewording). I took main's side so the formatter would not immediately undo the merge. One of those also dropped `_index_dim1`, which was defined and never called.

Three were genuine two-sided changes, both sides kept:

- **`__main__.py`** — this branch adds the MTP sidecar save; main replaced the runtime dispatch with `write_gguf_runtime_package` (dual onnx-genai/ort-genai, #543). The code *after* the conflict already referenced both, so dropping either would not have compiled.
- **`_group_query_attention.py` + test** — main added sliding-window support (`local_window_size` from the attention bias) and a `TestSlidingWindowSurvivesFusion` suite; this branch had only a reformat there.

**`inference_metadata_test.py` needed more than conflict resolution.** The branch's copy predates main's removal of `add_explicit_package_io` and referenced it seven times, so taking the branch side broke *collection* for six modules. I took main's file and re-applied this branch's `TestMtpSpeculatorMetadata` block (with its `_MtpBackboneConfig` fixture and `_seed_backbone_metadata` helper) plus the missing import.

### What I liked

- **Presence-driven, no flag.** `emit MTP iff the source ships the head` is the right shape — no CLI toggle to get wrong, and a source without `nextn` is byte-for-byte unchanged.
- **Dims derived, never hardcoded.** `derive_mtp_config` inherits everything from the backbone config, which is what Rule 2 asks for.
- **The auto-detect is actually tested both ways** — nextn GGUF emits the head, plain GGUF omits it. A one-sided test would not have caught a detector stuck on `true`.
- **The PR body states what was *not* run** (full 27B int4 export) rather than implying full coverage.

### Test results

- `src/mobius/integrations/gguf/` + `src/mobius/rewrite_rules/` — **475 passed, 1 skipped**.
- `inference_metadata_test.py` — 4 failures that **pre-exist on main** (reproduced by running main's own test file on main: same 4), plus one MTP schema case, below.

### Finding: the committed JSON schema is stale, not this branch

`TestMtpSpeculatorMetadata::test_matches_onnx_genai_json_schema` fails, but the emitted block is **not** wrong:

- The runtime's `SpeculatorConfig` (`generation.rs:65`) does use `proposal_type` / `num_speculative_tokens` — exactly what this branch writes.
- `parser.rs:26` documents "Every field is resolved from the `speculative` metadata section", so the key is right too.
- But `schema/inference_metadata.schema.json` still binds `speculative` to **`SpeculativeContract`** (`required: proposer, target, vocabulary, max_proposal_width`) — a different, later structure. The committed schema and the Rust types have diverged.

Also worth flagging for whoever owns the runtime side: the only place that parses this exact shape is a test helper named **`parse_legacy_speculator`**. If `speculative: SpeculatorConfig` is genuinely legacy, this branch is writing a format the runtime intends to retire, and that decision should be made deliberately rather than discovered later. I did not change the emitted format — that is a runtime-contract call, not a merge-conflict call.

Merging on the strength of the above; the schema divergence needs a separate fix on the onnx-genai side.

PR #533, opened 2026-08-21
Add SenseNova-U1.5-8B-MoT (NEO-unify) unified any-to-any support

justinchuby:
Adds Mobius support for [`sensenova/SenseNova-U1.5-8B-MoT`](https://huggingface.co/sensenova/SenseNova-U1.5-8B-MoT) (`model_type: neo_chat`, architecture `NEOChatModel`) — a **native unified any-to-any** model doing multimodal understanding, text-to-image generation and image editing from one backbone.

Pinned revision `1f6ec60423d29939dde4202fd82ae340b144e280`, Apache-2.0. **17,532,854,464 parameters** (≈50.2 GB) despite the `8B` in the name.

## Upstream source: a Hub packaging gap, resolved

The HF repo declares remote code it does not ship — `configuration_neo_chat.py` and `modeling_neo_chat.py` both **404** at the pinned revision *and* at `main`, so `trust_remote_code=True` cannot work against the Hub repo alone. The model card resolves it: the reference implementation is the `sensenova_u1` package at [OpenSenseNova/SenseNova-U1](https://github.com/OpenSenseNova/SenseNova-U1) (branch `feat/u1.5`), under `src/sensenova_u1/models/neo_unify/`. That package is the authority used here — nothing was inferred from names.

Two further packaging quirks, both verified benign:
- Shards `model-000{02,03,04}-of-00016.safetensors` are **absent**, but `model.safetensors.index.json` references no tensor in them. All 1116 tensors resolve; the checkpoint is complete. Numbering slack, not missing weights.
- No `preprocessor_config.json` and no `tokenizer.json` — image preprocessing exists only in code, and the tokenizer must be built from `vocab.json` + `merges.txt`. This has a direct consequence for metadata (below).

## Architecture, root-caused

**Mixture of Transformers.** All 42 Qwen3 layers carry two complete, disjoint weight sets. The understanding branch handles text and reference images; every module suffixed `_mot_gen` handles the noisy image tokens of the flow-matching sampler. There is no `lm_head_mot_gen` — the generation branch feeds the image head instead.

Routing is **per forward call**, not per token, and upstream `Qwen3Attention.forward` / `Qwen3DecoderLayer.forward` **raise `NotImplementedError`** for a mixed pass (upstream issue #207). Production always splits at token-type boundaries: an understanding pass writes a KV cache, then the generation pass reads that frozen cache once per denoise step. Exporting two decoder graphs is therefore the *only faithful* decomposition, not a convenience.

**Three rotary axes** inside `head_dim` 128 — temporal `[0:64)` at `rope_theta` 5e6, height `[64:96)` and width `[96:128)` at `rope_theta_hw` 1e4. QK-norm applies per *half* (`q_norm` over the 64 temporal dims, `q_norm_hw` over the 64 spatial dims), which is why the checkpoint's norm tensors are shape `[64]`, not `[128]`.

**Block-causal attention**: `allow[i][j] = (t[j] == t[i]) or (j <= i)`. All patches of one image share a temporal index, so they attend bidirectionally. For pure text every index is distinct and this degenerates to plain causal — which is exactly why text-only parity passes even with a naive causal mask and image parity does not.

**Vision tower has no transformer blocks at all** — `Conv2d(3,1024,k16,s16)` → GELU → *interleaved* 2-D RoPE → `Conv2d(1024,4096,k2,s2)`. One LLM token covers a 32×32 pixel tile.

**No VAE.** `use_pixel_head` drives a pixel-shuffle `ConvDecoder` (ONNX `DepthToSpace` `mode="CRD"`) straight to RGB. It predicts **x0**, and the sampler converts via `v = (x0 - z) / max(1 - t, t_eps)`.

Full write-up, including the upstream code paths that are dead for inference, in `docs/design/sensenova-u1-neo-unify.md`.

## What's added

`SenseNovaU1Config` (lifts text fields from the nested `llm_config`), a bare vision extractor hook, `SenseNovaU1Model`, `SenseNovaU1Task`, and registry wiring for `neo_chat`. Five components:

| Component | Contents |
|---|---|
| `embedding` | `embed_tokens` + reference-image scatter |
| `vision_encoder` | understanding-branch patchify tower |
| `decoder` | 42 MoT layers + `norm` + `lm_head` |
| `image_gen_embedding` | `vision_model_mot_gen` + timestep / noise-scale embedders |
| `image_gen_denoiser` | 42 `_mot_gen` layers + `norm_mot_gen` + pixel head |

`decoder` and `image_gen_denoiser` share the KV-cache layout, which is what lets the denoiser consume the prefill cache.

## Validation

**Weight alignment** — all **1116/1116** checkpoint tensors map onto ONNX initializers: 0 unmapped, 0 missing, 0 shape mismatches. The only initializers without a checkpoint source are derived RoPE/timestep tables.

**L1/L2** — 24 co-located tests: graph build for f32/f16/bf16, I/O contracts, config extraction, weight routing, and executable CPU checks including `batch=2` with ragged padding and a test that image tokens sharing a temporal index really do attend bidirectionally.

**L3 synthetic parity vs upstream PyTorch**, component by component (fp32, tiny config):

| Component | max_abs | notes |
|---|---|---|
| Vision tower | 7.5e-08 | grids 4×4, 6×2, 2×8 |
| Understanding decoder | 7.2e-07 | text-only and text+image t/h/w |
| Timestep embedder | 6.0e-08 | t = 0.0 / 0.37 / 1.0 |
| `fm_head` ConvDecoder | 6.6e-07 | square and non-square |
| Generation denoiser + head | 2.8e-07 | with a real KV prefix |

**L4/L5 real 17.5 B weights on an H200** — exported with `mobius.build(..., load_weights=True)` (35.16 GB fp16 + 31 GB fp32 generation branch) and run through onnxruntime CUDA.

| Modality | Result | Performance |
|---|---|---|
| Text | coherent multi-sentence output | TTFT **0.225 s**, **53.1 tok/s** |
| Text→image | 512×512 photoreal image | prefill 0.34 s, **7.7 steps/s** (3 passes/step incl. CFG) |
| Image edit | sky replaced, structure preserved | vision 0.19 s, **7.1 steps/s** |

Upstream PyTorch reference for comparison: 35.07 GB bf16, 35.41 GB CUDA peak, 2.52 steps/s.

**Stage-by-stage vs upstream, shared initial latent** (so every delta is attributable to the export, not sampling randomness):

| Stage | rel L2 | Pearson r |
|---|---|---|
| Prefill KV, layer 0 | 0.0036 | 0.999994 |
| Prefill KV, layer 41 | 0.0207 | 0.999786 |
| `image_embeds`, step 0 | 0.0037 | 0.999995 |
| x0 prediction, step 0 | 0.0165 | 0.999834 |
| x0 prediction, step 7 | 0.0306 | 0.998934 |
| **Final image** | **0.0338** | **0.998606** |

Final-image **PSNR 36.42 dB** — consistent with fp16-vs-bf16 drift, not architectural error. The two images are visually indistinguishable (below).

## Precision finding

The two branches must be exported at different precisions, matching how the checkpoint itself stores them:

- The understanding branch is BF16 upstream and exports cleanly to **float16**.
- The generation branch is stored **F32** and its activations exceed fp16 range. At float16 the flow-matching loop overflows and produces `NaN` part-way through sampling (reproducibly at step 15 of 20 at 512×512). It needs **float32**.

Mixing works end to end — the sampler casts the fp16 KV cache once after prefill.

## Known gaps (deliberate, documented, test-pinned)

**Canonical inference metadata is not emissible yet.** Export and execution are complete; the blockers are elsewhere and are separated precisely by `TestInferenceMetadataStatus` so the docs cannot drift:

1. *Image-processor contract* — the emitter dispatches on the vision graph's input signature. The NEO tower takes a plain `(1, 3, H, W)` image and the Hub ships no `preprocessor_config.json`, so no processor program binds and no `image_mean`/`image_std` are declared.
2. *No contract shape* — current emitters cover diffusion, audio codec, VLM and speech-to-text. A text decoder plus a flow-matching branch sharing one KV cache matches none; emitting plain VLM metadata would describe only half the model.
3. *Runtime* — ORT GenAI has no pipeline that alternates two weight branches over one KV cache with a sampler loop between them.

**GQA fusion does not fire** for either decoder graph on CUDA/fp16 (Mobius warns at export). Three independent guards each decline correctly: the block-causal bias contains an `Or`; the denoiser has no `attention_mask` input; the three-axis RoPE is not `op.RotaryEmbedding`. Correctness is unaffected; documented so it isn't misdiagnosed later.

## Waivers

The fast suite reports **23 pre-existing failures**, all `TypeError: save() got an unexpected keyword argument 'max_shard_size_bytes'`. These come from the installed `onnx_ir` 0.2.1 lacking that kwarg while `origin/main`'s `_model_package.py` already passes it. That file is untouched by this branch (`git diff --name-only origin/main...HEAD` confirms). Excluding them: **4218 passed**, ruff check and format clean.

## Test commands

```bash
python -m pytest src/mobius/models/sensenova_u1_test.py -q          # 24 passed
python -m pytest tests/build_graph_test.py tests/cli_test.py src/ -q \
  -k "not phi4mm and not apply_weights_unknown" -n auto             # 4218 passed
```

---

### Controlled comparison — identical initial latent, 8 steps, 512×512

Upstream PyTorch (bf16) vs Mobius ONNX (fp16 understanding + fp32 generation). Pearson **r = 0.998606**, **PSNR 36.42 dB**.

| Upstream reference | Mobius ONNX |
|---|---|
|  |  |

### Text-to-image, 20 steps, cfg 4.0

> *"A cinematic mountain lake at sunrise, realistic photography."*

Independently sampled (numpy vs torch RNG), so these are different draws of the same prompt — both photoreal, confirming the exported model is not merely close to the reference but genuinely generative.

| Upstream PyTorch — 2.52 steps/s | Mobius ONNX — 7.75 steps/s |
|---|---|
|  |  |

### Image editing through ONNX

Input: the upstream image above. Prompt: *"Change the sky to a dramatic starry night. Keep the mountains, lake and rocks unchanged."*

The reference image is encoded by the `vision_encoder` tower (32×32 patches → 256 tokens), scattered into the prompt by the `embedding` graph, and conditions the `_mot_gen` branch. Mountains, shoreline, reflection and foreground rocks are preserved while the sky is replaced — end-to-end proof that the understanding image path, `get_thw_indexes` position derivation and edit conditioning are all correct.

### Text decoding through ONNX

TTFT **0.225 s**, **53.06 tok/s**, fp16 on H200:

```
<think>
The user is asking me to describe the colours of a sunrise over a mountain lake.
This is a descriptive, creative writing task. I should write a vivid, sensory-rich
description that captures the beauty and atmosphere of this scene.

Let me plan this:

1. **Writing Type**: Creative descriptive prose / nature writing
```


justinchuby:
Canonical hashless metadata and executable generic shared-state workflow validation are now implemented in #540, with the runtime contract in justinchuby/onnx-genai#1700. #540 carries this branch's model implementation rebased onto current main and preserves the existing L4/L5 evidence; ordinary CI uses only deterministic tiny graphs.

justinchuby:
Superseded by #540, which rebased and merged the complete SenseNova U1.5 implementation and its existing L4/L5 evidence onto current `main`, then added the five-component hashless canonical producer metadata and tiny conformance package. I verified #540 contains every implementation/documentation path from this PR and adds the producer/runtime integration paths.

Post-merge corrections discovered by executing the real pinned 17.5B package through ONNX GenAI's canonical engine are in #545 (seeded noise, independent CFG prefix/KV symbols, forward x0 flow schedule, and bounded image output). Generic runtime support and synthetic execution are tracked in https://github.com/justinchuby/onnx-genai/pull/1700.

Published validated package: https://huggingface.co/justinchuby/sensenova-u1.5-8b-mot-onnx-canonical

Closing this PR to avoid maintaining two divergent implementations; its evidence remains preserved here and in the merged design documentation.

PR #543, opened 2026-08-22
feat(gguf): support Q6_K, fix byte-level tokenizers, emit loadable packages

justinchuby:
Three defects blocked running a Q4_K_M GGUF (`tencent/Hy-MT2-1.8B`) end to end.

## 1. Q6_K was unsupported

Q4_K_M is a **mixed** preset. This file is 192 × Q4_K, 33 × Q6_K, 129 × F32, and the Q6_K third is not a rounding error — it holds **34% of the bytes**, including the 193 MB `token_embd.weight` that this model also uses as its `lm_head` (there is no `output.weight`).

`repack_dequantized_tensor` already named Q6_K as an expected input for exactly this reason, but no unpacker existed, so the build failed outright.

Adds `_repack_q6_k`, following ggml's `dequantize_row_q6_K` interleave: each 128-element half consumes 64 `ql` bytes, 32 `qh` bytes and 8 scales, emitting four 32-element groups with `is = l // 16`.

**Verified bit-for-bit** against `gguf`'s own dequantizer on six real tensors, ~1M elements each — zero differing elements. This matters because every plausible-but-wrong permutation of that interleave still yields finite values of the right magnitude; only exact comparison distinguishes correct from subtly transposed.

## 2. Byte-level tokenizers were decoded as SentencePiece

The reconstructor applied Metaspace (`▁`) unconditionally. This model declares `tokenizer.ggml.model = gpt2`, whose vocabulary spells a leading space as `Ġ` — a different alphabet. Every inter-word space was dropped:

```
"Hello, world!"  ->  "Hello,!world!"
```

The code detected the bad round-trip, **warned, and emitted the broken tokenizer anyway**. Now it dispatches on the declared family.

Fixing the alphabet alone was not enough. `tokenizer.ggml.pre` names the *split* rule, which the byte-level model does not imply — llama.cpp keeps per-family regexes because families disagree on how digits, CJK and punctuation group. A mismatch **still round-trips**, so it is invisible to a round-trip check and only shows at the id level: a tab sample decomposed as `[184, 25056]` where the reference gives `[20061, 421]`.

Adds the family split table. All six samples — ASCII, CJK, emoji, repeated spaces, tabs/newlines — now match the HF tokenizer **id-for-id** as well as round-tripping exactly.

## 3. The Python API could not produce a loadable package

`mobius build-gguf --runtime onnx-genai` emitted tokenizer + inference metadata, but `build_from_gguf` callers got a bare graph that loads nowhere — the model directory was missing both artifacts.

Rather than document a second call sequence for API users, `write_gguf_runtime_package` now owns the full artifact set, and the CLI routes through it. One function knows what a complete package is, so the two paths cannot drift.

## Incidental

Q6_K was missing from `type_can_omit_zero_points`, which raised `KeyError` — caught by the existing suite. That test asserted Q6_K was *unsupported*, which is no longer true, so it is retargeted to Q5_K (still unsupported), preserving the "no silent float fallback" guarantee it was written to protect. A new test covers Q6_K selecting the asymmetric 4-bit target.

## Validation

`python -m pytest src/mobius/integrations/gguf/ -q` — **330 passed, 0 failed**.

End-to-end, the Python API now emits a complete package with no tokenizer warning:

```
  tokenizer: ...\tokenizer.json
  inference_metadata: ...\inference_metadata.yaml
  model.onnx.data   989.74 MB
```


src/mobius/integrations/gguf/_runtime_package.py
```diff
@@ -0,0 +1,73 @@
...
+
+def write_gguf_runtime_package(
+    pkg: Any,
+    gguf_path: str | Path,
+    output_dir: str | Path,
+    *,
+    save_model: bool = True,
+    **save_kwargs: Any,
+) -> dict[str, str]:
+    """Write a complete, loadable onnx-genai package for a GGUF-built model.
```
justinchuby: Make sure we support different backends

justinchuby:
Addressed all four review points in 7e39032.

**@justinchuby — support different backends.** `write_gguf_runtime_package` was hardwired to onnx-genai. It now takes `runtime="onnx-genai"|"ort-genai"` and dispatches to the matching writer; the CLI passes the flag through. I also removed the `build-gguf --runtime ort-genai` rejection, so mobius can emit for both today rather than waiting on the other side to land — an unknown runtime name still fails closed.

**@Copilot — misleading test name.** You were right, and this was the worst of the four: `test_dequantization_matches_gguf_reference_exactly` claimed bit-for-bit but asserted a tolerance bound *after* the lossy 4-bit requantization. Split the dequantization into `_dequantize_q6_k` so the test compares the dequantized floats against gguf's reference with `assert_array_equal` — which is what the name says, and what I had actually verified by hand on six real tensors. The requantization bound is now its own test under its own name.

**@Copilot — stale docstring.** Fixed; it now states that a failed round-trip is reported but not fatal, and why (the ids are correct by construction from the ggml ordering, and a byte-incomplete vocab cannot necessarily represent arbitrary probe text).

**@github-code-quality — CodeQL `__eq__`.** `_FakePackage` subclassed `dict` and added attributes without overriding `__eq__`. The emitter only calls `save()` and reads `config`, so the mapping behaviour was unused state — dropped the base class rather than adding an `__eq__` no caller needs.

Tests: **333 passed, 0 failed** (330 before; +3 covering the ort-genai path, the unknown-runtime guard, and the split requantization bound).

PR #548, opened 2026-08-22
Fix missing YaRN mscale softmax-scale correction in DeepSeek MLA

justinchuby:
## Root cause

DeepSeek-V2/V3's MLA attention applies **two independent** YaRN corrections in HuggingFace's reference implementation (`modeling_deepseek_v2.py`):

1. `attention_scaling` on the RoPE cos/sin cache — rescales only the *rope* portion of Q/K. Mobius's `YarnRope` already implements this correctly.
2. A separate softmax-scale multiplier applied inside `DeepseekV2Attention.__init__` (HF's `yarn_apply_mscale`): `scaling = qk_head_dim**-0.5 * mscale**2`, where `mscale = 0.1 * mscale_all_dim * log(factor) + 1.0`. This rescales the **entire** attention logit (rope + nope portions alike).

`src/mobius/components/_deepseek_mla.py` implemented (1) but not (2), on the mistaken assumption (captured in the removed code comment) that (1) alone reproduces the needed effect. That's false whenever `mscale == mscale_all_dim` (as in `deepseek-ai/DeepSeek-V2-Lite`, both 0.707): (1)'s ratio-based `attention_factor` collapses to exactly `1.0` (a no-op), while (2) is still a real ~1.59x correction that HF applies. `deepseek-ai/DeepSeek-V3`'s official config (`mscale=mscale_all_dim=1.0`) has the same gap (missing ~1.87x correction).

### Symptom

`DeepSeekMLA` reproduced HF logits almost exactly at **position 0** (a single-key softmax is scale-invariant to any positive multiplier) but diverged by up to **~0.9 absolute / 93% of elements at position ≥ 1**, for both `DeepSeek-V2-Lite` and a real `DeepSeek-V3` config — reproducible even in a single **dense** (non-MoE) layer, which isolated the bug to attention/RoPE scaling rather than MoE routing.

## Fix

Added `yarn_apply_mscale()` in `src/mobius/components/_rotary_embedding.py`, mirroring HF's helper of the same name exactly (same guards: no-op when `rope_type == "default"`, no-op when `mscale_all_dim` is falsy). `DeepSeekMLA.__init__` now calls it to compute `self.scaling` whenever no explicit `scale=` override is passed.

## Before / after evidence

Using real HF configs (random weights, single dense MLA layer / full 4-layer model):

| Config | Before (max abs diff, position ≥ 1) | After |
|---|---|---|
| `deepseek-ai/DeepSeek-V2-Lite` (factor=40, mscale=mscale_all_dim=0.707) | up to ~0.9 | ~5e-6 (FP noise) |
| `deepseek-ai/DeepSeek-V3` config (factor=40, mscale=mscale_all_dim=1.0) | missing ~1.87x scale (confirmed via config inspection) | corrected |

## Tests

- `tests/integration_test.py::test_deepseek_v2_lite_prefill_logits_match` and `::test_deepseek_non_mla_decoder_prefill_logits_match` — **both previously FAILED** with large divergence at prefill positions ≥ 1; **now PASS** at `rtol=atol=1e-3`.
- `src/mobius/components/_deepseek_mla_test.py` — 4 new unit tests: exact `mscale^2` correction value (mirroring `DeepSeek-V2-Lite`'s real rope_scaling), the `rope_type == "default"` no-op guard, the `mscale_all_dim == 0` no-op guard, and that an explicit `scale=` override still wins.
- `tests/synthetic_parity_test.py`: removed the `deepseek_v3` atol override of `0.04` (previously attributed to "MoE dispatch FP accumulation" — actually this same bug being masked) and replaced it with an empirically measured `0.0025`. Residual FP-order noise from parallel MoE dispatch measured at ~3.5e-5 in a bare interpreter and a reproducible ~1.2e-3 under this repo's pytest process (ORT intra-op thread-schedule dependent, not seed-dependent); `0.0025` keeps ~2x headroom over the higher measurement while being 16x tighter than the old `0.04`.

### Commands run

```
python3 -m pytest src/mobius/components/_deepseek_mla_test.py src/mobius/components/_rotary_embedding_test.py -q
python3 -m pytest tests/integration_test.py -k "deepseek_v2_lite_prefill_logits_match or deepseek_non_mla_decoder_prefill_logits_match" -q
python3 -m pytest "tests/synthetic_parity_test.py::test_synthetic_parity" -k "deepseek or dots1 or glm4" -q
python3 -m pytest tests/synthetic_parity_test.py -q   # full suite: only pre-existing, unrelated `modernbert-decoder` failure (confirmed present on origin/main too)
ruff check src/mobius/components/_deepseek_mla.py src/mobius/components/_deepseek_mla_test.py src/mobius/components/_rotary_embedding.py tests/synthetic_parity_test.py
ruff format --check <same files>
```

All pass (25/25 MLA unit tests incl. 4 new, 63/63 combined MLA+rotary unit tests, 2/2 target integration tests, full synthetic-parity suite green aside from the pre-existing unrelated `modernbert-decoder` failure).

## Out of scope (not included in this PR)

- GLM-5.2 (`glm_moe_dsa`) registration/export — separate, unrelated follow-up (tracked via PR #404).
- The `deepseek_v4` synthetic-parity config-override bug in `tests/_test_configs.py` (`Weight shape mismatch for model.layers.0.self_attn.q_b_proj.weight`) — separate, unrelated fast-follow.

---
Working as Sapper (Systems Dev for Models & Preprocess).

### DIFF
--- a/src/mobius/components/_deepseek_mla.py
+++ b/src/mobius/components/_deepseek_mla.py
@@ -20,6 +20,7 @@ from mobius.components._common import Linear
 from mobius.components._rms_norm import RMSNorm
 from mobius.components._rotary_embedding import (
     apply_rotary_pos_emb,
+    yarn_apply_mscale,
 )
 
 if TYPE_CHECKING:
@@ -94,12 +95,18 @@ class DeepSeekMLA(nn.Module):
             bias=False,
         )
 
-        # Attention scale: 1/sqrt(qk_head_dim).
-        # YaRN mscale is handled by the cos/sin cache (attention_factor
-        # is applied to cos/sin in YarnRope), so no additional scaling
-        # is needed here. The cos/sin scaling produces attention_factor^2
-        # effect on the attention logits automatically.
-        self.scaling = scale if scale is not None else self.qk_head_dim**-0.5
+        # Attention scale: 1/sqrt(qk_head_dim), with an additional YaRN
+        # mscale^2 softmax-scale correction when YaRN is active (see
+        # ``yarn_apply_mscale``). This is separate from the RoPE cos/sin
+        # ``attention_factor`` applied inside YarnRope, which only rescales
+        # the rope portion of Q/K, not the full attention logit.
+        self.scaling = (
+            scale
+            if scale is not None
+            else yarn_apply_mscale(
+                config.rope_type, config.rope_scaling, self.qk_head_dim**-0.5
+            )
+        )
 
         # Whether RoPE uses interleaved layout (DeepSeek-V3 uses this)
         self._rope_interleave = config.rope_interleave
--- a/src/mobius/components/_deepseek_mla_test.py
+++ b/src/mobius/components/_deepseek_mla_test.py
@@ -5,6 +5,8 @@
 
 from __future__ import annotations
 
+import math
+
 import pytest
 
 from mobius._testing import (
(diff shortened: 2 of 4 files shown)


justinchuby:
## Independent Lead review — VERDICT: **APPROVE**

Reviewed read-only (no local edits/pushes): full diff vs `origin/main` (via a throwaway git worktree, since discarded), HF `transformers` `modeling_deepseek_v2.py` reference (fetched live), and independent re-execution of the test suite pre- and post-patch.

### Scope
Diff against `origin/main` is exactly the 4 files reported by the PR (`+115/-10`): `_deepseek_mla.py`, `_rotary_embedding.py`, `_deepseek_mla_test.py`, `tests/synthetic_parity_test.py`. No unrelated changes.

### Correctness vs HF reference
HF's `yarn_get_mscale`/`yarn_apply_mscale` + `DeepseekV2Attention.__init__`'s `self.scaling = yarn_apply_mscale(...)` were compared line-by-line against the new `yarn_apply_mscale(rope_type, rope_scaling, scaling)` in `_rotary_embedding.py`. Formula and all three guards (`rope_type == "default"` no-op, `mscale_all_dim` falsy no-op, `factor <= 1` no-op) match exactly in intent and numerics. Confirms the PR's core claim: the softmax-scale `mscale²` correction is independent of, and in addition to, `YarnRope`'s existing `attention_factor` on cos/sin.

One minor, non-blocking divergence: HF reads `rope_parameters["factor"]` unconditionally (crashes `KeyError` on a malformed config), Mobius does `.get("factor") or 1.0` (fails gracefully). Not reachable for any real DeepSeek checkpoint — defensible.

### API surface — `config.rope_type` never `None` for real DeepSeek configs
Traced `_configs/_base.py`'s `RoPEConfig` construction: `rope_type` always resolves through `_first_not_none(..., default="default")`, so it is only ever the *field* `None` when the whole `RoPEConfig` is `None` (model has no RoPE at all — not the DeepSeek case, since DeepSeek always sets `rope_theta`). `yarn_apply_mscale`'s guard (`rope_type == "default" or not rope_scaling`) is also safe even in a hypothetical `None` state, since `rope_scaling` is sourced from the same object. Confirmed further by `deepseek_v4_flash_test.py` asserting `config.rope_type == "yarn"` on a real-shaped config. No live bug here.

### Independent re-execution (this is the substantive verification)
- Checked out the PR head via a temporary worktree, ran `pytest src/mobius/components/_deepseek_mla_test.py`: **25/25 pass** post-patch (6 repeat runs, no flakiness).
- Reverted only the two source files to `origin/main` (keeping the new tests): `test_yarn_mscale_softmax_scale_correction` **fails** (`0.25` vs expected `0.397...`), the other 3 new guard tests pass either way (expected — they test no-op paths). This directly and independently confirms the PR's fail-before/pass-after claim.
- Ran `tests/synthetic_parity_test.py -k deepseek_v3` 8 times: **deterministic** `max_abs_diff=0.001224` every run (fully seeded, not flaky) — matches the PR's own cited "~1.2e-3 under pytest" almost exactly. `0.0025` gives ~2.04x headroom over this measured, reproducible number, and is 16x tighter than the old `0.04`. Did **not** independently reproduce the cited "~3.5e-5 bare interpreter" figure (different harness), but the more conservative pytest-measured number alone already justifies the chosen tolerance — verdict: evidence-backed, not arbitrary.
- Confirmed the synthetic `deepseek_v3` test config genuinely exercises the YaRN path (`rope_type="yarn"`, `mscale_all_dim=1.0`, `factor=4.0` — a real ~30% scale change, not a near-1 no-op).

### CI / numerical risk
`ruff check` and `ruff format --check` both clean on all 4 changed files. `dict | None` syntax is fine under `requires-python = ">=3.10"` with `from __future__ import annotations` present. `self.scaling` is a scalar Python float (not a tensor op) — no dtype/broadcast risk.

### Conclusion
No high-confidence blocking issues found. Fix is correct, minimal, well-tested, and its two headline claims (fail-before/pass-after, tolerance justification) were independently reproduced rather than taken on faith. **APPROVE.**

PR #555, opened 2026-08-23
Add GLM-5.2 (glm_moe_dsa) model support with DSA/IndexShare + full-attention fallback

justinchuby:
## Summary

Registers the `glm_moe_dsa` model_type — **`zai-org/GLM-5.2`**, architecture `GlmMoeDsaForCausalLM` (78 layers, hidden 6144, MLA+DSA, 256 routed + 1 shared expert, top-8, sigmoid/noaux_tc routing, `routed_scaling_factor=2.5`, 1 MTP layer, default RoPE) — by **reusing existing DeepSeek MLA/MoE/QMoE-fusion machinery** rather than merging draft PR #404 or sibling branch `feat/glm-dsa-indexshare` wholesale.

Both stale branches were diffed and classified hunk-by-hunk against current `main`. Several correctness bugs present in the sibling branch were identified and **not** carried over — fixed from scratch instead:
1. Indexer RoPE split only rotating `qk_rope_head_dim` of the wider `index_head_dim` (sibling rotated the whole head).
2. A 4D-query × 3D-key `MatMul` broadcast that silently mispairs axes whenever `batch_size != seq_len` (sibling omitted an `unsqueeze`).
3. (Found during independent review of *this* PR, fixed here) `IndexShare`'s query/key/value inputs must be cast to f32 per the frozen op schema (`docs/architecture/INDEXSHARE_DESIGN.md` in onnx-genai) regardless of the model's compute dtype — the real `zai-org/GLM-5.2` checkpoint is bfloat16.

## What's included

- **`src/mobius/models/glm_moe_dsa.py`** (new): `GlmMoeDsaIndexer` (DSA top-k scoring/selection), `GlmMoeDsaAttention` (subclasses `DeepSeekMLA`, exports the frozen `pkg.nxrt::IndexShare` op per the "full"/"shared" indexer schedule), `GlmMoeDsaDecoderLayer`, `GlmMoeDsaTextModel`, `GlmMoeDsaCausalLMModel` (subclasses `DeepSeekV3CausalLMModel`). Existing DeepSeek MLA math, MoE gate/FFN, and QMoE fusion (#550) are reused unchanged.
- **`config.use_dsa=False`** (new **`--glm-full-attention`** CLI feature) switches to plain dense MLA attention, executable on stock ONNX Runtime today, as an explicit fallback until the native `IndexShare` runtime kernel lands in onnx-genai. Wired through both the `--config` (local dir) and `--model` (HF id, including the diffusers-dispatch fallback branch) build paths, validated to only apply to `glm_moe_dsa`.
- **`src/mobius/_configs/_base.py`**: 6 new `ArchitectureConfig` fields for DSA (`index_topk`, `index_head_dim`, `index_n_heads`, `index_topk_freq`, `index_skip_topk_offset`, `use_dsa`) and HF-config extraction, including a `model_type == "glm_moe_dsa"`-conditional default so `scoring_func`/`topk_method` correctly resolve to GLM's sigmoid/noaux_tc routing. (The real `GlmMoeDsaConfig` does not declare these as dataclass fields at all when omitted from `config.json` — they're silently absent, not defaulted — so without this fix GLM-5.2 would silently inherit DeepSeek's softmax/greedy routing defaults.)
- **`src/mobius/_registry.py`**, **`src/mobius/models/__init__.py`**: registration wiring.
- **`src/mobius/models/glm_moe_dsa_test.py`** (new, 21 tests): registry mapping; indexer schedule formula (cross-checked against the real installed `transformers.models.glm_moe_dsa` config's own `.indexer_types` derivation, not just a hand-written expectation); `scoring_func`/`topk_method` regression test; graph-build structural tests for both the DSA and full-attention-fallback paths (`IndexShare`/`Attention` op counts, indexer-weight presence, QMoE-fusion node count, **and a regression test asserting `IndexShare`'s query/key/value inputs are float32 under a non-float compute dtype** — verified to fail without the fix above and pass with it); `preprocess_weights` tests (MTP-layer and indexer-weight dropping); two numeric-parity tests against the real HF reference implementation (raw index-score parity to ~1e-6, and a tie-tolerant "true top-k + strictly increasing indices" property test — selected-index-*set* equality is deliberately not asserted because `torch.topk` and ONNX `TopK` have different, individually valid tie-breaking conventions at degenerate/tied score boundaries; this was verified empirically, not assumed).
- **`src/mobius/integrations/transformers/_builder_test.py`**: tests for the new `glm_full_attention` override/validation on both the main and diffusers-dispatch code paths.

## Explicitly out of scope (no open tracking issue found — searched `onnxruntime/mobius` issues for GLM/MTP/DSA/IndexShare/GGUF, none exist)

- **MTP (multi-token-prediction) layer.** The real `transformers.models.glm_moe_dsa` reference implementation itself has no MTP module — `preprocess_weights` drops MTP-layer weights with a warning, matching upstream's own `_keys_to_ignore_on_load_unexpected` pattern. Base-model logits never accidentally include an unsupported MTP path. Exporting MTP is a follow-up.
- GGUF import support.
- CUDA residency policy / q* / scheduler work.

## Local evidence

- `ruff check` / `ruff format --check`: clean (pinned lintrunner version `0.16.2`).
- `pytest src/mobius/models/ -q` (full directory, excluding `arch_validation`/`golden`/`generation`/`benchmark` markers): **523 passed, 4 skipped**.
- `pytest src/mobius/models/glm_moe_dsa_test.py src/mobius/models/deepseek_test.py src/mobius/models/deepseek_v4_test.py src/mobius/_registry_test.py src/mobius/integrations/transformers/ src/mobius/_configs/ -q`: **146 passed**.
- Independent review (via a separate agent pass over the staged diff) found and this PR fixes: the `IndexShare` f32-input-cast bug above (blocker) and a missing `glm_full_attention` validation gap on the diffusers-dispatch branch (should-fix, also fixed + tested here).

## Remaining runtime/MTP/full-model blockers & next onnx-genai gate

- **Native `IndexShare` runtime kernel**: this PR only emits the frozen `pkg.nxrt::IndexShare` op into the graph; actually generating tokens end-to-end for the DSA path requires the corresponding onnx-genai/ORT custom-op kernel to exist and execute it. That is the concrete next onnx-genai-side gate — without it, only the `--glm-full-attention` dense fallback (stock-ORT-executable) can currently run tokens today.
- **MTP layer**: tracked as a documented follow-up above, no export path yet.
- **GGUF import**: not attempted in this slice.

### DIFF
--- a/src/mobius/_configs/_base.py
+++ b/src/mobius/_configs/_base.py
@@ -495,6 +495,30 @@ class ArchitectureConfig(BaseModelConfig):
     swiglu_limit: float = 0.0
     num_nextn_predict_layers: int = 0
 
+    # GLM-5.2 (``glm_moe_dsa``) DeepSeek Sparse Attention (DSA).
+    #
+    # ``use_dsa`` is not an upstream HF field -- it is a Mobius export-time
+    # toggle (default True) flipped to False by ``--glm-full-attention`` /
+    # ``config_overrides={"use_dsa": False}`` to fall back to plain dense MLA
+    # (reusing ``DeepSeekV3TextModel`` unchanged) on runtimes that cannot yet
+    # execute ``pkg.nxrt::IndexShare``.
+    use_dsa: bool = True
+    # Per-layer indexer schedule ("full" runs the indexer; "shared" reuses the
+    # top-k selection from the closest preceding "full" layer). When the
+    # checkpoint config omits this list, it is derived from
+    # ``index_topk_freq`` / ``index_skip_topk_offset`` -- see
+    # ``mobius.models.glm_moe_dsa._indexer_types``, which mirrors HF
+    # ``GlmMoeDsaConfig.__post_init__`` exactly.
+    indexer_types: list[str] | None = None
+    index_topk_freq: int | None = None
+    index_skip_topk_offset: int | None = None
+    indexer_rope_interleave: bool = False
+    # Whether the (currently unexported, see ``glm_moe_dsa.py``) MTP layer is
+    # meant to reuse the target's shared top-k indices rather than running its
+    # own indexer pass. Recorded for round-tripping/documentation; not yet
+    # acted on since MTP export itself is out of scope this cycle.
+    index_share_for_mtp_iteration: bool = False
+
     # Vision shared fields (accessed as top-level config.X by tasks)
     mm_tokens_per_image: int | None = None
     image_token_id: int | None = None
@@ -897,8 +921,20 @@ class ArchitectureConfig(BaseModelConfig):
             n_group=getattr(config, "n_group", 1),
             topk_group=getattr(config, "topk_group", 1),
             routed_scaling_factor=getattr(config, "routed_scaling_factor", 1.0),
-            scoring_func=getattr(config, "scoring_func", "softmax"),
-            topk_method=getattr(config, "topk_method", "greedy"),
+            # ``scoring_func``/``topk_method`` are not real HF dataclass
+            # fields on ``GlmMoeDsaConfig`` -- transformers hardcodes
+            # sigmoid + e_score_correction_bias + (degenerate, n_group=1)
+            # top-k routing for GLM-5.2 rather than exposing it as a config
+            # knob. A checkpoint that omits these keys entirely (unlike the
+            # real zai-org/GLM-5.2 config.json, which sets them explicitly)
+            # would otherwise silently default to the wrong (softmax/greedy)
+            # DeepSeek-V2 routing.
+            scoring_func=getattr(
+                config, "scoring_func", "sigmoid" if model_type == "glm_moe_dsa" else "softmax"
+            ),
+            topk_method=getattr(
+                config, "topk_method", "noaux_tc" if model_type == "glm_moe_dsa" else "greedy"
+            ),
             first_k_dense_replace=getattr(config, "first_k_dense_replace", 0),
             n_shared_experts=getattr(config, "n_shared_experts", None),
             # Multi-head Latent Attention (MLA)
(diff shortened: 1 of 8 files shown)


src/mobius/_configs/_base.py
```diff
@@ -495,6 +495,30 @@ class ArchitectureConfig(BaseModelConfig):
...
     num_nextn_predict_layers: int = 0
 
+    # GLM-5.2 (``glm_moe_dsa``) DeepSeek Sparse Attention (DSA).
+    #
+    # ``use_dsa`` is not an upstream HF field -- it is a Mobius export-time
+    # toggle (default True) flipped to False by ``--glm-full-attention`` /
+    # ``config_overrides={"use_dsa": False}`` to fall back to plain dense MLA
+    # (reusing ``DeepSeekV3TextModel`` unchanged) on runtimes that cannot yet
+    # execute ``pkg.nxrt::IndexShare``.
+    use_dsa: bool = True
```
justinchuby, line 505: Can you move to model specific config files?

src/mobius/integrations/transformers/_builder.py
```diff
@@ -158,11 +158,18 @@ def build_transformers_model(
...
     kv_cache_scales: dict[int, tuple[float, float]] | None = None,
     prune_prefill_prefix: bool = False,
+    glm_full_attention: bool = False,
 ) -> ModelPackage:
     """Build a model package from a Transformers checkpoint.
 
     If the repository is not a supported Transformers checkpoint, dispatch to
     the Diffusers integration so :func:`mobius.build` remains ecosystem-agnostic.
+
+    ``glm_full_attention`` is the ``--glm-full-attention`` CLI feature: it
```
justinchuby, line 168: Generalize this so that it is not model specific

PR #556, opened 2026-08-23
Derive workflow input admission instead of leaving it to the consumer

justinchuby:
## The failure

The ONNX GenAI producer-conformance job rejects every shared-state pixel-flow request:

```
Error: required workflow package input 'request.latent' was not supplied
```

failing `mobius_shared_state_pixel_flow_executes_reference_image_edit` and
`mobius_shared_state_pixel_flow_executes_text_and_generation_paths`. Every other
workflow in the suite passes, which is why this reads as a pixel-flow problem.

Signal failure: onnx-genai#1785.

## Root cause

`request.latent` is not a precondition for that workflow. The package derives its own
starting noise from the generic `seed`/`width`/`height` request roles and treats a
caller-supplied latent as an *override*, gated on `request.latent_present`. None of the
three paths it serves — text understanding, text-to-image, reference-image edit — needs
the caller to attach one.

The runtime demanded it anyway because the producer never said otherwise. Mobius
published `required` as an optional, hand-written key, so a builder with nothing to say
about admission wrote nothing — and `WorkflowInput::required` in `onnx-genai-metadata` is
`#[serde(default = "default_true")]`. Silence did not mean "optional"; it meant
"mandatory on every path", the strongest possible reading of an unstated contract,
applied to a value the package computes for itself.

That is a property of how workflow metadata is emitted, not of this model. The pixel flow
was just the first workflow whose optional attachment a conformance test actually
reached. Auditing the other 14 packages on `main` found the same defect still live:

| package | input | published |
| --- | --- | --- |
| `adapter` | 5 selection inputs | no `required` key — required by omission |
| `shared_state_pixel_flow` | `request.prompt_tokens`, `request.negative_prompt_tokens` | no `required` key |
| `decoder`, `static_cache` | `package.eos_ids` | `required: true` **and** a `literal` source **and** a `default` |

That last row is two contracts that cannot both hold.

## The fix

Admission is now **derived from the published program** at `_publish_workflow_v1` — the
single point all 15 workflow builders route through — alongside the existing
`declare_request_alignment` pass it mirrors:

- an input the package can satisfy itself (a `default`, or a package-owned `literal`
  source) is not a caller obligation;
- an input whose absence the program *observes*, through a `present_as` symbol the steps
  actually branch on, is by construction executable without it;
- everything else is externally required and now says so **explicitly**, so no reader has
  to supply a default for a key that isn't there.

Both contradictions **fail closed** rather than being silently resolved. A declaration
claiming an escape *and* `required: true` is rejected — and so is the mirror case, an
input relaxed to optional that the workflow has no default, package source or presence
gate for. That second check refuses the tempting non-fix for this very bug: flipping
`required` without giving the workflow a way to proceed doesn't make a request
serviceable, it moves a clean admission rejection to an unbound value part-way through
execution.

The pass reads only structural facts already in the document. It names no model, class or
checkpoint constant; model-specific values stay in the builders and config adapters.
Reachability is deliberately path-insensitive — it answers "does this workflow ever read
this value", which is what an admission decision needs. A producer that guessed at which
branch a given request takes would be describing one caller rather than its own contract.

### Designs rejected

- **"branch-only ⇒ not required"** is unsound: `video`'s `request.encoder_hidden_states`
  and `speculative`'s `request.grammar_transition_table` are legitimately caller-required
  yet only consumed inside a branch.
- **"`externally_suppliable: true` ⇒ not required"** is not an invariant here:
  `diffusion`'s `request.noise` is both.
- **"unused in `steps` ⇒ not required"** is wrong: `adapter`'s selection inputs are read
  from `metadata["adapters"]["selection"]`, outside the workflow program.

No `model.io` compatibility is introduced; everything stays on serialized canonical
`pipeline.workflow`.

## Tests

- `TestInputAdmissionIsDerivedNotDefaulted` — the derivation directly, including both
  fail-closed directions and a `present_as` gate no step reads.
- `canonical_workflow_contract_test.py` — the invariant across **every** checked-in
  fixture package: no input leaves its requirement to the reader, and a required input is
  one the package cannot supply itself.
- `sensenova_u1_test.py` — the pixel-flow contract **per path**: the externally required
  set is exactly `{request.prompt_tokens, request.negative_prompt_tokens}`, latent and
  image are presence-gated with live gates, and text, text-to-image and reference-image
  edit are each admissible from their own attachments.

## Validation

Reproduced and fixed against the real ONNX GenAI conformance suite, not a metadata shape
assertion.

| check | result |
| --- | --- |
| `generate_onnx_genai_validation_packages.py` (15 packages) | ok |
| `compare_onnx_genai_validation_packages.py` vs fixtures | ok |
| `validate_metadata` over all 15 packages | ALL PACKAGES VALID |
| `cargo test -p onnx-genai-engine --test onnx_genai_workflow_conformance` with `MOBIUS_WORKFLOW_CONFORMANCE_REQUIRED=1` | **15 passed, 0 failed** (129s) |
| `pytest tests/build_graph_test.py tests/cli_test.py tests/canonical_workflow_contract_test.py src/` | **4830 passed, 56 skipped** |
| `ruff check` / `ruff format` on changed files | clean |

Attributable before/after, same suite and same runtime: against the revision the
conformance job pins today, both pixel-flow tests still fail with the exact error above
(`0 passed; 2 failed`); on this branch all 15 pass.

The fixture diff is deliberately small — 7 inputs gain an explicit `required: true` that
was previously implied, and `package.eos_ids` drops the contradictory `required: true` in
two packages.

## Dependencies

**None on onnx-genai#1723.** That PR is the native workflow backend, not a schema change.
The schema offers no conditional-requiredness construct, so `required: false` +
`present_as` + a branch is the correct and only way to express this contract. The fix is
entirely producer-side.

## Out of scope

`speculative` and `video` declare branch-consumed inputs
(`request.grammar_transition_table`, `request.draft_ms`, `request.target_ms`,
`request.encoder_hidden_states`) as genuinely required. Because reachability is
path-insensitive those are untouched; downgrading them would trade an admission rejection
for a late unbound-value failure. Whether those contracts are right is a separate
question about those workflows.


justinchuby:
Pushed `e986c500`, addressing a Claude Opus 5 review pass. Three latent soundness gaps in the derivation, none of which changed any generated metadata — all 15 packages still regenerate byte-identical:

1. **`published_value_references` missed two read positions outside `steps`.** A bounded or growing state cell reads its own `recurrence.max` / `recurrence.increment`, and a state-service group reads its `update.capacity`. Six checked-in packages name declared workflow inputs in exactly those slots (`decoder`→`package.max_context`, `static_cache`→`package.cache_capacity`, and four more). Harmless today because the set is only consulted for `present_as` symbols, which are scalar bools and cannot occupy an extent — but the helper is exported, and the next caller would have inherited the hole. The `state_service` comment claimed it "names ports rather than values", which was wrong for `update.capacity`; it now says what it means.

2. **The exhaustiveness test pinned the walker instead of the schema.** It asserted set equality against a workflow containing neither position, so it passed with both misses in place. It now carries a bounded recurrence and an update capacity, and separately asserts that port names and the `kind` discriminator stay *out*.

3. **A `literal` source is only an escape together with a `default`.** The consumer binds a literal-sourced input from `input.default` and nothing else, so one without a default binds no value at all — calling that optional would admit a request and then fail part-way through on an unbound value, which is precisely what this pass exists to prevent. Now fails closed with its own message. No package declares that shape today.

Re-validated after the change: 15/15 packages regenerate and compare clean, `validate_metadata` accepts all 15, `onnx_genai_workflow_conformance` passes **15/15** with `MOBIUS_WORKFLOW_CONFORMANCE_REQUIRED=1`, and the Python suite is 4831 passed / 56 skipped with `ruff` clean.

PR #561, opened 2026-08-23
Disable ONNX Runtime telemetry repo-wide

justinchuby:
Repo rule: ONNX Runtime does not phone home from this repository's test runs or CI.

## Why the rootdir conftest

ORT reads `ORT_DISABLE_TELEMETRY` when its native library loads, so it must be in the environment **before** anything imports `onnxruntime`. A fixture — even session-scoped and autouse — is too late, because test modules import ORT at module scope. The rootdir `conftest.py` is imported before those. 39 files here import `onnxruntime`, so no narrower hook would cover them.

## Where it's set

| Where | Covers |
| --- | --- |
| `conftest.py` (rootdir) | every pytest run |
| `.github/workflows/*.yml` | all 10 workflows, at workflow level so every job and step inherits it — including steps that reach ORT without going through pytest |

`arch_diff.yml` and `benchmark.yml` already had a top-level `env:`, so the key was merged into those rather than adding a second block. Both mechanisms assign only if unset, so an explicit `ORT_DISABLE_TELEMETRY=0` still wins for deliberate debugging.

## Deliberately not set inside the package

mobius is a library. Mutating a process-wide environment variable at import would silently change ORT's behaviour for anything that depends on us. The rule covers this repo's own entry points; it isn't ours to impose on downstream callers.

## Verified rather than assumed

`ORT_DISABLE_TELEMETRY` is present in the 1.29.0 `libonnxruntime` and pybind binaries. Worth checking first — a variable ORT ignored would look like a fix while changing nothing, which is worse than no fix at all.

## Guard

`tests/telemetry_policy_test.py` covers both mechanisms, because config duplicated across eleven files with no shared code is exactly what rots. It also asserts the workflow glob is non-empty, so the parametrized case can't silently pass by matching nothing.

I confirmed the guards aren't vacuous by reverting each mechanism in turn — 2 of the 12 cases fail, one per mechanism.

## Context

This came out of an unrelated debugging session on the MLX EP, where ORT's 1DS telemetry HTTP worker was seen locking an already-destroyed mutex during interpreter teardown and aborting the process (`Abort trap: 6`) *after* all tests had passed — a background-thread throw that is never caught, so it takes the run down and reds CI on an unrelated change. That was on macOS/arm64 and isn't necessarily reachable here, so this PR is scoped to the privacy rule rather than claiming a stability fix.

## Testing

`733 passed / 218 skipped` for `tests/telemetry_policy_test.py` + `tests/model_coverage_test.py`. `lintrunner` clean.

### DIFF
--- a/.github/workflows/arch_diff.yml
+++ b/.github/workflows/arch_diff.yml
@@ -23,6 +23,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
+  ORT_DISABLE_TELEMETRY: "1"
   BASE_REF: ${{ github.event.pull_request.base.sha || github.event.inputs.base_ref || 'origin/main' }}
   HEAD_REF: ${{ github.event.pull_request.head.sha || github.event.inputs.head_ref || 'HEAD' }}
 
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -24,6 +24,7 @@ concurrency:
 # For PRs: uses the PR base and head SHAs for exact reproducibility.
 # For workflow_dispatch: uses the input values or falls back to main.
 env:
+  ORT_DISABLE_TELEMETRY: "1"
   BASE_REF: ${{ github.event.pull_request.base.sha || github.event.inputs.base_ref || 'main' }}
   HEAD_REF: ${{ github.event.pull_request.head.sha || github.event.inputs.head_ref || 'main' }}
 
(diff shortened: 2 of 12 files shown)


justinchuby:
Closing — mobius is a Microsoft product, so ORT usage telemetry is fine here and there's nothing to opt out of. The rule I was applying came from onnxruntime-mlx (a personal repo), where it's both a privacy choice and a workaround for ORT's 1DS telemetry thread aborting at interpreter teardown. Neither rationale transfers.

Landed there instead: justinchuby/onnxruntime-mlx#45.

PR #589, opened 2026-08-24
Add Moonshine Streaming (moonshine_streaming) speech-to-text support

justinchuby:
Adds complete support for the HuggingFace architecture `moonshine_streaming`, validated end-to-end against `moonshine-ai/moonshine-streaming-tiny` at pinned revision `f8e9dfd8c562c257c151a907b7b7f2fe8ff8511a`. The revision is threaded through config extraction, the processor, weight download, golden generation, the L4/L5 test harness and the CLI.

## What's different from offline Moonshine

Moonshine Streaming is not a variant of Moonshine's front end — it replaces it and changes the encoder's positional scheme. All of the following are reproduced exactly:

| Stage | Behaviour |
|---|---|
| **Framing front end** | Waveform reshaped into fixed 5 ms frames (80 samples @ 16 kHz), per-frame CMVN, learned `asinh(exp(log_k) * x)` compression, bias-free framing linear + SiLU |
| **Frame mask** | `padding_mask.sum(-1) // frame_len`, applied multiplicatively |
| **Causal downsampling** | Two **left-padded** stride-2 convolutions (kernel 5, `pads=[4, 0]`), so no future frame leaks backwards. The mask is pushed through the same receptive field with an "any valid" reduction, matching upstream's `conv1d(mask, ones, stride) > 0` |
| **No encoder RoPE** | Encoder self-attention is purely content based |
| **Per-layer asymmetric windows** | `[(16,4), (16,4), (16,0), (16,0), (16,4), (16,4)]`. `right` is the strict lookahead in encoder frames; `right == 0` is fully causal. Modelled per layer, not as one uniform window, because this is what bounds streaming latency |
| **Unit-offset LayerNorm** | Affine-free LayerNorm scaled by `gamma + 1`; the checkpoint stores `gamma`, not `weight` |
| **Context adapter** | Decoder adds a learned absolute `pos_emb` to the encoder output, then projects to the decoder width when they differ |
| **Decoder** | Partial interleaved RoPE (0.8 → 32 of head_dim 40), cached causal self-attention, cross-attention, fused gate/up SiLU MLP, untied `proj_out` |

The window schedule is genuinely load-bearing: with identical weights, HuggingFace's own encoder output differs by up to **3.29** between `[[2,1],[3,0]]`, `[[16,4],[16,0]]` and a global window — and the ONNX graph matches HuggingFace for all three.

ONNX initializer names match the checkpoint **1:1**; no renames beyond the shared `model.` prefix strip.

## Reuse

The decoder is the same architecture as offline Moonshine, so `MoonshineAttention`, `MoonshineDecoderLayer` and `MoonshineDecoderModel` are reused directly instead of duplicated. Everything the streaming variant actually changes is new code.

## Encoder context is added exactly once

HuggingFace's `MoonshineStreamingDecoder` adds its absolute position table to `encoder_hidden_states` with `+=`, mutating the caller's tensor — a single decoder call moves it by up to **1.73**, and a 34-step decode leaves it **58.8** away from the encoder output. Cached cross-attention means the mutated tensor is never read again, so generation is in fact unaffected: decoding three ways (`generate()`, a manual loop over a deliberately accumulating buffer, and a manual loop with a pristine clone each step) produces **identical tokens**.

A golden reference must not depend on a dead-store argument, so `scripts/generate_golden.py` wraps the speech-to-text reference in `non_mutating_encoder_context`, which clones `encoder_hidden_states` on decoder entry. Regenerating L4 and L5 under the guard reproduces both files **byte for byte** — that is the evidence the previous references were not silently accumulating. The exported ONNX decoder takes the encoder output as a read-only graph input and adds the position table on every call, which is the same semantics.

Three tests lock this down:

- the guard is proven non-vacuous — the unguarded decoder is asserted to mutate its argument, the guarded one is asserted not to;
- the ONNX decoder is asserted not to accumulate — after a full decode its encoder input buffer is unchanged, and replaying step 0 reproduces the original step-0 logits exactly;
- every decode step is compared against a non-mutating HuggingFace reference on the **full vocabulary distribution**, not just the argmax, and the resulting sequence is checked against the committed L5 golden.

## Supporting changes

- `MoonshineAttention` gains keyword-only `hidden_size`/`head_dim` (so an encoder of a different width can reuse it) and `qkv_bias`/`o_bias`. Previously all four projections were hardcoded bias-free while `MoonshineConfig` extracted `attention_bias`, so a bias-enabled checkpoint would have silently mis-loaded. Upstream gates Q/K/V on `attention_bias` and keeps the **decoder's** `o_proj` bias-free unconditionally, while the streaming **encoder** gates all four. Defaults unchanged (`False`).
- `Conv1d.padding` accepts a `(left, right)` pair so a convolution can be made causal without a separate `Pad` node.
- `SpeechToTextConfig.encoder_output_size` lets the decoder graph declare the real encoder width instead of assuming it equals the decoder's.
- `ArchitectureConfig.validate_execution_provider` — a new optional builder hook (see below).
- `scripts/generate_golden.py` now forwards the pinned `revision` on the speech-to-text path — it was previously unpinned.
- ORT GenAI export rejects `moonshine_streaming` for the same reason it rejects `moonshine`.

## float16 on CUDA is explicitly rejected

ORT's CUDA float16 fused attention kernel returns encoder frame 0 as **exactly zero** (`std 0.00000` where fp32 gives `std 0.52855`). That frame is the sparsest masked query row — it sees only 4 of 456 keys under the `(16, 4)` window. The same graph on **CPU** float16 gives `std 0.52861`, and CUDA float32/bfloat16 are correct with identical masks.

Root-caused with four controlled experiments:

1. Mask fill `-65504` → `-1e4`: still zero → **not** the fill magnitude.
2. Additive float bias → bool mask: still zero → **not** the mask representation.
3. Isolated `Conv` repro with `pads=[4,0]`, stride 2, fp16: CUDA == CPU bit-for-bit → **not** the causal convolution.
4. Replace the window bias with an all-allow bias: frame 0 becomes **correct** → the masked attention row is the trigger.

The combination was previously still selectable and silently produced a wrong first encoder frame. `ArchitectureConfig` now exposes an optional `validate_execution_provider(ep)` hook that the builder calls with the requested EP once the dtype is known, and `MoonshineStreamingConfig` overrides it to reject float16 + CUDA with an actionable message. Verified through the public `build()` API and therefore the CLI; float32/CUDA, bfloat16/CUDA and float16/CPU all still build.

This is deliberately **not** a graph-side workaround. Compensating for an execution-provider defect inside the model would hide it and diverge from upstream. `float32` remains the default and is unaffected.

## Evidence

Environment for every command (this machine has an editable `mobius` install pointing at a different worktree):

```
cd <this worktree>
$env:PYTHONPATH="$PWD\src"; $env:PYTHONIOENCODING="utf-8"
# verified mobius.__file__ resolves into this worktree's src/
```

### L1 / L2 / L3

| Level | Result |
|---|---|
| L1 | `build_graph_test.py -k moonshine` + `weight_alignment_test.py` — new cases for causal `pads=[4,0]`, `gamma` naming, `proj` only when widths differ, window-length validation, bias gating |
| L2 | `arch_validation_test.py [moonshine_streaming]` config-downloads / full-graph-builds / shapes; `yaml_schema_test.py` 252 passed |
| L3 synthetic | Encoder parity under heavy padding (60 of 200 frames valid, mask exact); three window schedules; with and without `attention_bias`; batch-3 **ragged** padding (150 / 90 / 40 valid frames) with per-row parity; decoder prefill + cached decode |
| L3 real weights | Real nonzero speech (`652-129742-0006.flac`, peak 0.59, 456 encoder frames) |

Real-weight fp32, ONNX encoder → **ONNX** decoder (not an HF intermediate):

- encoder `max|diff|` **8.3e-6**, encoder mask exact
- chained prefill vs the full HF model `max|diff|` **2.0e-5**
- full ONNX greedy generation **== `HF generate()` token-for-token (34 tokens)**, identical transcript:
  `Cauliflower mayonnaise, take cold boiled cauliflower, break into branches, adding salt, pepper and vinegar to season.`
- per-step full-logit parity against the non-mutating reference at `rtol=atol=1e-3` for **every** step

`tests/moonshine_streaming_integration_test.py`: **23 passed**.

### L4 / L5 goldens

Generated **independently from the upstream HuggingFace pipeline** under the non-mutating guard and committed:

- `testdata/golden/audio/moonshine-streaming-tiny.json` — top1 `9243`
- `testdata/golden/audio/moonshine-streaming-tiny_generation.json` — 33 tokens (≥ 20)

The YAML case sets `exact_match: true`, so L5 asserts **full sequence length + exact token IDs + exact decoded transcript**. Both pass on CPU and on CUDA (`MOBIUS_TEST_DEVICE=cuda MOBIUS_TEST_BUILD_EP=cuda`).

### Multi-dtype and multi-EP (real audio; reference = HF fp32, `|hidden|max` 6.70, `|logits|max` 24.53)

| dtype / EP | encoder max abs diff | prefill logit max abs diff | tokens | transcript |
|---|---|---|---|---|
| f32 / CPU | 0.0000 (rel 1.2e-6) | 0.0000 | exact | exact |
| f16 / CPU | 0.0391 (rel 5.8e-3) | 0.0219 | exact | exact |
| bf16 / CPU | — | — | — | **unsupported**: ORT CPU has no bf16 `MatMul` kernel |
| f32 / CUDA | 0.0412 (TF32) | 0.0110 | exact | exact |
| bf16 / CUDA | 0.1566 (p99.9 0.085) | 0.3621 | exact | exact |
| f16 / CUDA | — | — | — | **rejected at build time** (see above) |

CPU-vs-CUDA f32: encoder `max|diff|` 0.041, logit `max|diff|` 0.011, tokens equal. CUDA CLI build with `--execution-provider cuda` succeeds.

### Olive quantization (CPU)

NF4 (`OnnxBnb4Quantization`, olive 0.13.0) on the decoder: **138.5 MB → 60.2 MB (0.435)**, loads, and reproduces the **exact golden token sequence and transcript**.

### CLI

```
python -m mobius build --model moonshine-ai/moonshine-streaming-tiny \
  --revision f8e9dfd8c562c257c151a907b7b7f2fe8ff8511a <out>            # encoder/ + decoder/
python -m mobius build ... --runtime onnx-genai <out>                   # + audio_processor.json, inference_metadata.yaml
python -m mobius build ... --execution-provider cuda --dtype f32 <out>  # CUDA-specialized graph
```

The emitted `audio_processor.json` carries `pad_to_multiple_of: 80`, exactly the frame alignment the encoder's reshape requires. A test asserts the processor contract directly (`do_normalize False`, 16 kHz, `return_attention_mask True`, `pad_to_multiple_of == config.frame_length`, `len(input_values) % frame_length == 0`).

### Full suite

`tests/build_graph_test.py tests/cli_test.py src/` → **4892 passed**, 7 failed. All 7 failures were reproduced on a stashed baseline of this same branch and are pre-existing and unrelated (gguf q4_0, deepseek_v4 qmoe ×2, decompose_attention softcap ×2, qwen_image CUDA low-precision ×2). `lintrunner init` + `lintrunner f` + `lintrunner -a` clean with the pinned RUFF 0.16.2 / RUFF-FORMAT actually running.

## Waivers

- **ORT GenAI runtime load/generation**: not applicable. `write_ort_genai_config` rejects `moonshine_streaming` exactly as it already rejects `moonshine` — the runtime cannot host a variable-length raw-waveform encoder. A parametrized test asserts the rejection for both. The `onnx-genai` workflow runtime *is* supported and emits a complete package.
- **Foundry Local**: not run; it consumes ORT GenAI packages, which this architecture cannot produce.
- **bf16 on CPU**: ORT has no bf16 `MatMul` CPU kernel (`NOT_IMPLEMENTED` at `encoder/embedder/linear/MatMul`). bf16 is validated on CUDA instead.
- **f16 on CUDA**: rejected at build time, with the ORT kernel defect root-caused above.

## Intentional mixed precision

Per-frame CMVN and the `asinh` gain are computed in float32 and cast to the build dtype before the framing linear; `log_k` is marked `_keep_float32`. Squared raw-waveform amplitudes are ~1e-6 and collapse into float16 subnormals, so evaluating that stage in fp16 would be numerically meaningless. Everything downstream runs in the build dtype.


justinchuby:
@copilot update from main

src/mobius/_configs/_base.py
```diff
@@ -1459,6 +1459,22 @@ def validate(self) -> None:
                 "Invalid ArchitectureConfig:\n" + "\n".join(f"  - {e}" for e in errors)
             )
 
+    def validate_execution_provider(self, execution_provider: str) -> None:
+        """Reject build targets this architecture cannot execute accurately.
```
justinchuby: We don't need this or assume executor capability

src/mobius/_configs/_base.py
```diff
@@ -4474,6 +4500,198 @@ def from_transformers(cls, config, parent_config=None) -> MoonshineConfig:
...
+                f"{len(self.encoder_sliding_windows)} entries but the encoder has "
+                f"{self.encoder_num_hidden_layers} layers."
+            )
+
+    @property
+    def encoder_output_size(self) -> int:
+        """Encoder width; the decoder projects it when it differs from its own."""
+        return self.encoder_hidden_size
+
+    def validate_execution_provider(self, execution_provider: str) -> None:
```
justinchuby: Remove

src/mobius/_configs/_base.py
justinchuby: Removed the execution-provider validation hook from the builder and config base class; model configs no longer assume executor capabilities.

src/mobius/_configs/_base.py
justinchuby: Removed the Moonshine Streaming float16/CUDA execution-provider override and its tests.

src/mobius/_configs/_base.py
```diff
@@ -3161,6 +3171,175 @@ def from_transformers(cls, config, parent_config=None) -> MoonshineConfig:
...
+    #: Per-layer ``(left_window, right_window)`` attention spans of the encoder.
+    #: ``left`` counts the query position itself; ``right`` is the strict
+    #: lookahead. ``right == 0`` means the layer is fully causal.
+    encoder_sliding_windows: tuple[tuple[int, int], ...] = ((16, 4),)
+    encoder_sample_rate: int = 16_000
+    encoder_frame_ms: float = 5.0
+    #: Epsilon of the per-frame cepstral mean/variance normalisation.
+    encoder_cmvn_eps: float = 1e-6
+    #: Initial value of the learned asinh compression gain (stored as ``log_k``).
+    encoder_compression_k: float = 0.75
```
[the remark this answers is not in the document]
justinchuby: Removed `encoder_compression_k`; `log_k` is exclusively a checkpoint weight and no longer has an unused config default.

PR #578, opened 2026-08-24
GGUF export preflight: block-preserving lossless-only classification + budget (reconciled into #581)

justinchuby:
## What

Metadata-only, **block-preserving / lossless-only** classification and budget for
GGUF export, **folded into the #581 preflight library** (`mobius.integrations.gguf._preflight`).
This is the reconciled revision of the rejected PR: the standalone
`mobius.preflight_gguf` module and its duplicate `preflight-gguf` subcommand are
gone; the additive value lives on the single #581 code path.

Rebased/rebuilt onto current `main` (now includes #581 `7003bc7` and #576
`059bf5f`). Base retargeted from the deleted `sebastian/export-preflight-streaming`
to `main`. Single focused commit; no dependency on #576.

## Why the rejection is addressed

#581 shipped a `preflight-gguf` CLI + preflight library after this branch was
cut. The old branch added a *second* subparser (breaking `build_parser()` on
Python 3.12 with `conflicting subparser: preflight-gguf`) and duplicated header
parsing / shard resolution. It also carried three now-invalid blockers.

**Removed (net-new, now-false):**
- Duplicate module + duplicate `preflight-gguf` CLI ownership — exactly one
  registration remains (#581's); the parser builds on 3.12.
- Hardcoded `split.count > 1` refusal — `GgufShardSet` imports sharded sets
  directly, so GLM-5.2 `UD-IQ1_S` (6 shards) is **not** rejected for sharding.
- `merged_input_bytes` / second-copy disk budget — direct random access means no
  merged file; the budget reflects **real** artifacts only.
- `/mnt` staging recommendation — no environment paths in product defaults.
- No static DeepSeek 401/404 blocker, no bandwidth/tok-s inference, no payload
  download (headers only).

**Preserved / folded in (validated against the real repacker):**
- Dynamic architecture lookup (`glm-dsa → glm_moe_dsa`, from #581).
- Per-tensor-type lossless disposition against the **real** repacker
  (`native_block_spec` / `repack_quant_params`) + float passthrough — no
  model-name allowlist: `passthrough` / `native-preserve` / `repack` /
  `unsupported`.
- Real-artifact budget: ONNX external-data output estimate and a weights-only
  VRAM figure derived from the **export artifact** (not the source dtype).
- Honest blockers: the sparse-MoE fusion gate (#581) plus the **exact**
  unsupported native qtypes (e.g. `Q5_K/Q2_K/Q3_K`) are the real remaining code
[...]

PR #576, opened 2026-08-24
Stream sharded safetensors export + resumable full-checkpoint preflight

justinchuby:
## Motivation

Exporting an official full-size sharded checkpoint (e.g. `zai-org/GLM-5.2` — 753.3B params BF16, **1.507 TB across 282 shards**) with the current eager loader concatenates every shard into one in-RAM `state_dict`. Measured peak host RAM for that path is ~1.5–1.9 TB, which **exceeds available RAM on a 1.7 TB box and OOMs**. There is also no way to know the disk/RAM/VRAM budget before committing to a hundreds-of-GB download.

This lands the generic, model-agnostic code to (a) stream a sharded checkpoint to ONNX external data with a bounded working set, and (b) refuse a download up front when the budget is insufficient. No model-name allowlist.

## What

- **`stream_safetensors_to_model()`** (`integrations/_weight_loading.py`): a pass-through lazy per-tensor loader. Each ONNX initializer is bound to an `ir.LazyTensor` that reads its tensor from the owning safetensors shard on demand (`safe_open` reads headers only), so peak RAM is bounded to one tensor, not the whole checkpoint. It **refuses** rather than mis-exporting when the source is quantized (fp8 / `*_scale_inv`) or when a graph initializer has no matching checkpoint tensor (needs fusion/split/quantization) — those must use the eager `apply_weights` path.
- **`external_data_checksums()`**: deterministic sha256+size manifest for verifying byte-reproducible re-exports.
- **`preflight.py` + `mobius preflight` CLI**: a resumable dry-run that validates all shard URLs/metadata from the Hub or a local dir (no success-shaped fallback), computes exact source/output disk, peak host RAM (eager vs stream) and weights VRAM (sized from the export artifact), sums download+output on a shared filesystem, checks free disk/RAM, and **refuses (exit 2)** before download when budget is insufficient, RAM can't be verified, or metadata is unavailable. Resumable JSON state refuses on commit-identity drift.

The eager `apply_weights` / `_download_weights` path (incl. its pytorch + fp8 support) is unchanged.

## Proven behavior (real metadata, no download)

| model | source | int4 output | peak RAM stream / eager | int4 VRAM | verdict |
|---|---|---|---|---|---|
| GLM-5.2 (int4-qmoe) | 1.507 TB | 0.436 TB | 10.7 GB / 1942.2 GB | 500.8 GB | READY on 8×80 GB agg; REFUSED on single 80 GB |
[...]

PR #593, opened 2026-08-24
feat(deepseek-v4): default-off native CompressedSparseAttention (HCA ratio-128) export [C1, DRAFT]

justinchuby:
## Summary

**Slice C1** of the DeepSeek-V4-Flash CSA/HCA custom-kernel path. Adds a **default-off, property-gated** Mobius exporter feature that emits the already-merged frozen `pkg.nxrt::CompressedSparseAttention` v1 op (**ratio-128 "Heavily Compressed Attention"** subset) for property-matching layers, replacing the zero-valued *shape-anchor* preservation with **real op dataflow** and **threaded compressed state IO**.

This is **exporter-only**. No onnx-genai runner/KV changes; **no PagedAttention-compatibility claim**.

Design: `deckard-deepseek-v4-csa-hca-cuda-slice.md` (§5 property ABI, §8 tiny-test spec, §10 C1 slice).

## What changed
- **`mobius.models._deepseek_v4_csa`** (new): `plan_native_csa` property gate + `emit_hca_attention` op emission. The gate matches native-op **properties** (compression ratio, block size, head/latent/rope dims, dtype, carry planes), **not** a model-name/shape allowlist.
- **`ArchitectureConfig.native_csa: bool = False`** — export-time opt-in (no HF equivalent; `config_overrides={"native_csa": True}`).
- **`DeepSeekV4Attention._forward_native_hca`** — emits the 11-input / 3-output op with pre-rotated `query`/`current_kv` (dense sliding-window ring over the existing dense KV IO), real *unrotated* compressor activations, f32 casts for the f32-cache-format op inputs, and threaded `past_*`/`present_*` compressed state. The shared inverse RoPE + output projection run unchanged on `Y`.
- **`DeepSeekV4Task`** — threads deterministic `past_compressed_kv.{i}` / `past_compression_carry.{i}` inputs and `present_*` outputs (explicit shapes, since the custom op has no Python shape inference), with a **dynamic record axis** and **fixed carry planes** so present state chains back into past inputs for **≥16 decode steps**.

## Fail-closed (never silent dense)
Requesting native CSA for **ratio-4 CSA**, an **MTP** layer, **quantized** compressor weights, or an **unknown ratio** raises `NativeCsaExportError`. Ratio-0 dense layers and feature-off legitimately return `None` (dense) — those are not a suppressed fallback.

## Off = byte-identical
With the feature off, the exported graph is **byte-identical** to the existing dense correctness export (verified) and emits **zero `pkg.nxrt` ops / no compressed IO**.

## Tests (shape-faithful / structural — no weight download, no ORT run)
The frozen op has no Python shape inference and no Rust EP in the Python env, so tests assert structure: exact op attrs + 11/3 IO, threaded compressed-state IO names/shapes/dtypes, decode chainability, **f32 inputs under a FLOAT16 model dtype** (GLM DSA precedent), **no dead anchor** for the HCA layer, byte-identical disabled baseline, defaults-off + opt-in, and typed rejects (ratio-4 / quantized / MTP / unknown-ratio). Existing DeepSeek-V4 tests unchanged.

Local validation (this env):
- `deepseek_v4_flash_test.py` + `deepseek_v4_test.py`: **30 passed**
- `_configs`/`tasks`/`glm_moe_dsa` related tests: **168 passed**
- `ruff check` + `ruff format --check`: clean

## Scope / non-goals
- ❌ No onnx-genai runner or KV/cache changes (later slice).
- ❌ No PagedAttention-compatibility claim.
- ❌ ratio-4 learned-indexer/top-k selection is a follow-up slice.

## Status
**DRAFT — stop for independent review. Do not merge.** CI is async; local tests above are the validation of record.

### DIFF
--- a/src/mobius/_configs/_base.py
+++ b/src/mobius/_configs/_base.py
@@ -494,6 +494,17 @@ class ArchitectureConfig(BaseModelConfig):
     num_hash_layers: int = 0
     swiglu_limit: float = 0.0
     num_nextn_predict_layers: int = 0
+    # DeepSeek-V4 native compressed-sparse-attention export toggle. Not an
+    # upstream HF field -- a Mobius export-time opt-in (default False) that
+    # replaces the dense CSA/HCA correctness fallback with the frozen
+    # ``pkg.nxrt::CompressedSparseAttention`` v1 op for property-matching
+    # ratio-128 (HCA) layers. Off by default so every shipped graph stays
+    # byte-identical and ``pkg.nxrt``-free unless explicitly requested via
+    # ``config_overrides={"native_csa": True}``. When requested, layers that
+    # do not satisfy the frozen op contract raise a typed export error rather
+    # than silently emitting the dense fallback (see
+    # ``mobius.models._deepseek_v4_csa``).
+    native_csa: bool = False
 
     # GLM-5.2 (``glm_moe_dsa``) DeepSeek Sparse Attention (DSA).
     #
@@ -963,6 +974,11 @@ class ArchitectureConfig(BaseModelConfig):
             ),
             swiglu_limit=getattr(config, "swiglu_limit", 0.0),
             num_nextn_predict_layers=getattr(config, "num_nextn_predict_layers", 0),
+            # DeepSeek-V4 native CSA/HCA export opt-in. No HF equivalent --
+            # every checkpoint defaults to the dense correctness fallback;
+            # ``config_overrides={"native_csa": True}`` opts a build into the
+            # frozen ``pkg.nxrt::CompressedSparseAttention`` v1 ratio-128 path.
+            native_csa=getattr(config, "native_csa", False),
             # GLM-5.2 DSA. ``use_dsa`` has no HF equivalent -- every checkpoint
             # defaults to the sparse path; ``--glm-full-attention`` overrides
             # it afterwards via ``dataclasses.replace``/``config_overrides``.
--- a/src/mobius/_configs/_base.py
+++ b/src/mobius/_configs/_base.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 
 import dataclasses
 import math
+from typing import TYPE_CHECKING
 
 import onnx_ir as ir
 import torch
@@ -21,6 +22,9 @@ from mobius._configs._sub_configs import (
     TTSConfig,
     VisionConfig,
 )
+
+if TYPE_CHECKING:
+    from mobius.integrations._block_quant import BlockQuantScheme
 
 DEFAULT_INT = -42
 
@@ -631,7 +635,16 @@ class ArchitectureConfig(BaseModelConfig):
     codec_decoder: CodecDecoderConfig | None = None
     codec_encoder: CodecEncoderConfig | None = None
     quantization: QuantizationConfig | None = None
-
+    # Deferred block-scaled FP8 / packed-FP4 scheme (DeepSeek-V4-Flash native
+    # CSA). Recorded by ``from_transformers`` when ``native_csa`` opts into
+    # deferring #602's config-resolution block-quant reject so that graph
+    # construction can progress past the former generic weight-shape mismatch.
+    # The runtime-capability gate
+    # (``mobius.models._deepseek_v4_csa.assert_native_runtime_supports_block_quant``)
+    # then fails closed on the *runnable* full export until nxrt advertises real
+    # block-FP8 / planar-FP4 format strings. ``None`` for every ordinary,
+    # per-tensor-fp8, or non-native path.
+    block_quant_scheme: BlockQuantScheme | None = None
     # HuggingFace model_type and special token IDs — populated by from_transformers()
     # so that genai_config.json can be written without re-fetching the HF config.
     model_type: str | None = None
@@ -1207,8 +1220,28 @@ class ArchitectureConfig(BaseModelConfig):
         if resolved is not None:
             options["dtype"] = resolved
 
-        # Quantization config
-        quant = QuantizationConfig.from_transformers(config)
+        # Quantization config. Block-scaled FP8 / packed-FP4 checkpoints
+        # (DeepSeek-V4-Flash) are rejected here by #602's typed
+        # ``BlockQuantExportError`` (the INT4/per-tensor path cannot load them).
+        # For a native-CSA export we *defer* that reject so graph construction
+        # can progress past the former generic weight-shape mismatch; the
+        # runtime-capability gate
+        # (``mobius.models._deepseek_v4_csa.assert_native_runtime_supports_block_quant``,
+        # invoked at weight-load / full-export) then fails closed until nxrt
+        # advertises real block-FP8 / planar-FP4 format strings. Every non-native
+        # path keeps #602's early, loud reject.
+        from mobius.integrations._block_quant import (
+            BlockQuantExportError,
+            BlockQuantScheme,
+        )
+
+        try:
+            quant = QuantizationConfig.from_transformers(config)
+        except BlockQuantExportError:
+            if not getattr(config, "native_csa", False):
+                raise
+            options["block_quant_scheme"] = BlockQuantScheme.from_hf_config(config)
+            quant = None
         if quant is None and parent_config is not None:
             quant = QuantizationConfig.from_transformers(parent_config)
         if quant is not None:
--- a/src/mobius/models/_deepseek_v4_csa.py
+++ b/src/mobius/models/_deepseek_v4_csa.py
@@ -68,6 +68,14 @@ import onnx_ir as ir
 from onnxscript import OpBuilder
 
 from mobius._configs import ArchitectureConfig
+from mobius.integrations._block_quant import (
+    MXFP4_MICROSCALE_BLOCK,
+    BlockQuantExportError,
+    BlockQuantScheme,
+    QuantizedTensorDescriptor,
+    QuantKind,
+    runtime_representation_gap,
+)
 
 # Frozen op identity -------------------------------------------------------
 CSA_OP_TYPE = "CompressedSparseAttention"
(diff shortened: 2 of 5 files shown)


justinchuby:
## B1 — native CSA export extended to the full interleaved ratio-4 + ratio-128 schedule (commit `3e10955`)

The C1 slice emitted the frozen `pkg.nxrt::CompressedSparseAttention` v1 op only for ratio-128 (HCA) and **typed-rejected the 21 ratio-4 CSA layers**, making a full `native_csa=True` export of the official checkpoint impossible. B1 extends the exporter to the whole official `deepseek-ai/DeepSeek-V4-Flash@60d8d707…` schedule: **both ratios emit, or the export typed-rejects before serialization — never partial-native + silent dense.**

### Real-config verification (metadata + graph only, no weight download)
Built the real 43-layer target graph (head_dim=512, 64 heads) with `native_csa=True` — **built with no typed reject** (1166.8 s, peak RSS ~1.5 GB):

- **41 `CompressedSparseAttention` nodes = 21 ratio-4 (all 19-in / 6-out, `fp8_e4m3_block64`) + 20 ratio-128 (all 11-in / 3-out, `f32`)** — exactly the official schedule.
- Threaded state IO: `past_/present_compressed_kv` ×41, `…compression_carry` ×41, and the ratio-4-only `past_/present_index_key` ×21, `…index_carry` ×21, `selected_indices` ×21 — deterministic layer-suffixed names, dynamic record axes.
- `plan_native_csa` widths on the real config are exact: ratio-4 `stored_width=583`, `index_stored_width=68`, `compressor_width=1024`, `index_compressor_width=256`, `carry_slots=8`; ratio-128 `stored_width=512`, `carry_slots=128`; MTP = `None`.

### Design (clean, no back-compat shims — per user directive)
`HcaLayerPlan → CsaLayerPlan` (every native-op property for both ratios); `plan_native_csa` returns a ratio-4 plan; `emit_csa_attention` branches 11/3 vs 19/6 and raises on any missing learned-indexer input; the indexer's `wq_b(q_lora)`/`weights_proj`/compressor tensors — previously dead anchors — are now **real dataflow** (op applies index RoPE/Hadamard/FP4 internally). MTP recurrence / unknown ratio / unsupported quant / invalid fp8-fp4 packing still fail closed. Feature stays **off by default**; disabled export byte-identical.

### Tests
35 shape-faithful/structural tests pass (no weight download / no ORT run): ratio-4 19/6 with exact attrs; interleaved `[0,4,128]` emits both; ratio-4 index cache (uint8 fp4 width 17) + f32 index carry + int32 `selected_indices`; learned indexer wiring (`index_query = Reshape(MatMul)`); no dead index anchor; prefill + ≥16-decode chainability; typed rejects for invalid packing dims / missing index config; ratio-4 plan contract; ratio-128 + existing DeepSeek-V4 tests migrated atomically. `ruff check` + `ruff format --check` clean.

### Blocker — B2 (FP8/FP4 weight loading, @Roy-1's domain)
Real-checkpoint weight loading via the stable streaming API `stream_safetensors_to_model` (#576, lazy per-shard, no all-shard state_dict) typed-refuses up front:

```
ValueError: Checkpoint appears quantized (fp8 / scaled weights, e.g.
['layers.0.attn.wkv.scale', 'layers.0.attn.wkv.weight', ...]); the pass-through
streaming loader cannot dequantize it and would drop the weight scale.
  at _weight_loading.py:519 in stream_safetensors_to_model
```

Census: **34,542 of 69,187 tensors are `F8_E4M3` block-fp8** (attention `wkv/wq_*/wo_*`, `indexer.wq_b`, experts). B2 is reached **only after** the native gate accepts both ratios, fires at **layer 0** (before any CSA layer would execute), and is a **typed refusal, never a silent dense fallback**. I coordinate through Roy-1's stable API only and don't touch `_weight_loading.py` / quantization primitives.

### Honest status
Structure (op wiring/attrs/IO/state carry) for the full official schedule is complete and validated. **No performance claim** — no ratio-4 layer has executed sparse and no ≥16-decode has run; real-checkpoint weights remain B2-blocked. Staying **DRAFT** for independent review; not merging, not waiting on CI (local validation done). Run artifacts on disk5 `converted/deckard-csa-c2-ratio4/`.

justinchuby:
### #602 consumption — local stacked-branch composition (fail-closed) ✅

Per the directive to consume Mobius #602's generic B2 API
(`mobius.integrations._block_quant`) **without waiting for merge**, I validated
the #593 × #602 composition on a **local, unpushed stacked test branch**
(`deckard/deepseek-v4-csa-blockquant-stack` = #593 `3e10955` + #602 `a38d38d7`).
**#593 itself stays independent** of the unreviewed contract (`_block_quant.py`
absent from this branch; HEAD unchanged; nothing merged/pushed here).

**What the composition proves (fail-closed, no perf claim):**
- `ArchitectureConfig.from_transformers` on the official block-fp8 + fp4-expert
  properties raises the typed `BlockQuantExportError` **at config resolution**
  (via #602's gate in `QuantizationConfig.from_transformers`) — **before** any
  `CompressedSparseAttention` node is planned or serialized. No silent dense, no
  partial-native graph.
- `native_csa=True` **cannot bypass** that gate (gate dominates the opt-in);
  fp4 experts alone still fail closed; the gate is **property-based**, not a
  model-name/native_csa allowlist (a non-block-quant DeepSeek-V4 schedule still
  resolves and plans native CSA for both ratios); per-tensor fp8 does not
  over-fire.
- No FP8/FP4 parsing added to `deepseek_v4.py` — detection is entirely #602's
  `_block_quant` contract.

**Real-checkpoint operational evidence** (rev `60d8d707…`, metadata/header-only,
no weights): the real `config.json` trips the fail-closed gate; `_block_quant`
classifies real layer-0 tensors by property — `attn.wkv`/`attn.wq_b` →
`block_fp8` (F8_E4M3 + F8_E8M0 2D block scale), `ffn.experts.0.w1` →
`fp4_packed` (I8 planar + F8_E8M0 block-32 micro-scale) — each with a non-None
`runtime_representation_gap`; `plan_routed_expert_bank` typed-rejects the fp4
bank with the exact nxrt ABI gap.

**Tests/lint on the stack:** 80 passed (35 #593 CSA + 40 #602 + 5 new
composition); ruff `check` + `format --check` clean. One caller update (no BC
shim): `DeepSeekV4Attention.forward` returns a 3-tuple on #593, so #602's
fused-GQA window test helper unpacks three values.

**Gate for a non-fail-closed export** remains the onnx-genai native block-FP8 /
planar-FP4 CSA **runtime** slice (separate; not yet advertising typed support).
When #602's review approves, it merges into #593 and these composition tests
migrate atomically. Staying DRAFT.

justinchuby:
### Rebased onto main (#578 + #602 merged); two-stage block-quant composition

Rebased this branch onto latest `main` now that **#578** (`6992181`) and **#602** (`d16cd4a`, `mobius.integrations._block_quant` block-FP8 / packed-FP4 loading contract) are merged. Consumes `_block_quant` **directly from main**; the temporary stacked duplicate-parse branch is retired. New linear history: `5028076` + `db491a3` (ratio-4) + `c2e15ce` (C1) on `d16cd4a`/`6992181`. Force-with-lease pushed (`3e10955…5028076`). Still **DRAFT — do not merge.**

**Behavior split into the two stages the directive requires** — *"graph construction should progress past the former generic shape mismatch, full export must typed-reject at the runtime capability gate":*

| Path | Stage | Behavior |
|---|---|---|
| Non-native (default) | config resolution | keeps #602's early, loud `BlockQuantExportError` (unchanged) |
| `native_csa` | config resolution | **defers** the reject: records `block_quant_scheme`, `quantization=None` |
| `native_csa` | **graph construction** (`build_from_module`) | **PROGRESSES** — emits CSA nodes + compressed state IO (no shape mismatch) |
| `native_csa` | **full export** (`preprocess_weights` → `apply_weights`) | **typed-rejects** at the runtime-capability gate |

The gate (`assert_native_runtime_supports_block_quant`) sources its gap string from `_block_quant.runtime_representation_gap`, so it **tracks the real nxrt format strings and opens automatically** when the native block-FP8 / planar-FP4 runtime slice lands — no change here. Never a silent dense fallback, never partial-native. The CLI convert path is exactly two-staged (`build_from_module` then `preprocess_weights`+`apply_weights`), so the gate replaces the former generic `"Weight shape mismatch"` at the correct stage; the streaming loader refuses `preprocess_weights`-needing models and routes them to the eager path, so DeepSeek-V4's full export always passes through the gate.

**No BC shims** (per the no-backward-compat directive): `ArchitectureConfig` gains a `block_quant_scheme` field; `deepseek_v4.py` adds **no** FP8/FP4 weight parsing — it delegates to `_block_quant`.

**Validation (local; not waiting on CI):**
- `deepseek_v4_flash_test.py`: **45 pass** (37 + 8 composition tests) — non-native fail-closed at config resolution; fp4-experts-alone owned; native_csa defers (scheme recorded, quant None); full export typed-rejects at the runtime-capability gate; graph construction progresses past block-quant (CSA node built); `preprocess_weights` enforces the gate; property-gated (not a blanket V4/native_csa refusal); per-tensor fp8 not over-owned.
- #602 + config suites unchanged: `_block_quant_test` + `_base_test` + `_extractors_test` = **69 pass**.
- `ruff format` + `ruff check` clean on the 4 touched files (`_configs/_base.py`, `models/_deepseek_v4_csa.py`, `models/deepseek_v4.py`, `models/deepseek_v4_flash_test.py`).

**No performance claim.** Ratio-4 sparse execution and ≥16-decode stay blocked on the unmerged native block-FP8 / planar-FP4 runtime; this slice only makes graph construction progress and the runnable export fail closed at the honest capability gate. Stopping for independent review.

<sub>Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com></sub>

justinchuby:
## Roy — independent review, HEAD `50280766a42e85c3321c0092f81df49ad1097157`

**Verdict: APPROVE EXPORT SCHEMA SLICE**

Read-only review, no edits. Worktree rebased cleanly on latest `main` (base = current tip, one file auto-merged with #599, zero conflicts).

### Schema/contract — verified against authoritative Rust kernel source
Cross-checked `_deepseek_v4_csa.py`'s op-contract constants against `crates/onnx-runtime-ep-cpu/src/kernels/compressed_sparse_attention.rs` in onnx-genai: exact match on required inputs (11 base, +8 for ratio-4 = 19), input/output arity bounds (11..20 / 3..6), attribute set, `sink_mode='logit_only'`, both cache formats (`f32` for HCA, `fp8_e4m3_block64`/`fp4_e2m1_block32` packed widths for CSA — `_fp8_block64_width`/`_fp4_width` mirror the Rust `stored_width` formulas exactly), and `carry_slots` (128 for HCA, fixed 8 for CSA per the Rust `CARRY_SLOTS` constant). The optional 20th input (`attention_bias`) is genuinely unused by DeepSeek, not a gap.

### Full official schedule / no partial-native fallback
`test_native_csa_emits_both_ratios_for_interleaved_schedule` (compress_ratios=[0,4,128]) proves a single graph emits **both** frozen-op variants with ratio-correct arity/attrs, never a partial-native+dense mix. `_forward_native_csa` fires unconditionally whenever `layer.csa_plan is not None` — there is no silent-fallback branch. Disabled path is proven byte-identical (`_fill_and_serialize(model_unset) == _fill_and_serialize(model_false)`, plus zero CSA nodes / zero compress-named IO).

### No zero anchors (compressor **and** indexer)
Single guard (`self.compressor is not None and self.csa_plan is None`, `deepseek_v4.py:982`) wraps both the compressor's *and* the indexer's zero-shape-anchor construction — there is no separate indexer-only anchor path that could leak through when the compressor anchor is skipped. Traced `_forward_native_csa`: when `plan.is_ratio4`, `self.indexer(op, hidden_states, query_lora)` is called with real activations (confirmed by `test_native_csa_ratio4_wires_learned_indexer`/`test_native_csa_ratio4_no_dead_index_anchor` — producer op-type and initializer-presence assertions, not just "doesn't crash").

### Two-stage #602 composition
Verified in code and by 6 dedicated tests (defers/progresses/full-export-typed-reject/preprocess_weights-enforces/property-gated-not-blanket/per-tensor-fp8-not-owned): `native_csa=True` defers `BlockQuantExportError` at config resolution (records `block_quant_scheme`, `quantization=None`); graph construction progresses (CSA node built); but `assert_native_runtime_supports_block_quant` — the literal first statement of `preprocess_weights` — still fails closed **today**, because onnx-genai #2054 (which would add `block_fp8`/`fp4_planar` to `NXRT_BLOCK_FORMATS`) is **closed, not merged** (`gh pr view 2054`: `state: CLOSED, mergedAt: null`). The gate fires only on the block-quant *properties*, never on the model name or on `native_csa` alone — confirmed by the explicit over-firing guard tests.

### CLI/streaming bypass
`native_csa` has **zero CLI exposure** in this slice — opt-in only via `config_overrides={"native_csa": True}` (Python API), a stronger safety margin than a CLI flag would be. `stream_safetensors_to_model` (pre-existing, unmodified) cannot silently bypass the gate for a real DeepSeek-V4 checkpoint: it independently refuses any quantized source (fp8/`*_scale_inv` detection) and any checkpoint whose keys don't already 1:1 match graph initializer names — which a model needing `preprocess_weights` renaming always trips — so it fails closed via its own pre-existing guards, not via this PR's logic.

### MTP exclusion — defense in depth
Two independent mechanisms both fail closed: (1) MTP layers are constructed with `layer_id = num_hidden_layers + index`, always past `compress_ratios`' length, defaulting to dense; (2) `plan_native_csa(..., is_mtp=True)` explicitly rejects with `NativeCsaExportError` regardless of ratio (`test_plan_native_csa_rejects_mtp`) — property-based on the `is_mtp` flag, not name/index-length coincidence.

### Tests / lint — independently reproduced
- `deepseek_v4_flash_test.py`: **45 passed** (exact match to claim).
- `_block_quant_test.py` + `_base_test.py` + `_extractors_test.py`: **69 passed** (exact match to claim).
- Broader regression sweep (`models/`, `_configs/`, `integrations/_block_quant_test.py`, `integrations/_weight_loading_test.py`, `tasks/`): **832 passed, 4 skipped, 0 failed**.
- `ruff check` + `ruff format --check` on all 5 touched files: clean.

### Should-fix, non-blocking
The Rust kernel's `execute_ratio4` hard-asserts `head_dim==512 && qk_rope_head_dim==64` at *runtime*; the Python export-time gate (`_fp8_block64_width`) only enforces the weaker generic packing constraint `(head_dim − qk_rope_head_dim) % 64 == 0`, not the exact 512/64 pair. An out-of-family config could pass export and only fail later at ORT execution time. Confirmed moot for the actual target model today: `_v4_hf_config`/`test_real_config_fields_extract` pin DeepSeek-V4-Flash's real architecture to exactly `head_dim=512, qk_rope_head_dim=64` ("mirrors the official rev"), and any divergence still fails closed (never silently wrong), just later than ideal. Recommend a follow-up export-time assert mirroring the exact Rust constraint for ratio-4 layers — not a merge blocker.

### Exact dependencies to state
1. **onnx-genai #2054** (or a successor landing `block_fp8`/`fp4_planar` into `NXRT_BLOCK_FORMATS`/`runtime_representation_gap`) — required before the runtime-capability gate can ever open for a real block-quant DeepSeek-V4-Flash checkpoint. Currently closed/unmerged; gate is genuinely fail-closed today, not a stale promise.
2. **Runtime state threading** (`tasks/_deepseek_v4.py`): `_compressed_inputs`/`_compressed_outputs` correctly thread `past_compressed_kv`/`past_compression_carry` (+ `past_index_key`/`past_index_carry` for ratio-4) per-layer with dynamic record axes and fixed carry-plane shapes; MTP's `_build_mtp` never touches this threading (2-output `module.mtp[0](...)` signature, consistent with guaranteed-dense MTP). No further onnx-genai dependency beyond #2054 — the CSA/HCA op schema/kernel itself already exists in onnx-genai independent of the block-quant runtime-format gate.

Posting this as the review verdict.

PR #591, opened 2026-08-24
Fuse GLM native-block routed experts into sparse pkg.nxrt::BlockQuantizedMoE (+ build_from_gguf wiring)

justinchuby:
## Native-block sparse-MoE fusion + `build_from_gguf` integration

Collapses GLM-style **routed** native-block expert storms (a per-expert
`pkg.nxrt::BlockQuantizedMatMul` for every gate/up/down of every expert,
dispatched by an `Equal` one-hot mask) into **one sparse top-k
`pkg.nxrt::BlockQuantizedMoE` per layer**, reusing the existing native blocks
**byte-for-byte** (no dequantize, no requantize). Gated on graph/tensor
structure, never on model names.

### What's here
1. **Rewrite** (`rewrite_rules/_block_quantized_moe_fusion.py`) — the transform:
   finds the dense-fallback MoE subgraph, validates **all** layers, then emits.
   Two-pass so an unfusable layer fails **atomically** with the graph untouched.
2. **Builder integration** (`integrations/gguf/_builder.py`) — wires the fusion
   into `build_from_gguf` at the **single** post-export authority point (step
   9b, after `pkg.apply_weights`, once every native block carries real packed
   bytes), then a final-graph-state honesty gate.

### Applicability verdict (Phase-1 audit)
- GLM-5.2 UD-IQ1_S/M routed experts are **per-projection mixed-format** per
  layer (fused `gate==up` in iq1_s/iq1_m; `down` in a higher-bit format such as
  iq4_xs). A v1 single-`format` `BlockQuantizedMoE` therefore fuses **0** GLM
  layers.
- Expressing a mixed-format layer needs the **`block_layout_version=2`
  per-projection ABI** (`fc1_format`/`fc2_format`/`fc3_format`). That runtime
  ABI lives on an **unmerged onnx-genai branch**, so mobius's runnable runtime
  does **not** yet have it.
- **Honest consequence:** GLM-5.2 mixed-format layers **typed-reject**
  (`SparseMoEExportError`, message names `block_layout_version=2`) until the
  runtime ships v2 — we never emit an unrunnable v2 node by default. **Uniform**
  native-block routed MoE **does** fuse today (v1).

### Honesty / fail-closed policy
- Default `allow_dense_moe=False`. Unsupported layers (mixed w/o v2, per-expert
  bias, ragged/incomplete groups) raise the shared **`SparseMoEExportError`**
  atomically; the graph is left as the original dense storm.
[...]

PR #581, opened 2026-08-24
Direct multi-shard GGUF import + GLM-5.2 (glm-dsa→glm_moe_dsa) resolution

justinchuby:
## Summary

Enables importing **large split-set GGUF checkpoints directly** (north star: `unsloth/GLM-5.2-GGUF` UD-IQ1_{S,M} — six-file, 201.832 / 212.801 GiB sets) without merging shards into a second on-disk file, resolves the GLM-5.2 `glm-dsa` architecture to the canonical `glm_moe_dsa` model type, and adds a fail-closed sparse-MoE honesty gate plus a metadata-only preflight.

This is independent of PR #1854. It does **not** touch PagedAttention or onnx-genai Slice7A.

### 1. Multi-shard reader — `_shard_set.py`
- `GgufShardSet` assembles a `-000i-of-000N.gguf` split set into one logical, `GGUFModel`-compatible reader. Tensors are read on demand from the **owning shard** (each memory-mapped by `gguf.GGUFReader`) — **never concatenated into a second ~228 GB GGUF**. Memory stays bounded to one tensor/block chunk + metadata.
- Native IQ blocks preserved **byte-for-byte** into the existing repacker/external-data path.
- Discovery confined to the requested directory; driven by authoritative `split.*` metadata cross-checked against filenames (no blind string concatenation).
- **Fail-closed** validation: contiguous `split.no`, declared `split.count`/`split.tensors.count` agreement, unique tensor names, tensor counts, offset/alignment bounds (catches truncated/corrupt shards), architecture/tokenizer/template identity, optional sha256/size checks. Rejects missing / duplicate / mixed-revision / mixed-quant / corrupt shards.
- `open_gguf_model()` returns a plain `GGUFModel` for single files (**behaviour unchanged**) or a `GgufShardSet` for split sets.

### 2. Architecture resolution — `_config_mapping.py`
- Explicit, **metadata-driven** `glm-dsa`/`glm_dsa` → `glm_moe_dsa` format bridge (no filename/model-name heuristics) + GLM-5.2 MLA/MoE/DSA key map.
- `assert_glm_moe_dsa_resolvable()` verifies head/layer/expert/MLA/DSA-indexer properties **before** selecting `GlmMoeDsaCausalLMModel`, accumulating **all** missing properties as precise rejection reasons. Wired into `build_from_gguf`.

### 3. Sparse-MoE honesty gate — `_builder.py`, `_flags.py`
[...]

PR #599, opened 2026-08-24
Add opt-in --paged-attention export for dense MLA (LATENT PagedAttention) [Slice 3B]

justinchuby:
## Slice 3B — Mobius opt-in `--paged-attention` export (dense MLA, LATENT)

Emits `com.microsoft::PagedAttention` v1 in **LATENT / absorbed-MLA** mode for property-compatible **dense MLA** (DeepSeek-V2/V3, GLM-5.2 `--glm-full-attention`). Builds on the merged onnx-genai native runtime contract: audit/typed-validator/oracle (#1940), KV index emission (#1955), CUDA LATENT kernel (#1978).

### Guarantees
- **Default off.** Feature-off exports are **byte-identical** to the current dense-MLA graph (all new code is gated behind the flag/task). A byte-identical test locks this.
- **Property-based eligibility, never model names.** Decided from semantic geometry: latent width, `v_head_size`, partial-RoPE suffix (`rotary_offset = kv_lora_rank`), cache dtype (fp16/bf16), and the onnx-genai-kv page/block constraints (head_size %8, kv_lora %8, rotary_dim %16, rotary_offset %8, block pow2 ≥16).
- **Typed reject, never silent fallback.** Active query-dependent sparse selection (GLM **DSA/IndexShare**, DeepSeek-V4 **CSA/HCA**), **MTP**, **quantized cache**, **head_sink**, **qk-norm**, **sliding window** all raise a typed reason at model construction / build. Feature-on with an incompatible geometry errors. GLM's *vestigial* indexer config (dropped under `--glm-full-attention`) does **not** reject.
- **One cache authority.** The op consumes/mutates caller-owned page buffers in place; Mobius **allocates no pages** and creates no second cache manager. `onnx-genai-kv` stays the sole authority. The graph binds caller-owned `block_table` / `slot_mapping` / cumulative + past lengths / per-layer LATENT `key_cache`, aliases the cache in place, and derives token positions from the length tensors (no `position_ids` input).

### Emitted contract
- One `PagedAttention` node per layer, `_domain="com.microsoft"`, LATENT attrs (`kv_cache_layout=LATENT`, `kv_num_heads=1`, explicit `scale`, `v_head_size=kv_lora_rank`, `rotary_offset=kv_lora_rank`, `do_rotary=1`); `rotary_dim` is **derived** from `cos_cache`, never emitted.
- Model inputs: `input_ids`, `block_table` (i32), `slot_mapping` (i32), `cumulative_sequence_length` (i32), `past_seqlens` (i32), per-layer `key_cache.{i}` `[num_blocks, block_size, 1, head_size]`. Outputs: `logits`, `updated_key_cache.{i}` (aliases input).
- `kv_b_proj` is absorbed into the query/output projections at weight-apply time; numeric contract mirrors the onnx-genai equivalence oracle.

### Tests (same commit)
- **component** (`components/_paged_mla_test.py`): eligibility rejects, absorption, numpy **decomposed-vs-absorbed LATENT parity** (rel < 1e-6), structural emission.
- **model-level** (`models/paged_mla_export_test.py`): feature-off byte-identical; feature-on structural + exact op attrs/inputs/outputs/model IO; typed rejects (DSA/CSA/HCA/MTP/window); torch-tensor weight absorption; DeepSeek-V3 path proves eligibility is not name-gated.
- **CLI** (`tests/cli_test.py`): `--features paged-attention` plumbing + task resolution + mutual-exclusion errors.
- Existing GLM / DeepSeek / task / build-graph tests pass unchanged.

### Scope / gates
- **No full-size performance claim.** Tiny correctness only. Native runtime full-size verification is separate.
- Executing the emitted graph needs **ORT ≥ 1.29 CUDA** (contrib op is unknown to the base onnx checker, so model-level assertions are structural).
- Avoids #578 (GGUF preflight) and #591 (BQMoE) files.

**Draft — do not merge.** Independent review required (reviewer excluding Leon/Sapper; Gaff or Roy final approval).

### DIFF
--- a/src/mobius/__main__.py
+++ b/src/mobius/__main__.py
@@ -44,6 +44,7 @@ _BUILD_FEATURES: dict[str, str] = {
     "prune-prefill-prefix": "prune_prefill_prefix",
     "text-only": "text_only",
     "glm-full-attention": "glm_full_attention",
+    "paged-attention": "export_paged_attention",
 }
 
 
@@ -269,6 +270,23 @@ def _cmd_build(args: argparse.Namespace) -> None:
         }
     else:
         static_cache_params = None
+
+    # PagedAttention (LATENT dense-MLA) export uses the paged-cache task with
+    # caller-owned page buffers. It is a distinct cache authority, so it cannot
+    # be combined with the static-cache task or an explicit --task.
+    export_paged_attention = getattr(args, "export_paged_attention", False)
+    if export_paged_attention:
+        if static_cache_params is not None:
+            raise SystemExit(
+                "Error: --features paged-attention cannot be combined with "
+                "--features static-cache."
+            )
+        if task is not None:
+            raise SystemExit(
+                "Error: --features paged-attention cannot be combined with --task. "
+                "Remove --task to use --features paged-attention."
+            )
+        task = CausalLMTask(paged_cache=True)
     trust_remote_code = args.trust_remote_code
     revision = args.revision
     output_dir = args.output_dir
@@ -358,6 +376,16 @@ def _cmd_build(args: argparse.Namespace) -> None:
                     f"model_type 'glm_moe_dsa' (got '{model_type}')."
                 )
             config = dataclasses.replace(config, use_dsa=False)
+        if export_paged_attention:
+            from mobius.components._paged_mla import paged_attention_rejection
+
+            config = dataclasses.replace(config, export_paged_attention=True)
+            reason = paged_attention_rejection(config)
+            if reason is not None:
+                raise SystemExit(
+                    f"Error: --features paged-attention is not supported for this "
+                    f"model: {reason}"
+                )
         if static_cache_params is not None:
             task = _resolve_static_cache_task(model_type)
         elif task is None:
@@ -406,6 +434,7 @@ def _cmd_build(args: argparse.Namespace) -> None:
             kv_cache_scales=kv_cache_scales,
             prune_prefill_prefix=prune_prefill_prefix,
             glm_full_attention=args.glm_full_attention,
+            export_paged_attention=export_paged_attention,
         )
 
     _save_package(pkg, output_dir, args, optimize, component_filter)
--- a/src/mobius/_configs/_base.py
+++ b/src/mobius/_configs/_base.py
@@ -503,6 +503,12 @@ class ArchitectureConfig(BaseModelConfig):
     # (reusing ``DeepSeekV3TextModel`` unchanged) on runtimes that cannot yet
     # execute ``pkg.nxrt::IndexShare``.
     use_dsa: bool = True
+    # Opt-in export of ``com.microsoft::PagedAttention`` (LATENT / absorbed-MLA
+    # mode) for property-compatible dense MLA. Default off. When off, exports are
+    # byte-identical to the current dense-MLA graph. Eligibility is decided from
+    # semantic geometry (see ``mobius.components._paged_mla``), never model names;
+    # an incompatible geometry raises rather than silently falling back.
+    export_paged_attention: bool = False
     # Per-layer indexer schedule ("full" runs the indexer; "shared" reuses the
     # top-k selection from the closest preceding "full" layer). When the
     # checkpoint config omits this list, it is derived from
--- a/src/mobius/components/__init__.py
+++ b/src/mobius/components/__init__.py
@@ -207,6 +207,24 @@ from mobius.components._multimodal import (
     MLPMultiModalProjector,
 )
 from mobius.components._muse_glimmer_vision import MuseGlimmerVisionModel
+from mobius.components._paged_mla import (
+    PagedCacheState as PagedCacheState,
+)
+from mobius.components._paged_mla import (
+    PagedLatentMLA as PagedLatentMLA,
+)
+from mobius.components._paged_mla import (
+    absorb_mla_weights as absorb_mla_weights,
+)
+from mobius.components._paged_mla import (
+    mla_paged_geometry as mla_paged_geometry,
+)
+from mobius.components._paged_mla import (
+    paged_attention_eligible as paged_attention_eligible,
+)
+from mobius.components._paged_mla import (
+    paged_attention_rejection as paged_attention_rejection,
+)
 from mobius.components._parakeet_audio import ParakeetFastConformerEncoder
 from mobius.components._pixtral_vision import (
     Mistral3MultiModalProjector as Mistral3MultiModalProjector,
--- a/src/mobius/components/_paged_mla.py
+++ b/src/mobius/components/_paged_mla.py
@@ -187,41 +187,51 @@ def paged_attention_rejection(config: ArchitectureConfig) -> str | None:
         _has("index_n_heads")
         or _has("index_head_dim")
         or _has("index_topk")
-        or bool(getattr(config, "indexer_types", None))
+        or bool(config.indexer_types)
     )
-    if getattr(config, "use_dsa", False) and _indexer_configured:
+    if config.use_dsa and _indexer_configured:
         return (
             "PagedAttention LATENT cannot express GLM DeepSeek Sparse Attention "
             "(DSA/IndexShare): query-dependent sparse indices have no operator "
             "input. Export dense MLA (--glm-full-attention) to opt in."
         )
-    if getattr(config, "compress_ratios", None):
+    if config.compress_ratios:
         return (
             "PagedAttention LATENT cannot express DeepSeek-V4 compressed sparse "
             "attention (CSA/HCA): query-dependent compression is not an operator "
             "input."
         )
-    if getattr(config, "o_lora_rank", None) or getattr(config, "o_groups", 1) not in (0, 1):
+    if config.o_lora_rank or config.o_groups not in (0, 1):
         return (
             "PagedAttention LATENT cannot express DeepSeek-V4 grouped/low-rank "
             "output projection (o_groups/o_lora_rank)."
         )
-    if getattr(config, "hc_mult", 1) not in (0, 1):
+    if config.hc_mult not in (0, 1):
         return "PagedAttention LATENT cannot express Hyper-Connections (hc_mult > 1)."
 
     # --- Multi-token prediction is out of scope. ---
-    if getattr(config, "num_nextn_predict_layers", 0) > 0:
+    if config.num_nextn_predict_layers > 0:
         return (
             "PagedAttention LATENT export does not cover Multi-Token Prediction "
             "(num_nextn_predict_layers > 0)."
         )
 
     # --- Optional operator modes not implemented in this slice. ---
-    if getattr(config, "attention_sink", False) or getattr(config, "head_sink", False):
-        return "PagedAttention LATENT slice does not implement the head_sink input."
-    if getattr(config, "use_qk_norm", False) or getattr(config, "qk_layernorm", False):
+    # Per-head QK-norm resolves into the canonical ``attn_qk_norm`` /
+    # ``attn_qk_norm_full`` fields (the HF ``use_qk_norm`` / ``qk_layernorm``
+    # passthrough is consumed there at extraction time), so reject on those.
+    # The dense-MLA targets (DeepSeek-V2/V3, GLM full-attention) leave both
+    # False; a qk-norm MLA would need the operator's q_norm/k_norm inputs, which
+    # this slice does not claim.
+    if config.attn_qk_norm or config.attn_qk_norm_full:
         return "PagedAttention LATENT slice does not implement q_norm/k_norm inputs."
-    window = getattr(config, "sliding_window", None)
+    # Learned attention sinks (the operator's head_sink input) have no
+    # ArchitectureConfig representation for dense MLA: the only sink-bearing MLA
+    # family is DeepSeek-V4, already rejected above by its CSA/HCA fields, and a
+    # non-MLA sink model (e.g. gpt-oss) is rejected by the MLA-geometry check.
+    # PagedLatentMLA structurally never emits a head_sink input, so there is no
+    # config knob to probe here.
+    window = config.sliding_window
     if window is not None and window > 0:
         return (
             "PagedAttention LATENT slice does not implement windowed attention "
--- a/src/mobius/components/_paged_mla.py
+++ b/src/mobius/components/_paged_mla.py
@@ -1,0 +1,492 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+"""Paged, absorbed Multi-head Latent Attention (LATENT ``PagedAttention``).
+
+This module emits the ``com.microsoft::PagedAttention`` v1 operator in its
+``kv_cache_layout="LATENT"`` (absorbed-MLA) mode for *dense* MLA models whose
+geometry is expressible by the operator (DeepSeek-V2/V3, GLM-5.2
+``--glm-full-attention``). It is an opt-in export path
+(``config.export_paged_attention``); the default dense-MLA export
+(:class:`~mobius.components._deepseek_mla.DeepSeekMLA`) is byte-identical when
+the flag is off.
+
+Eligibility is decided purely from semantic geometry (see
+:func:`paged_attention_rejection`), never from model names. Query-dependent
+sparse selection (GLM DSA/IndexShare, DeepSeek-V4 CSA/HCA) is *not* expressible
+because the operator has no query-dependent sparse-index input, so those modes
+are typed-rejected rather than silently falling back to dense.
+
+Absorbed-MLA contract (mirrors the onnx-genai oracle
+``crates/onnx-genai-paged-attention/tests/equivalence.rs``):
+
+* ``l`` = ``kv_lora_rank`` = absorbed content width = ``v_head_size``.
+* ``r`` = ``qk_rope_head_dim`` = decoupled RoPE width.
+* ``head_size = l + r``; RoPE covers the suffix ``[l, l + r)``.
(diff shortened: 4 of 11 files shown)


justinchuby:
## Independent review round 1 — verdict: 1 BLOCKER (fixed), 1 should-fix (fixed), 1 nit

An independent reviewer (not Leon/Sapper) ran the full suite (84 paged/GLM/DeepSeek/CLI tests pass), ruff clean, and cross-checked the emitted `com.microsoft::PagedAttention` v1 schema against the merged native op (`crates/onnx-runtime-ep-cuda/.../paged_attention.rs`, `crates/onnx-genai-paged-attention/src/validate.rs`). Findings addressed in `ee9dbdd`:

**BLOCKER (fixed):** `use_dsa` defaults to `True` on every `ArchitectureConfig` and is only set `False` by `--glm-full-attention` (gated to `glm_moe_dsa`). Plain DeepSeek-V2/V3 never reads `use_dsa` (its text model is always dense), so gating the DSA typed rejection on `use_dsa` alone wrongly rejected real DeepSeek-V3 paged export. DSA is now treated as active only when `use_dsa` is set **and** an indexer is actually configured (`index_n_heads`/`index_head_dim`/`index_topk`/`indexer_types`). Result: DeepSeek-V3 (default `use_dsa=True`, no indexer) → eligible; GLM full-attention (`use_dsa=False`, vestigial indexer) → eligible; GLM DSA active → typed reject. Added a model-level regression proving DeepSeek-V3 with the default `use_dsa=True` still emits paged nodes.

**SHOULD-FIX (fixed):** Tightened the LATENT geometry gate to mirror the native validator's `check_rotary_caches` (`validate.rs:510`): dense MLA always emits `do_rotary=1` with cos/sin caches, so `head_size % 16 == 0`, which given `rotary_dim % 16` forces `kv_lora_rank % 16 == 0`. Closes the gap where an 8-aligned-but-not-16-aligned latent width would pass export yet be rejected at native load. Real GLM/DSV3 (`kv_lora_rank=512`) unaffected.

**NIT (acknowledged):** byte-identical feature-off test is close to tautological by construction (feature default-off); kept as a guard against accidental default flips.

No behavioral change when the feature is off (still byte-identical). No full-size perf claims. Native runtime full-size verification remains a separate gate. **Do not merge** — awaiting Gaff/Roy final approval.

justinchuby:
## APPROVE 3B FOR MERGE

Independently reviewed at exact HEAD `75b5bf99223b31f0bc18e6b6f70df16507a76c6f` (`git rev-parse` confirmed). Read-only; no edits/merge performed.

### Op-contract verification (against onnx-genai `crates/onnx-genai-paged-attention`, current `origin/main`)
- **Input order (17)**: cross-checked `PagedLatentMLA.forward`'s positional `op.PagedAttention(...)` call 0–16 against `PagedAttentionInputs` field order in `params.rs` (excluding the two derived `*_storage_dtype` fields) — **exact match**: query, key, value, key_cache, value_cache, cumulative_sequence_length, past_seqlens, block_table, cos_cache, sin_cache, slot_mapping, head_sink, q_norm_weight, k_norm_weight, k_scale, v_scale, attention_metadata.
- **Outputs (3)**: `output, key_cache_out, _value_cache_out` matches `provider.rs`'s output-1-aliases-input-3 in-place LATENT semantics.
- **Attrs**: `kv_cache_layout="LATENT"`, `kv_num_heads=1`, explicit `scale`, `v_head_size=kv_lora_rank`, `rotary_offset=kv_lora_rank`, `do_rotary=1`, `rotary_interleaved` all line up with `PagedAttentionAttributes` and `validate.rs::check_inputs`'s LATENT-specific branch (value_cache must be absent, kv_num_heads==1, head_sink/q_norm/k_norm absent, v_scale/v_quant/v_cache_dtype unset) — every one of these is honored by the emitted node (all corresponding inputs passed `None`).
- **Geometry gate**: PR's `head_size % 8`, `kv_lora_rank % 8`, `qk_rope_head_dim % 16`(or 0), `rotary_offset % 8` constraints match `validate.rs` exactly (`rotary_dim` is cos_cache-derived, never trusted as an attribute — correctly never emitted). `DEFAULT_PAGED_BLOCK_SIZE=16` matches `onnx-genai-kv::MIN_PAGED_BLOCK_SIZE`. `key_cache` shape `[num_blocks, block_size, 1, head_size]` matches `check_kv_cache`.

### Independently reproduced (not just read)
- Ran the full `_paged_mla_test.py` + `paged_mla_export_test.py` (38 passed) and full `tests/cli_test.py` (46 passed, incl. the 3 new paged-attention CLI tests).
- Ran the entire `mobius/models/` suite (558 passed, 4 skipped, 2 pre-existing unrelated failures — `qwen_image_test.py` golden-file `FileNotFoundError`s, reproduced identically against `main`@`3474c96` when invoked the same way from `src/`; confirmed cwd-relative-path artifact, unrelated to this PR).
- `ruff check` / `ruff format --check` scoped to all 11 touched files: clean.
- Clean merge against `origin/main` (`0665542`, #591 BQMoE fusion) — zero conflicts, zero file overlap.
- Directly executed extra scenarios beyond the PR's own tests to stress the "typed reject, never silent fallback" claim:
  - GLM `use_dsa=True` + flag on → rejects (`ValueError`, DSA/IndexShare message).
  - DeepSeek `compress_ratios=[4,8]` → rejects (CSA/HCA message).
  - DeepSeek quantized cache dtype (`INT8`) → rejects.
  - GLM `use_dsa=False` (full-attention) + flag on → constructs and exports fine, 2 `PagedAttention` nodes, composes cleanly with the model's MoE FFN layers (no BQMoE/#591 overlap — verified by building a real GLM-MoE-DSA-off graph end-to-end, 234 nodes, no crash, no shared op-name collision).
  - `_absorb_paged_mla_weights` / `PagedLatentMLA`'s **untested** `q_proj` (no-lora, e.g. DeepSeek-V2-Lite-style) branch: manually exercised — shapes and construction are correct. Not a bug, but genuinely absent from the PR's own test suite (see findings).
- Numerical parity tests (`TestNumericalParity`) are real: decomposed-vs-absorbed LATENT comparison over full prefill+decode causal-attention progression, rel-err `< 1e-6`, 3 seeds × 2 RoPE-interleave settings + real GLM/DeepSeek head dims (`kv_lora_rank=512, qk_rope=64, qk_nope=128`). This is a legitimate oracle-equivalence proof, not a shape-only check.

### Findings (non-blocking; recommend as explicit follow-ups)
1. **Dead-code rejection branches**: `paged_attention_rejection`'s `head_sink`/`attention_sink`/`use_qk_norm`/`qk_layernorm` `getattr(..., False)` checks reference config attributes that **do not exist anywhere** in `ArchitectureConfig` or any model in this repo (grep-verified) and have **zero test coverage**. This is currently harmless: DeepSeek-V4 (which has an unconditional, always-present `attn_sink` parameter in `DeepSeekV4Attention.__init__`) is already reliably excluded by three independent, *real*, tested signals — `assert config.o_lora_rank is not None` (a hard, unconditional assertion for V4 construction to succeed at all), `config.sliding_window` (V4-Flash always sets this per the code's own comment), and `compress_ratios` (non-empty list, even all-zero, is still truthy). But the PR body/docstring's claim that "head_sink, qk-norm... raise a typed reason" is currently aspirational, not verified. Recommend either wiring these to a real field once one exists, or removing/annotating the dead branches so the claim isn't overstated.
2. **`q_proj` (no-lora Q) path is untested** in both `_paged_mla_test.py` and `paged_mla_export_test.py` — only the `q_lora_rank`-set (DeepSeek-V3/GLM-5.2) branch is exercised. Manually verified correct, but should get a real test if DeepSeek-V2-Lite-style (no q-lora) checkpoints are in scope.
3. `--features paged-attention --runtime ort-genai` is unguarded in `_validate_ort_genai_compatibility` (would silently emit a `genai_config.json` using `past_key_names`/`past_value_names` templates that don't match the paged cache I/O names). Low practical risk — this targets Microsoft's separate `onnxruntime-genai` consumer, not the onnx-genai native Rust runtime this slice is built for, and the PR makes no claim of `--runtime ort-genai` support.
4. Cosmetic: a duplicated, unreachable `l % 8 != 0` check in `paged_attention_rejection` (harmless dead branch); the `"batch + 1"` dim-param string for `cumulative_sequence_length` is an unrelated symbolic-dim label to `batch` (fine for ONNX, just not textually correlated).

None of the above affect correctness for the PR's stated target configurations (DeepSeek-V2/V3 dense MLA with q-lora, GLM-5.2 `--glm-full-attention`), and the default-off flag means zero blast radius for anyone not opting in.

### Verdict
**APPROVE 3B FOR MERGE.** Property-based eligibility (verified non-model-name-gated via the DeepSeek-V3-passes-without-name-check test + my own repro), exact op contract, byte-identical feature-off, real typed rejection for every currently-constructible incompatible model (GLM DSA, DeepSeek CSA/HCA/o-groups/hc_mult/MTP/quantized-cache/sliding-window), single cache authority (Mobius allocates no pages; `PagedCacheState` is 100% caller-owned), and no #591/BQMoE overlap are all independently confirmed, not just trusted.

**Default-off seam may merge before full-size A/B.** Exact next full-size-GLM/DeepSeek gate before any full-size checkpoint export+correctness claim is admissible:
1. A real end-to-end functional test that loads a full-size (or shape-faithful reduced) DeepSeek-V3/GLM-5.2 checkpoint, exports with `--features paged-attention`, and executes the graph under **ORT ≥ 1.29 CUDA** (not just structural/attr assertions — actual kernel execution) with output parity vs. the dense-MLA baseline.
2. A native onnx-genai-runtime-side integration test that consumes an actual Mobius-exported paged-attention graph end-to-end (today the Rust-side oracle/#1978 kernel tests and this PR's Mobius-side tests are verified independently of each other, not cross-validated against a shared real artifact).
3. Close finding #2 above (q_proj/no-lora coverage) if DeepSeek-V2-Lite-style checkpoints are in scope for full-size validation.
4. No full-size perf claim until measured per `cuda-perf-measurement`/`measurement-discipline` skill discipline — this PR correctly makes none.

justinchuby:
## Follow-up: no-backward-compat honesty cleanup (`8a17cfb`)

Per the no-BC-during-development directive, tightened the eligibility validator with clean breaking access (no compat shims), keeping correctness / property gates / default-off / typed refusals mandatory. All in-repo callers/fixtures/tests updated atomically.

**Direct declared-field access.** Every field `paged_attention_rejection` probes is a declared `ArchitectureConfig` field, so the defensive `getattr(config, ..., default)` calls are replaced with direct attribute access (`use_dsa`, `indexer_types`, `compress_ratios`, `o_lora_rank`, `o_groups`, `hc_mult`, `num_nextn_predict_layers`, `sliding_window`), plus the `export_paged_attention` reads in `deepseek.py`/`glm_moe_dsa.py`. Behavior is identical — each `getattr` literal default matched the declared default. (`_has(name)` stays: it is a by-name "present and positive" helper for the `int | None` geometry fields, not a shim.)

**Correctness fix — real qk-norm refusal.** Per-head QK-norm is consumed into the canonical `attn_qk_norm` / `attn_qk_norm_full` fields at config extraction (the HF `use_qk_norm` / `qk_layernorm` passthrough is **not** an `ArchitectureConfig` field). The prior check probed those always-absent names, so it was **unreachable** — a qk-norm MLA would have slipped through and silently missed the operator's `q_norm`/`k_norm` inputs. Now rejects on the real fields; DeepSeek-V2/V3 and GLM full-attention leave both `False`, so targets are unaffected. Reachable + tested at component and model level.

**Removed a provably-dead branch.** `head_sink` / `attention_sink` are never `ArchitectureConfig` fields, so that guard was always `False`. Coverage is subsumed and documented: the only sink-bearing MLA family (DeepSeek-V4) is already rejected by its CSA/HCA fields, a non-MLA sink model (gpt-oss) is rejected by the MLA-geometry check, and `PagedLatentMLA` structurally never emits a `head_sink` input. No silent-miscompute path is introduced.

**Tests:** added `test_qk_norm_rejected` (`attn_qk_norm` / `attn_qk_norm_full`) and `test_window_rejected` component tests, plus a model-level qk-norm typed-reject case; the DeepSeek-V3 default-`use_dsa=True` regression now asserts directly. Full paged/GLM/DeepSeek/V4/tasks/CLI suites pass (247 across the sweeps); ruff 0.16.2 clean. Default-off remains byte-identical.

Still a DRAFT — **do not merge**; awaiting Gaff or Roy final approval.

justinchuby:
## Roy — Final re-review of PR #599 at HEAD `8a17cfb4d668104ea92b52c3a9e6d98957905fde`

Scope: delta `75b5bf9..8a17cfb` (the independent-review fixup: `ee9dbdd` DSA/geometry fix + `8a17cfb` honesty cleanup) plus integration against latest `origin/main`. My prior `75b5bf9` approval stands corrected — the independent reviewer caught a real functional bug my own repro had masked. Verified all 5 requested points independently, not by trusting the PR's own tests.

### 1. Direct `ArchitectureConfig` field access — VALID everywhere
`ArchitectureConfig` (`_configs/_base.py:396`) is a single, non-subclassed `@dataclass` used uniformly by every model. Every field now accessed directly (`sliding_window`, `attn_qk_norm`, `attn_qk_norm_full`, `o_groups`, `o_lora_rank`, `compress_ratios`, `hc_mult`, `num_nextn_predict_layers`, `use_dsa`, `export_paged_attention`, `indexer_types`) is a genuinely declared field with a default. Only remaining `getattr` is inside `_has()`, a justified DRY helper for the 7 int|None MLA-geometry/indexer fields (all real declared fields) — not a defensive shim. No `AttributeError` risk for any model_type.

### 2. qk-norm refusal — reachable, correct, no false-reject
Old check probed `getattr(config, "use_qk_norm", False)` — a literal field that never exists post-extraction on `ArchitectureConfig`, i.e. permanently dead/unreachable code. Fix reads the real canonical `attn_qk_norm` / `attn_qk_norm_full` fields, which are derived at config-extraction time (`_base.py:~832-852`) from a `model_type` allowlist (gemma3/qwen3/olmo families) or raw HF `use_qk_norm` passthrough. Neither DeepSeek nor GLM model_types appear in either allowlist and neither sets a raw `use_qk_norm` — both fields correctly default `False` for PR targets. No false-reject for DeepSeek-V2/V3/GLM-5.2; check is now genuinely reachable for any checkpoint that does set qk-norm.

### 3. Dead `head_sink` branch removal — coverage preserved
Removal is safe: DeepSeek-V4 (the only model_type where `head_sink`/attention-sink semantics could matter) is already independently excluded from paged-MLA eligibility via its CSA/HCA fields (established in the earlier Cycle-1 review), and `PagedLatentMLA` structurally never emits a `head_sink` input regardless. No coverage gap opened by the removal.

### 4. head_size % 16 fix — independently confirmed against native source
Verified directly against `crates/onnx-genai-paged-attention/src/validate.rs::check_rotary_caches` (onnx-genai repo, line 510): `head_size % 16 != 0` is the real, exact native-loader constraint — not `% 8`. The old `% 8` check would have let an 8-aligned-but-not-16-aligned `kv_lora_rank` export successfully from Mobius, then fail at native ORT load — a real silent-broken-artifact risk, now closed correctly.

### 5. New/updated tests — non-vacuous
- `test_deepseek_v3_default_use_dsa_is_eligible`: proves `use_dsa=True` (true default) + no indexer configured → eligible. This is the exact regression guard for the BLOCKER.
- `test_dsa_indexshare_rejected`: updated to also set indexer fields (since `use_dsa` alone no longer triggers rejection) — still correctly rejects when DSA is *actually* active.
- `test_qk_norm_rejected`: independently tests `attn_qk_norm=True` and `attn_qk_norm_full=True`, plus a baseline-eligible case.
- `test_window_rejected`: tests `sliding_window=64` rejects and `sliding_window=None` doesn't.
- `test_geometry_constraints_rejected`: updated to the new 16-boundary (`kv_lora_rank=8, qk_rope_head_dim=16` → head_size=24, correctly rejected as the exact 8-aligned-not-16-aligned edge case).
- `paged_mla_export_test.py::_deepseek_config` — critically, the explicit `use_dsa=False` test override was **removed**; `test_deepseek_v3_emits_paged_not_name_gated` now asserts `config.use_dsa is True` before checking paged nodes emit, directly exercising the true production default rather than an artificially-forced config.

Each new/changed test has both a rejecting case and a non-rejecting baseline/None case — not vacuous.

### 6. Feature-off / no regression
Feature-off code paths untouched by this delta. Full regression run:
- `mobius/components/_paged_mla_test.py` + `mobius/models/paged_mla_export_test.py`: **41/41 passed**
- `mobius/models/deepseek_test.py`, `deepseek_v4_test.py`, `deepseek_v4_flash_test.py`, `glm_moe_dsa_test.py`, `mobius/tasks/`: **163/163 passed**
- `tests/cli_test.py`: **46/46 passed**
- Full `mobius/models/` suite: **558 passed, 2 failed, 4 skipped**. The 2 failures (`qwen_image_test.py::test_edit_vae_matches_diffusers_on_real_source_image`, `::test_deterministic_l4_l5_image_edit_golden`) are pre-existing `FileNotFoundError`s for missing local testdata assets in an unrelated diffusion image-editing model, not touched by this PR in any way — confirmed unrelated, and the 558-pass count matches the prior baseline exactly.
- `ruff check` + `ruff format --check` on all 5 touched files: **clean**.
- Clean merge against latest `origin/main` (includes `#578`, `#581`, `#576`) — zero conflicts, re-verified.

### Verdict: **APPROVE FINAL HEAD**

All four requested verification points are independently confirmed correct, the fix genuinely closes the missed BLOCKER (production DeepSeek-V3 configs with the true `use_dsa=True` default would previously have been wrongly rejected), the should-fix (16-alignment) is exactly correct per native source, the honesty cleanup introduces no new risk, and no regressions were found across 808 total tests run plus lint/format. PR #599 is approved for merge at `8a17cfb4d668104ea92b52c3a9e6d98957905fde`.

— Roy

PR #622, opened 2026-08-25
fix(deepseek-v4): per-layer CSA/HCA compressed-record symbolic dim

justinchuby:
## Problem

The native `pkg.nxrt::CompressedSparseAttention` exporter (#593) stamped a
**single shared** symbolic dim (`past_compressed_records` /
`present_compressed_records`) on **every** CSA layer's compressed-record axis,
on the assumption that "every CSA layer advances its cache together."

That assumption is false for the official DeepSeek-V4-Flash **interleaved
schedule** (21 ratio-4 CSA + 20 ratio-128 HCA). A ratio-4 layer pools one
compressed record per ~4 tokens; a ratio-128 layer pools one per ~128. Their
record counts diverge as the sequence grows, so a shared symbol forces ORT to
bind the same dim to two different sizes and it (correctly) rejects the graph:

```
symbol component.model.past_compressed_records bound to conflicting sizes
2 and 0 across bound inputs
```

This makes any exported graph with a *mixed* ratio schedule unloadable — the
exact schedule the real checkpoint uses.

## Fix

Give each layer its own record-axis symbolic dim via new `CsaLayerPlan`
properties:

- `past_compressed_records.{layer_id}`
- `present_compressed_records.{layer_id}`
- `selected_records.{layer_id}`

**Within** a layer, the attention cache and the learned-index cache still share
that one per-layer symbol — they advance in lockstep at the same ratio, which is
a real invariant worth expressing. **Across** layers of different ratios the
axes no longer alias.

## Tests

- `test_native_csa_emits_both_ratios_for_interleaved_schedule` — extended to
  assert the ratio-4 and ratio-128 layers carry **distinct** record axes, and
  that within the ratio-4 layer the attention/index caches share their layer's
  axis.
- Ratio-128 IO test — updated to the per-layer symbol name.
- Full `deepseek_v4_flash_test.py` (45 tests) passes; `ruff format --check` and
  `ruff check` clean on the touched files.

No BC shim (dev-time API change, per the no-backward-compat directive).

## Downstream

This is the prerequisite for loading a mixed-ratio CSA/HCA graph in onnx-genai's
native decode runtime; the companion onnx-genai E2E proof (stacked on #2063)
regenerates its tiny fixture from this fix.

⚠️ DRAFT — stopping for independent review; do not merge.

### DIFF
--- a/src/mobius/models/_deepseek_v4_csa.py
+++ b/src/mobius/models/_deepseek_v4_csa.py
@@ -284,6 +284,26 @@ class CsaLayerPlan:
     def selected_indices_name(self) -> str:
         return f"selected_indices.{self.layer_id}"
 
+    # -- dynamic record-axis symbolic dim names ---------------------------
+    # Each layer's compressed-record axis is a *per-layer* symbolic dim. A
+    # mixed schedule pools at different ratios per layer (ratio-4 keeps one
+    # record per 4 tokens, ratio-128 one per 128), so their record counts
+    # diverge and MUST NOT share one symbol -- ORT rejects binding the same
+    # symbolic dim to two sizes. Within a layer the attention cache and the
+    # learned-index cache advance in lockstep (same ratio), so they share this
+    # one per-layer symbol; that lockstep constraint is real and worth stating.
+    @property
+    def past_records_axis_name(self) -> str:
+        return f"past_compressed_records.{self.layer_id}"
+
+    @property
+    def present_records_axis_name(self) -> str:
+        return f"present_compressed_records.{self.layer_id}"
+
+    @property
+    def selected_records_axis_name(self) -> str:
+        return f"selected_records.{self.layer_id}"
+
 
 def _layer_compress_ratio(config: ArchitectureConfig, layer_id: int) -> int:
     ratios = config.compress_ratios or []
--- a/src/mobius/models/deepseek_v4_flash_test.py
+++ b/src/mobius/models/deepseek_v4_flash_test.py
@@ -880,8 +880,8 @@ def test_native_csa_threads_compressed_state_io():
     pres_kv = outputs["present_compressed_kv.1"]
     assert past_kv.dtype == ir.DataType.FLOAT
     assert pres_kv.dtype == ir.DataType.FLOAT
-    assert [str(d) for d in past_kv.shape] == ["batch", "past_compressed_records", "16"]
-    assert [str(d) for d in pres_kv.shape] == ["batch", "present_compressed_records", "16"]
+    assert [str(d) for d in past_kv.shape] == ["batch", "past_compressed_records.1", "16"]
+    assert [str(d) for d in pres_kv.shape] == ["batch", "present_compressed_records.1", "16"]
 
     past_carry = inputs["past_compression_carry.1"]
     pres_carry = outputs["present_compression_carry.1"]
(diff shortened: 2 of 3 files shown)


justinchuby:
## Review verdict — **APPROVE** (numerics/correctness) · Chew (Code Reviewer, Numerics)

Reviewed HEAD `3e7368010189344b96fcfafd2e533d650f2e671c` (1 commit, 3 files, **+53/-11**; base `main`, stacked on #593/#599/#602…). Review-only — no edit, no merge.

### What it fixes
#593 stamped ONE shared symbolic dim on the record axis of *every* CSA/HCA layer. On the interleaved ratio-4 + ratio-128 schedule the record counts diverge, so ORT rejects binding one symbol to two sizes (*"bound to conflicting sizes 2 and 0"*). The PR moves symbol creation from **outside → inside** the per-layer loop and derives per-layer names `{past,present}_compressed_records.{layer_id}` / `selected_records.{layer_id}` via three new `CsaLayerPlan` properties. Within a layer the attention cache and learned-index cache still **share** the one per-layer symbol (lockstep preserved); across ratios they no longer alias.

### Gate results
1. **Per-layer record axis / no invalid sharing** ✅ — verified in `_compressed_inputs`/`_compressed_outputs` (symbols now created per-layer inside the loop), in the regression test, in an IR export, and in the **serialized ONNX proto** (`dim_param` `past_compressed_records.1` vs `.2`).
2. **Shape/dtype/name-pairing/metadata vs `CsaStateRole` contract** ✅ — the PR changes only the symbolic-dim **string**; dtypes, ranks, tensor names (`.{layer}`), roles are unchanged. The runtime contract is role-typed and *"never names a model, a layer, or a tensor spelling"* — the engine binds by `CsaStateRole` + runtime record **size**. The string `past_compressed_records` appears nowhere in the runtime; the tiny-CSA E2E fixture binds per-layer names and treats `records == 0` at prefill as a runtime count. ⇒ the rename is **invisible to the runtime** and cannot break the contract.
3. **No loss of useful within-layer equality** ✅ — attention + index caches share the same per-layer symbol (`r4_kv_axis == r4_index_key_axis == past_compressed_records.1`; reproduced at IR + proto). `selected_records.{layer}` is a **distinct** per-layer symbol for the top-k width `min(records, topk)` and correctly does **not** alias the record count. past ≠ present; carries stay records-independent `[batch, slots, planes, width]`.
4. **Checker / shape inference / load** ✅ — **45/45** `deepseek_v4_flash_test.py` pass locally; `ruff format --check` + `ruff check` clean; **strict** `onnx.shape_inference.infer_shapes(strict_mode=True, data_prop=True)` passes on an exported tiny **mixed** ratio4/ratio128 model. (Plain `onnx.checker` on the unit-test structural graph trips an unrelated topo-sort/placeholder-weight artifact; the nxrt custom op is not ORT-loadable by design — the authoritative load+generate proof is the onnx-genai runtime E2E and CI L1 Smoke / L3 Synthetic Parity.)
5. **No unrelated changes** ✅ — exactly 1 commit, 3 files (2 exporter + 1 test).
6. **CI** — unit-test matrix (ubuntu/windows × 3.11/3.12/3.13), Lint, lintrunner, CodeQL, Build, Architecture Diff, Benchmark, **L1 Smoke**, **L3 Synthetic Parity** = SUCCESS.

_Bonus:_ the old code also shared `selected_records` globally across layers (a second latent aliasing bug); the PR correctly makes it per-layer too.

### Non-blocking conditions (not defects)
- **Integration (fast)**, **L4 Golden Comparison**, and **L5 Generation E2E** are still **QUEUED** — these are the strongest mixed-model *load + generate* proof and the exact place the old "conflicting sizes" bug surfaced. PR is DRAFT / mergeState BLOCKED; final merge should await them going green. The APPROVE is on the numerics/correctness of the change itself, which is complete and minimal.

### Scope
Unblocks **correct export of the mixed ratio-4/ratio-128 DeepSeek-V4-Flash CSA/HCA compressed-attention state** (required by onnx-genai #2063 E2E). It does **not** touch or enable **routed top-k MoE** — block-FP8 / planar-FP4 expert weights still typed-reject at the runtime-capability gate (`test_native_csa_full_export_typed_rejects_at_runtime_capability_gate`, `test_preprocess_weights_enforces_runtime_capability_gate`). Consistent with my #2088 verdict (planar matmul primitive only; BQMoE stays typed-reject).

— Chew · numerics/precision review · reproduced on host (no GPU needed)

PR #666, opened 2026-08-27
Add typed checkpoint weight records

xiaoyu-work:
## Summary

- add typed `WeightRecord`, `PackedWeight`, and component-owned `WeightBundle` abstractions
- group Olive and GPTQ/AWQ sidecars into one logical weight with completeness checks
- add a codec registry keyed by `quant_method`
- expose compatibility normalization through the existing proven packers
- leave the active checkpoint loader unchanged in this foundational PR

## Stack

- Design: #664
- Depends on: #665
- **This PR: typed weight pipeline foundation**
- Next: per-component checkpoint loader
- Final: model adapter migration

## Validation

- 10 record/codec unit tests
- 49 codec and existing Olive/GPTQ/AWQ compatibility tests

PR #670, opened 2026-08-27
Migrate composite model weight adapters

xiaoyu-work:
## Summary

- add an explicit model-weight adapter boundary while retaining legacy `preprocess_weights` compatibility
- make Qwen3.5/QMoE graph construction and expert packing consume the same decoder component plan
- make Gemma4 decoder, vision, audio, and embedding construction component-aware
- preserve clipped/scaled quantized subclasses and add explicit HF/local aliases for Gemma4 media modules
- route Gemma4 packed token tables only to the embedding graph
- hand T5 and Qwen35 sibling-component sidecars to the generic component codec without double preprocessing
- fail publication when any MatMulNBits/GatherBlockQuantized parameter remains unbound

## Stack

- Design: #664
- Component manifest: #665
- Typed weight pipeline: #666
- Depends on loader: #669
- **This PR: model adapter migration and binding gate**

## Validation

- 214 final adapter, model, codec, and component-matrix tests
- 54 Transformers loader/config resolver tests
- plan-only Qwen35 QMoE graph/weight alignment regression
- Gemma4 decoder regex exclusions, vision `.linear` alias, per-component layouts, sidecar routing, and quantized embedding regressions


xiaoyu-work:
@copilot resolve the merge conflicts in this pull request

xiaoyu-work:
Superseded by #733, which consolidates #666, #669, and #670 into one PR rebuilt directly on the latest main. The original branch is retained.

PR #671, opened 2026-08-27
Tied quantized lm_head fix

neilmsft:
## Description of bug: 

Gemma ties  lm_head  to  embed_tokens , so the GGUF has no  output.weight. But  Gemma4CausalLMModel  built a standalone quantized head anyway, declaring 3 initializers nothing could fill, which crashed the export and forced every build to silently fall back to fp32 embeddings.

## Fix
Use TiedQuantizedLMHead in the text-only path when the weights are tied and both sides are quantized, matching the base class. The multimodal path is untouched. Its decoder and embedding live in separate ONNX graphs and genuinely cannot share a table

### DIFF
--- a/src/mobius/models/gemma4.py
+++ b/src/mobius/models/gemma4.py
@@ -43,6 +43,7 @@ from mobius.components import (
     Linear,
     QuantizedEmbedding,
     RMSNorm,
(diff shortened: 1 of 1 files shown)


titaiwangms:
@neilmsft Do we still need this?

PR #719, opened 2026-09-01
Add native GPT-OSS MXFP4 QMoE export

titaiwangms:
## Summary

- preserve official GPT-OSS MXFP4 expert weights instead of dequantizing and requantizing them
- losslessly repack GPT-OSS `[E, N, K/32, 16]` blocks into the native QMoE `[E, K, N/2]` layout
- emit `com.microsoft::QMoE` with GPT-OSS routing, expert biases, global scales, and SwiGLU semantics
- stream indexed safetensors one expert projection at a time so GPT-OSS 120B does not require an eager second copy of all expert weights
- keep integer-affine GPTQ/AWQ/Olive QMoE and unquantized GPT-OSS paths unchanged

## Runtime requirements

Native GPT-OSS MXFP4 export currently requires:

- the CUDA execution provider
- FP16 or BF16 activations
- an ONNX Runtime build with FP4 QMoE enabled using CUDA 12.8 or newer

Unsupported execution providers and dtypes fail before graph construction or weight I/O.

## Draft validation remaining

The implementation blockers are resolved. Before marking this PR ready, the
remaining work is runtime qualification with official checkpoints:

- export an official GPT-OSS 20B and 120B checkpoint and record peak host
  memory, temporary disk use, output size, and wall time
- run prefill and cached decode on A100 with a pinned FP4-QMoE-enabled ONNX
  Runtime build
- characterize both the SM80 fused path and forced full-dequant fallback,
  including numerical parity and peak GPU memory

## Validation

- native nibble repacking is byte-exact across chunk boundaries
- E8M0 bytes `0x00` through `0xfe` are preserved and `0xff` is rejected
- cross-shard block/scale pairs are streamed and transformed without invoking the eager checkpoint loader
- graph construction emits the expected FP4 QMoE inputs, attributes, routing, biases, and global scales
- existing integer QMoE, generic streaming, configuration, builder, and unquantized GPT-OSS tests pass
- the resulting FP4 QMoE contract was exercised against ONNX Runtime CUDA with numerical parity to the GPT-OSS reference computation


src/mobius/_ir_utils.py
```diff
@@ -0,0 +1,84 @@
...
+from __future__ import annotations
+
+from collections.abc import Iterator
+from itertools import chain
+
+import onnx_ir as ir
+
+
+def _iter_type_dtypes(
+    type_proto: ir.TypeProtocol | ir.MapTypeProtocol | ir.DataType | None,
```
justinchuby: It's possible to just call `.dtype`

src/mobius/_ir_utils.py
```diff
@@ -0,0 +1,84 @@
...
+    """Yield leaf tensor dtypes from a possibly nested ONNX type."""
+    if type_proto is None:
+        return
+    if isinstance(type_proto, ir.DataType):
+        yield type_proto
+        return
+    if isinstance(type_proto, ir.TypeProtocol):
+        yield from _iter_type_dtypes(type_proto.elem_type)
+        return
+    if isinstance(type_proto, ir.MapTypeProtocol):
```
justinchuby: I don't think the map type is used?

src/mobius/_ir_utils.py
```diff
@@ -0,0 +1,84 @@
...
+                elif attribute.type == ir.AttributeType.TYPE_PROTOS:
+                    for type_proto in attribute.value:
+                        yield from _iter_type_dtypes(type_proto.type)
+
+
+def minimum_ir_version(graph: ir.Graph, *, baseline: int = 11) -> int:
+    """Return the minimum ONNX IR version required by tensor types in *graph*."""
+    dtype_floors = {
+        ir.DataType.FLOAT4E2M1: 11,
+        ir.DataType.FLOAT8E8M0: 12,
```
justinchuby: Feel free to bump IR version universally.

src/mobius/_model_package_test.py
```diff
@@ -14,6 +14,7 @@
 import pytest
 import torch
 
+import mobius._model_package as model_package_module
```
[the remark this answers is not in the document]
titaiwangms: Resolved in 2b39e011. The test now uses a single from-import style while retaining the private save helper needed by the monkeypatch/concurrency coverage.

src/mobius/_model_package_test.py
```diff
@@ -691,6 +700,91 @@ def test_save_creates_files(self, tmp_path):
...
+            if path == "first.onnx":
+                first_save_entered.set()
+                assert release_first_save.wait(timeout=5)
+            else:
+                second_save_entered.set()
+
+        def save(path):
+            try:
+                model_package_module._save_safetensors(_make_e8m0_model(), path)
+            except BaseException as error:
```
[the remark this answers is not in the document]
titaiwangms: Resolved in 2b39e011 by narrowing the worker exception handler from BaseException to Exception.

src/mobius/_ir_utils.py
titaiwangms: Resolved in 2b39e011 by removing the generic IR/type visitor entirely and adopting universal IR version 12 for task-built models.

src/mobius/_ir_utils.py
titaiwangms: Resolved in 2b39e011. The unused map-type traversal was removed together with the generic IR utility.

src/mobius/_ir_utils.py
titaiwangms: Resolved in 2b39e011. Task model construction now emits IR version 12 universally, as suggested.

src/mobius/models/gptoss_test.py
```diff
@@ -1,92 +1,891 @@
...
+    total_length = past_length + query_length
+    num_heads = 2
+    num_kv_heads = 1
+    head_dim = 8
+    tensor = onnx.helper.make_tensor_value_info
+    inputs = [
+        tensor("query", onnx.TensorProto.FLOAT16, [batch, query_length, 16]),
+        tensor("key", onnx.TensorProto.FLOAT16, [batch, query_length, 8]),
+        tensor("value", onnx.TensorProto.FLOAT16, [batch, query_length, 8]),
+        tensor(
```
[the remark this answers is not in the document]
titaiwangms: Resolved in 2b39e011. The focused GQA kernel parity model is now built with onnx_ir, saved through ir.save to a temporary ONNX file, checked through the same serialization path, and loaded by ORT from that file. Invalid GQA construction/session failures are no longer broadly converted to skips.

src/mobius/_model_package.py
```diff
@@ -157,6 +195,78 @@ def _validate_mtp_chain(package: ModelPackage) -> None:
...
+def _validate_streaming_safetensors_destination(
+    package: ModelPackage,
+    directory: str,
+    *,
+    external_data: str,
+    component_layout: _ComponentSaveLayout,
+) -> None:
+    """Reject model output directories that could overwrite lazy source shards."""
+    if external_data != "safetensors":
+        return
```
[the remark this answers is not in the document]
titaiwangms: Resolved by the current implementation in 2b39e011. Directory overlap remains safetensors-specific, but exact generated output files are checked for every external-data format using lexists/samefile, including ONNX model/data shards and package reports. Regression tests cover hard links and symlinks for ONNX and safetensors outputs.

src/mobius/models/gptoss.py
```diff
@@ -118,14 +391,137 @@ def __init__(self, config: ArchitectureConfig):
...
+
+        router_probs, _, normalize, _ = _realize_gate_and_get_qmoe_routing(
+            op, self.gate, hidden_states
+        )
+
+        return op.QMoE(
+            hidden_states,
+            _flatten_to_2d(op, router_probs),
+            self.fc1_experts_weights,
+            self.fc1_scales,
```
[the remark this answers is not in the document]
titaiwangms: Resolved in d14fc2a1. Added a CUDA-gated reduced-fixture parity test that executes the production-emitted native FP4 QMoE after checkpoint-format blocks/scales pass through the real preprocessing/repacking path. The independent NumPy reference decodes checkpoint E2M1/E8M0 directly and verifies top-1 expert dispatch, router weight+bias, clipped SwiGLU, and expert biases for prefill- and decode-shaped inputs. Both cases execute and pass on A100 with the SM80 fallback.

src/mobius/integrations/ort_genai/ep_config.py
```diff
@@ -67,6 +67,9 @@ def make_provider_options(
     if ep == "webgpu":
         options["enableGraphCapture"] = "1" if graph_capture else "0"
         options["validationMode"] = "disabled" if graph_capture else "basic"
+    elif ep == "dml" and enable_graph_capture is False:
+        # ORT GenAI enables DML graph capture when this option is omitted.
```
justinchuby: Do we still do dml?

src/mobius/integrations/ort_genai/auto_export.py
```diff
@@ -2356,8 +2354,6 @@ def auto_export(
     """
     from mobius.integrations.transformers import build
 
-    os.makedirs(output_dir, exist_ok=True)
```
justinchuby, line 2359: Could you explain this change?

titaiwangms: `ModelPackage.save()` now owns output-directory creation. Streaming packages are written to a sibling staging directory and published only after serialization succeeds; pre-creating `output_dir` here causes the transactional path to reject the destination. Non-transactional saves are unchanged because `ModelPackage.save()` creates the directory itself.

src/mobius/integrations/ort_genai/ep_config.py
titaiwangms: DML is still present in Mobius, but upstream has moved DirectML to sustained engineering and new feature development to WinML. This DML-specific graph-capture handling is unrelated to GPT-OSS, so I removed it and its added expectations in 7312cda2 while leaving the existing DML support unchanged.

src/mobius/integrations/transformers/_gptoss_weights.py
```diff
@@ -0,0 +1,330 @@
...
+) -> dict[str, object]:
+    """Bind native MXFP4 weights without ever constructing a checkpoint dict."""
+    if len(package) != 1:
+        raise ValueError(
+            "Native GPT-OSS MXFP4 streaming requires a single text model component."
+        )
+
+    try:
+        local = _local_weight_paths(pathlib.Path(model_id))
+    except FileNotFoundError as exc:
```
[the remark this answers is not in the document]
titaiwangms: Fixed in f6b1e95b (included in branch tip d87bff4b). `_local_weight_paths()` now preserves the already-validated lexical path under the snapshot instead of resolving the shard symlink before loading. The GPT-OSS adapter separately registers both the snapshot parents and resolved blob parents/files for save-collision protection. I also added an end-to-end sharded local snapshot test that binds and materializes the lazy initializers, plus Windows absolute and drive-relative path rejection coverage.

src/mobius/integrations/ort_genai/auto_export_test.py
```diff
@@ -2939,6 +2940,22 @@ def fake_save(self, directory, **kwargs):
         assert save_calls[0]["external_data"] == "safetensors"
         assert save_calls[0]["progress_bar"] is False
 
+    def test_streaming_package_can_transactionally_create_output(self, tmp_path):
+        from mobius.integrations.ort_genai.auto_export import export_package
+
+        pkg = self._make_pkg()
+        pkg.weight_loading_report = {
```
justinchuby, line 2947: What does the report do?

src/mobius/integrations/transformers/_builder.py
```diff
@@ -67,6 +69,59 @@ def _reject_unsupported_affine_qwen4(model_type: str, config: object) -> None:
     )
 
 
+def _is_native_gptoss_mxfp4(config) -> bool:
+    quantization = getattr(config, "quantization", None)
+    return bool(
+        getattr(config, "model_type", None) == "gpt_oss"
```
justinchuby, line 75: At some point we may move model specific logic to a different place, I think.

justinchuby (approved):

src/mobius/integrations/ort_genai/auto_export_test.py
titaiwangms: It serves two purposes here. First, it is persisted as `weight-loading-report.json` to record how the checkpoint weights were mapped/materialized (including the streaming/native-storage contract). Second, `streaming_external_data: true` tells `ModelPackage.save()` that lazy initializers will read source shards during serialization, so the package is published through a sibling staging directory and atomically renamed only after the complete save succeeds. This test sets the minimum report shape needed to exercise that transactional `export_package()` path.

src/mobius/integrations/transformers/_builder.py
titaiwangms: Agreed. This is currently kept next to the build orchestration because it freezes the source-checkpoint policy before dtype/config rewriting and before any weight I/O. Once more architectures need model-specific source policies, this should move behind a model/adapter-owned hook or registry rather than adding more branches to the generic builder.

PR #725, opened 2026-09-04
Emit current shape-only ONNX GenAI tensor contracts

justinchuby:
## Summary
- remove retired tensor-contract `rank` emission from the shared ONNX GenAI metadata builders and serializer
- retain exact scalar `shape: []`, named symbolic dimensions, and independent `Any` dynamic dimensions without inventing fixed extents
- retire the obsolete workflow capability list required by the current consumer
- refresh all 15 producer-conformance package snapshots
- add current TensorContract schema coverage across every signal package and stale-fixture guards

Fixes the producer failure reported by onnx-genai signal run https://github.com/justinchuby/onnx-genai/actions/runs/33886695719.

## Validation
- generated and snapshot-compared all 15 packages
- current onnx-genai Rust validator: 15/15 valid
- ONNX GenAI metadata suite: 722 passed, 3 skipped
- Sensenova/static-cache related suite: 50 passed
- Ruff 0.16.2 lint and format checks passed
- `git diff --check`

A broader local non-integration run also reproduced three unrelated failures unchanged on Mobius `origin/main` (GGUF evidence expectation, Qwen Image diffusers API drift, and ModernBERT synthetic tolerance).

### DIFF
--- a/src/mobius/integrations/onnx_genai/_metadata_io.py
+++ b/src/mobius/integrations/onnx_genai/_metadata_io.py
@@ -5,6 +5,7 @@
 
 from __future__ import annotations
 
+import copy
 from typing import Any
 
 import yaml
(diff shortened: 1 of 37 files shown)


justinchuby:
Closing per project decision to pause cross-repository producer coordination and make the onnx-genai Mobius conformance signal manual-only for now. The implementation remains available in this PR if producer conformance is resumed later.

PR #742, opened 2026-09-18
Add Qwen2.5 Omni Talker export

rui-ren:
Summary
- Add independent Thinker and Talker configs.
- Add codec embedding plus Talker decoder/KV export.
- Retain Thinker-only behavior.

Validation
- compileall and git diff --check passed.
- Focused pytest was blocked by TLS dependency bootstrap.


gramalingam:
## Findings

### [P2] Honor the checkpoint's audio-output disable flag

**File:** `src/mobius/_configs/_base.py`  
**Lines:** 1511-1516  
**Confidence:** High

The Talker is enabled whenever `talker_config` exists, but Transformers instantiates the Qwen2.5-Omni Talker only when the parent configuration's `enable_audio_output` flag is true. A checkpoint can retain `talker_config` while explicitly disabling audio output.

Exporting such a Thinker-only or fine-tuned checkpoint currently creates `talker_embedding` and `talker` graphs anyway. If Talker weights were stripped, those graphs retain unset initializers and package saving fails; if the weights remain, the export still adds a large component the checkpoint explicitly disabled.

Gate Talker configuration extraction on `enable_audio_output` (and any supported legacy equivalent), and add a configuration test where `talker_config` exists but audio output is disabled.

### [P2] Add numerical parity coverage for the new Talker path

**File:** `tests/synthetic_parity_test.py`  
**Lines:** 49-66  
**Confidence:** High

The existing Qwen2.5-Omni synthetic parity test explicitly sets `config.talker = None`. The new tests validate Talker graph signatures and cache shapes, but do not compare either Talker model against the Transformers reference.

Errors in Talker weight routing, the 3584-to-896 projection, MRoPE application, attention biases, or cached decoding can therefore produce structurally valid ONNX graphs with incorrect codec logits while all current tests pass.

Extend synthetic parity coverage to compare `talker_embedding` and `talker` outputs with a small native `Qwen2_5OmniTalkerForConditionalGeneration`, including prefill and one cached decode step.

## Summary

The PR adds nested Talker configuration extraction, Talker embedding and decoder models, Thinker hidden-state output, six-component Qwen2.5-Omni packaging, Talker weight routing, and graph/configuration tests. It should respect checkpoints that disable audio output and add numerical coverage for the new model path.

rui-ren:
@copilot+gpt-6-astra Please resolve the comments.

